PR #19, opened 2018-11-16
chore(stream): use parking_lot crate instead of std::sync for Mutex, …

ekrenzke:
…RwLock


src/stream/connection.rs
```diff
@@ -164,68 +165,62 @@ impl Connection {
...
-                        }))
-                    } else {
-                        trace!("Stream {} does not have any data to send", stream.id);
-                    }
-
-                    // Inform other side about closing streams
-                    if stream.is_closing() {
-                        trace!("Sending stream close frame for stream {}", stream.id);
-                        frames.push(Frame::StreamClose(StreamCloseFrame {
+            //new
```
emschwartz: this should be removed

src/stream/data_money_stream.rs
```diff
@@ -410,35 +389,29 @@ impl Write for DataStream {
...
                 ErrorKind::WouldBlock,
-                "Unable to get lock on outgoing",
+                "Not finished sending yet",
             ))
         }
     }
 }
 impl AsyncWrite for DataStream {
     fn shutdown(&mut self) -> Result<Async<()>, IoError> {
         match self.state.try_write() {
```
emschwartz: We should get rid of this try_write here -- don't need that anymore

PR #20, opened 2018-11-17
Add Interledger.rs Logo

khoslaventures:
I created a logo for the repo. Based on the license of the Rust logo, we'll have to add Creative Commons somewhere (https://www.rust-lang.org/en-US/legal.html). Please place wherever it seems best!

EDIT: With the latest gear/cog version, no need to do the above.


kincaidoneil:
Personally, I really liked that first logo! (with the thicker outline)

ekrenzke:
Nice! I really like the latest cog version. From first glance, the "highlight" colors might be slightly off compared to the Interledger.js and standard Interledger logo, but I think those two are actually inconsistent, so it looks great to me.

khoslaventures:
Went through a few iterations of both and I think what we really need is a cohesive design system for all of Interledger after discussing with @emschwartz. I'm thinking it's best to just throw this on here and come back to all of those logos down the line. I'd like for the ecosystem to have a consistent identity!

README.md
```diff
@@ -1,3 +1,71 @@
+<p align="center">
+  <img src="ilprs.svg" width="700">
+</p>
+
+
+# ilp.rs
```
emschwartz: You re-copied the whole readme again here

khoslaventures: ah, should be fixed now! This is ready to merge.

tarcieri:
Looks nice!

PR #28, opened 2018-12-03
Rename crate to interledger

emschwartz:
Supports running the CLI as either `interledger` or `ilp`


tarcieri:
Funny thing I've noticed with various GitHub projects lately: I've traditionally named my GitHub repos ala `interledger-rs`, but I noticed `interledger.rs` works just as well for a project name. Just started playing around with that sort of thing here:

https://github.com/iqlusioninc/keychain-services.rs

PR #29, opened 2018-12-05
Add stateless sending and receiving to STREAM

emschwartz:
- Implements the STREAM single packet receiver function mentioned in https://github.com/emschwartz/interledger-rs/issues/25
- Adds a `send_money` command that enables money to be sent more easily than creating a connection, creating a stream, and sending money on that stream
- Switches SPSP.pay to use `stream::send_money`

Note the simplified functions will also be useful for compiling this library to WASM. Creating a `stream::Connection` requires the future to be run with a Tokio executor (because there are calls to `tokio::spawn`). The `send_money` and `receive_money_statelessly` functions should be usable with the minimal [`future_to_promise`](https://rustwasm.github.io/wasm-bindgen/api/wasm_bindgen_futures/fn.future_to_promise.html) executor.

Open questions:
1. Should `receive_money_statelessly` take a function that will be called when money is received?
2. Should the SPSP server be switched to use the stateless receiver functionality? (This would require 1.)


ekrenzke:
>     1. Should `receive_money_statelessly` take a function that will be called when money is received?

If I am understanding this correctly, it seems like said function would be commonly used for business logic which should probably be handled by the connector. On the other hand, it seems really useful to pass in arbitrary functions depending on the event that occurs. It could make returning errors to the sender a lot easier as well.

It's still not clear to me whether or not this should be implemented here or in the connector (leaning towards the latter), but it would be nice to have a `Handler` trait for the connector to implement `on_amount`, `on_open`, `on_close`, and similar event functionality.

>     2. Should the SPSP server be switched to use the stateless receiver functionality? (This would require 1.)

What are the main advantages to switching to the stateless receiver for the SPSP server besides AWS Lambda? Are there any disadvantages? I think the default behavior should be as close to the reference implementation as possible, but optional stateless receiver functionality does seem convenient.

emschwartz:
> If I am understanding this correctly, it seems like said function would be commonly used for business logic which should probably be handled by the connector.

Nope, I don't think you understood it correctly. STREAM is never implemented by connectors, it is implemented by the sender and receiver.

> On the other hand, it seems really useful to pass in arbitrary functions depending on the event that occurs.

There is only one event here: money received. If you were using an application protocol that sent data also, you would want a single receiver handling all of the incoming packets so that it can reassemble data split across multiple packets correctly.

> It could make returning errors to the sender a lot easier as well.

What errors? The point of the `oneshot` implementation is to make sending and receiving money simple. If you don't want to receive all the money you might get in, the oneshot receiver isn't for you.

> What are the main advantages to switching to the stateless receiver for the SPSP server besides AWS Lambda?

Not needing to manage connection state or make sure you clean up inactive or closed connections. Also, if the SPSP receiver is only receiving money anyway, the stateless receiver function is a more targeted version of STREAM optimized for what SPSP needs right now.

ekrenzke:
> STREAM is never implemented by connectors, it is implemented by the sender and receiver.
>
> There is only one event here: money received. If you were using an application protocol that sent data also, you would want a single receiver handling all of the incoming packets so that it can reassemble data split across multiple packets correctly.

This makes sense to me now. My apologies, I was under the impression that a connector could be a "receiver," but it is now clear to me that this is for end-to-end payments.


> What errors? The point of the `oneshot` implementation is to make sending and receiving money simple. If you don't want to receive all the money you might get in, the oneshot receiver isn't for you.

Fair enough. My primary motivation for introducing a `Handler` trait was for processing and propagating [error codes](https://github.com/interledger/rfcs/blob/master/0027-interledger-protocol-4/0027-interledger-protocol-4.md#Error-Codes) (e.g. X99) at the application layer, but there is no use for it in `one-shot` and completely unrelated, as you pointed out. Sorry for my confusion.

Thank you for the clarification.

PR #36, opened 2019-02-11
ilp packet zero copy decoding

sentientwaffle:
## API changes

  * `Ilp{Packet,Prepare,Fulfill,Reject}` → `{Packet,Prepare,Fulfill,Reject}` to avoid naming stutter.
  * Packet fields are accessible via getter instead of public fields.
  * Add `Prepare.set_amount()` and `Prepare.set_expires_at()`.
  * Removed the `Serializable` trait, since it didn't seem to ever be needed. Packets are serialized with `Into<BytesMut>` and deserialized with `::new(BytesMut)`, which I think is more idiomatic. I'd prefer to use `TryFrom` but it's still not stabilized.
  * Remove `Unknown` from the `PacketType` enum.
  * Add `ErrorCode` type with associated constants for all canonical ilp errors.

## Fixes

  * Don't crash when decoding an empty buffer.
  * Fix bug that encoded VarOctetString length incorrectly.

## OER changes

  * Packets can be constructed using `{Prepare,Fulfill,Reject}Builder.build()`.
  * The `oer` module now exposes only 2 traits: `BufOerExt` and `MutBufOerExt`. The former supports fallible reads on a `&[u8]`. The latter supports infallible writes on a `BufMut`.
  * Reading a VarOctetString no longer copies any data, just borrows it from the buffer.
  * Some other misc optimizations (mostly avoiding additional copies/allocations).

## Other notes

  * The project is rearranged to use workspaces so that `interledger-packet` is its own crate.
  * Added benchmarks for packet serialization/deserialization.
  * The interledger-packet uses rust edition 2018 since it ended up using nll in a bunch of places.
  * CI is failing because of some tower-web change unrelated to this PR.


crates/interledger-packet/src/packet.rs
```diff
@@ -0,0 +1,735 @@
...
+        self.expires_at = expires_at;
+        let offset = self.content_offset + AMOUNT_LEN;
+        write!(
+            &mut self.buffer[offset..], "{}",
+            DateTime::<Utc>::from(expires_at)
+                .format(INTERLEDGER_TIMESTAMP_FORMAT),
+        ).unwrap();
+    }
+
+    /// The returned value always has a length of 32.
```
emschwartz: Should this return a `&[u8; 32]` then? If you want it as a slice you'd just do `&packet.execution_condition()[..]`

crates/interledger-packet/src/packet.rs
```diff
@@ -0,0 +1,735 @@
...
+    #[inline]
+    pub fn execution_condition(&self) -> &[u8] {
+        let begin = self.content_offset + AMOUNT_LEN + EXPIRY_LEN;
+        let end = begin + CONDITION_LEN;
+        &self.buffer[begin..end]
+    }
+
+    #[inline]
+    pub fn destination(&self) -> &[u8] {
```
emschwartz: Hardcore -- not even turning it into a String or &str. Alright :)

crates/interledger-packet/src/packet.rs
```diff
@@ -0,0 +1,735 @@
...
+pub struct PrepareBuilder<'a> {
+    pub amount: u64,
+    pub expires_at: SystemTime,
+    pub execution_condition: &'a [u8; 32],
+    pub destination: &'a [u8],
+    pub data: &'a [u8],
+}
+
+impl Prepare {
+    // TODO change this to `TryFrom` when it is stabilized
```
emschwartz: What do you think about calling this method `try_from` now? We can make it implement the proper trait when it's stabilized.

Looking at the API now, it seems a bit unusual for `new` to return a Result

crates/interledger-packet/src/packet.rs
sentientwaffle: That would be my first choice, but as far as I know there isn't a way to borrow a slice as an array (please correct me if I'm wrong).

emschwartz: Oh I think you're right, never mind

crates/interledger-packet/src/packet.rs
sentientwaffle: Yeah, converting to a `String` would be a copy, and to `&str` requires utf8 validation which shouldn't be necessary.

Most connector routers probably don't care about validating the destination at all, since if it is an invalid address at worst it will be forwarded to an unexpected destination (from the sender's point of view). And utf8 validation is overkill regardless since ILP addresses are a subset of ascii.

emschwartz:
LGTM once the tests are passing

emschwartz:
These commits were included in https://github.com/emschwartz/interledger-rs/pull/38

PR #54, opened 2019-03-25
Add redis-backed Interledger "node"

emschwartz:
The node includes:
- SPSP/STREAM sender and receiver
- ILP connector (that supports BTP and ILP-Over-HTTP)
- A basic API for managing accounts/peers and configuration
- XRP on-ledger settlement engine
- Redis-backed store
- Docker container that bundles together the Rust components, Redis, and the settlement engine


crates/interledger-api/src/lib.rs
```diff
@@ -0,0 +1,341 @@
+#![recursion_limit = "256"]
+#[macro_use]
+extern crate tower_web;
```
sentientwaffle: In rust 2018 you can use some_crate::some_macro; (I think it works for derives as well) and remove the extern crate items.

PR #61, opened 2019-05-02
[WIP] do not merge

### LINKED ISSUE #37: Check max data size and ILP address length
The packet parsing module should not accept packets whose data or address lengths exceed the max values

dora-gt:
Fixes #37 [WIP]

- Using LimitStream of [interledger-relay](https://github.com/coilhq/interledger-relay)
  - The license was unclear so [I'm asking @sentientwaffle](https://github.com/coilhq/interledger-relay/issues/2)


crates/interledger-http/src/server.rs
```diff
@@ -122,3 +133,65 @@ fn ilp_response_to_http_response(
...
+            prepare_data.execution_condition,
+            result.execution_condition()
+        );
+        // TODO this fails🤔 look into later.
+        // assert_eq!(prepare_data.expires_at, result.expires_at());
+        assert_eq!(prepare_data.data, result.data());
+    }
+
+    #[test]
+    #[should_panic]
```
emschwartz: Instead of making this test panic, you could make the helper function return a `Result<Prepare, ()>` and then use an `assert!(result.is_err())`

crates/interledger-http/src/server.rs
```diff
@@ -94,18 +98,25 @@ where
 
 fn parse_prepare_from_request(
     request: Request<Body>,
+    max_message_size: Option<usize>,
 ) -> impl Future<Item = Prepare, Error = Response<Body>> + 'static {
-    request
-        .into_body()
+    let max_message_size = max_message_size.unwrap_or(MAX_MESSAGE_SIZE);
```
emschwartz: It seems like setting `max_message_size` to `None` should imply there is no limit, rather than the default limit. I think it would be slightly clearer to pass in `Some(MAX_MESSAGE_SIZE)` above

crates/interledger-util/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "interledger-util"
+version = "0.1.0"
+authors = ["Taiga Nakayama <dora@dora-gt.jp>"]
+edition = "2018"
+publish = false
```
emschwartz: I think this needs to be publishable because the other crates depend on it

It also looks like we can use the [`max_message_size`](https://docs.rs/tungstenite/0.6.1/tungstenite/protocol/struct.WebSocketConfig.html#structfield.max_message_size) for the websocket connection, so we might want to just fold this helper function into the `interledger-http` crate instead of making a new util crate.

README.md
```diff
@@ -16,6 +16,10 @@
 - [Interledger.rs Architecture](./docs/architecture.md)
 - [Interledger Forum](https://forum.interledger.org) for general questions about the Interledger Protocol and Project
 
+## Understanding Rust
```
emschwartz: I don't think we need to link these resources here. We should add a `CONTRIBUTING.md` file though, and it could be useful to suggest those resources for someone that wants to contribute. For others that don't know Rust, we should have docs for how to run and use the dockerized bundle or CLI without understanding anything about the internals

crates/interledger-http/src/server.rs
dora-gt: Wow...indeed. It is really smarter😳

crates/interledger-util/Cargo.toml
dora-gt: Ah...indeed. That was wrong for `publish`, I'm sorry.

I'll refactor to include this function into `interledger-http`.

dora-gt:
Close this, the arranged PR is found here: https://github.com/emschwartz/interledger-rs/pull/62

PR #62, opened 2019-05-03
Fix #37

### LINKED ISSUE #37: Check max data size and ILP address length

dora-gt:
- Fix https://github.com/emschwartz/interledger-rs/issues/37
- Previous discussions are here: https://github.com/emschwartz/interledger-rs/pull/61
- Prevents from sending huge amount of data through HTTP requests using limit_stream.
  - max message size is defined in `interledger-http` and `interledger-btp` uses it.
- `limit_stream` is originally from [interledger-relay](https://github.com/coilhq/interledger-relay) by @sentientwaffle

### DIFF
--- a/crates/interledger-http/src/lib.rs
+++ b/crates/interledger-http/src/lib.rs
@@ -13,6 +13,9 @@ use url::Url;
 mod client;
 mod server;
 
+/// Originally from [interledger-relay](https://github.com/coilhq/interledger-relay/blob/master/crates/interledger-relay/src/combinators/limit_stream.rs).
+mod limit_stream;
+
 pub use self::client::HttpClientService;
 pub use self::server::HttpServerService;
 
--- a/crates/interledger-http/src/limit_stream.rs
+++ b/crates/interledger-http/src/limit_stream.rs
@@ -1,0 +1,155 @@
+// Copyright 2019 Coil Technologies
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the
+// License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+// either express or implied. See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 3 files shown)


crates/interledger-btp/Cargo.toml
```diff
@@ -17,6 +17,7 @@ hashbrown = "0.1.8"
 interledger-ildcp = { path = "../interledger-ildcp", version = "0.2.1" }
 interledger-packet = { path = "../interledger-packet", version = "0.2.1" }
 interledger-service = { path = "../interledger-service", version = "0.2.1" }
+interledger-http = { path = "../interledger-http", version = "0.2.1" }
```
emschwartz: BTP shouldn't depend on HTTP

crates/interledger-btp/src/server.rs
```diff
@@ -4,6 +4,7 @@ use super::{
 use base64;
 use bytes::{BufMut, Bytes, BytesMut};
 use futures::{future::result, Future, Sink, Stream};
+use interledger_http::MAX_MESSAGE_SIZE;
```
emschwartz: This should just be a constant in this file -- or the MAX_PACKET_LENGTH could be a constant exported by the `interledger-packet` crate

crates/interledger-btp/Cargo.toml
dora-gt: I thought so at first but your comment
https://github.com/emschwartz/interledger-rs/pull/61#discussion_r280514095
looked saying moving the functions into `interledger-http`. I'm sorry I misunderstood.

dora-gt:
Fixed not to affect `interledger-btp`.

dora-gt:
Btw, if I want to make a PR with a single commit after some fixes, I have to do either of:

- `push -f` like I did in this branch
- Make a new PR with a new branch

Which is preferred?

emschwartz:
I think force pushing to the same branch is better than creating a new branch each time. Force pushing to master is a no-no but I do it plenty when I modify or squash commits

emschwartz:
@dora-gt Can you trigger a rebuild on CircleCI? I don't know why the coverage report generation step is causing problems but sometimes it seems like kcov does non-deterministic things. If that doesn't work I'll just merge it and try to figure out why that step is failing later

dora-gt:
image
It seems that it requires write permissions.

dora-gt:
I'm not sure but this may help: https://github.com/SimonKagstrom/kcov/issues/151
It looks like a similar problem.

PR #71, opened 2019-05-28
Add echo service

dora-gt:
Though this is WIP (and actually not working), you might be able to confirm the base implementation.
If you have any opinion so far, could you comment here?

### DIFF
--- a/crates/interledger-service-util/src/echo.rs
+++ b/crates/interledger-service-util/src/echo.rs
@@ -1,0 +1,441 @@
+use byteorder::ReadBytesExt;
+use bytes::{BufMut, Bytes, BytesMut};
+use core::borrow::Borrow;
+use futures::future::err;
+use interledger_packet::{
+    oer,
+    oer::{BufOerExt, MutBufOerExt},
+    ErrorCode, Prepare, PrepareBuilder, RejectBuilder,
+};
+use interledger_service::*;
+use std::convert::TryFrom;
+use std::marker::PhantomData;
+use std::str;
+use std::time::SystemTime;
+
+/// A service that responds to the Echo Protocol.
+/// Currently, this service only supports bidirectional mode (unidirectional mode is not supported yet).
+/// The service doesn't shorten expiry as it expects the expiry to be shortened by another service
+/// like `ExpiryShortenerService`.
+
+/// The prefix that echo packets should have in its data section
+const ECHO_PREFIX: &str = "ECHOECHOECHOECHO";
+/// The length of the `ECHO_PREFIX`
+const ECHO_PREFIX_LEN: usize = 16;
+
+enum EchoPacketType {
+    Request = 0,
+    Response = 1,
+}
+
+#[derive(Clone)]
+pub struct EchoService<S, A> {
+    /// The ILP address which this ECHO service should respond for
+    ilp_address: Bytes,
+    next_incoming: S,
+    account_type: PhantomData<A>,
+}
+
+impl<S, A> EchoService<S, A>
+where
+    S: IncomingService<A>,
+    A: Account,
+{
+    pub fn new(ilp_address: Bytes, next_incoming: S) -> Self {
+        EchoService {
+            ilp_address,
+            next_incoming,
+            account_type: PhantomData,
+        }
+    }
+}
+
+impl<S, A> IncomingService<A> for EchoService<S, A>
+where
+    S: IncomingService<A>,
+    A: Account,
+{
+    type Future = BoxedIlpFuture;
+
+    fn handle_request(&mut self, mut request: IncomingRequest<A>) -> Self::Future {
+        let should_echo = request.prepare.destination() == self.ilp_address.borrow() as &[u8]
+            && request.prepare.data().starts_with(ECHO_PREFIX.as_bytes());
+        if !should_echo {
+            return Box::new(self.next_incoming.handle_request(request));
+        }
+
+        let mut reader = request.prepare.data();
+
+        // prefix is checked above, so skip prefix bytes.
+        reader.skip(ECHO_PREFIX_LEN).unwrap();
+
+        // check echo packet type
+        let echo_packet_type = match reader.read_u8() {
+            Ok(value) => value,
+            Err(error) => {
+                eprintln!("Could not read packet type: {:?}", error);
+                return Box::new(err(RejectBuilder {
+                    code: ErrorCode::F01_INVALID_PACKET,
+                    message: b"Could not read echo packet type.",
+                    triggered_by: self.ilp_address.borrow(),
+                    data: &[],
+                }
+                .build()));
+            }
+        };
+        if echo_packet_type == EchoPacketType::Response as u8 {
+            // if the echo packet type is Response, just pass it to the next service
+            // so that the initiator could handle this packet
+            return Box::new(self.next_incoming.handle_request(request));
+        }
+        if echo_packet_type != EchoPacketType::Request as u8 {
+            eprintln!("The packet type is not acceptable: {}", echo_packet_type);
+            return Box::new(err(RejectBuilder {
+                code: ErrorCode::F01_INVALID_PACKET,
+                message: format!(
+                    "The echo packet type: {} is not acceptable.",
+                    echo_packet_type
+                )
+                .as_bytes(),
+                triggered_by: self.ilp_address.borrow(),
+                data: &[],
+            }
+            .build()));
+        }
+
+        // check source address
+        let source_address = match reader.read_var_octet_string() {
+            Ok(value) => value,
+            Err(error) => {
+                eprintln!("Could not read source address: {:?}", error);
+                return Box::new(err(RejectBuilder {
+                    code: ErrorCode::F01_INVALID_PACKET,
+                    message: b"Could not read source address.",
+                    triggered_by: self.ilp_address.borrow(),
+                    data: &[],
+                }
+                .build()));
+            }
+        };
+
+        // create a new prepare packet to echo the prepare
+        let execution_condition =
+            <[u8; 32]>::try_from(request.prepare.execution_condition()).unwrap();
+        request.prepare = EchoResponseBuilder {
+            amount: request.prepare.amount(),
+            expires_at: request.prepare.expires_at(),
+            execution_condition: &execution_condition,
+            destination: source_address,
+        }
+        .build();
+
+        return Box::new(self.next_incoming.handle_request(request));
+    }
+}
+
+pub struct EchoRequestBuilder<'a> {
+    pub amount: u64,
+    pub expires_at: SystemTime,
+    pub execution_condition: &'a [u8; 32],
+    /// The ILP address that the initiator wants to Ping
+    pub destination: &'a [u8],
+    /// The ILP address of the initiator
+    pub source_address: &'a [u8],
+}
+
+impl<'a> EchoRequestBuilder<'a> {
+    pub fn build(&self) -> Prepare {
+        let source_address_len = oer::predict_var_octet_string(self.source_address.len());
+        let mut data_buffer = BytesMut::with_capacity(ECHO_PREFIX_LEN + 1 + source_address_len);
+        data_buffer.put(ECHO_PREFIX.as_bytes());
+        data_buffer.put_u8(EchoPacketType::Request as u8);
+        data_buffer.put_var_octet_string(self.source_address);
+        PrepareBuilder {
+            amount: self.amount,
+            expires_at: self.expires_at,
+            execution_condition: self.execution_condition,
+            destination: self.destination,
+            data: data_buffer.borrow(),
+        }
+        .build()
+    }
+}
+
+pub struct EchoResponseBuilder<'a> {
+    pub amount: u64,
+    pub expires_at: SystemTime,
+    pub execution_condition: &'a [u8; 32],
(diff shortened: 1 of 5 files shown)


crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,145 @@
...
+        let mut data_buffer = BytesMut::with_capacity(1024);
+        data_buffer.put(ECHO_PREFIX.as_bytes());
+        data_buffer.put_u8(EchoPacketType::Response as u8);
+
+        // change prepare parameters to be routed as appropriate
+        request.prepare.set_destination(source_address);
+        request
+            .prepare
+            .set_expires_at(request.prepare.expires_at() - Duration::from_millis(1000));
+        request.prepare.set_data(data_buffer.borrow());
```
dora-gt: calling `set_destination` and `set_data` seems inefficient, I'll fix this.

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,149 @@
...
+{
+    type Future = BoxedIlpFuture;
+
+    fn handle_request(&mut self, mut request: IncomingRequest<A>) -> Self::Future {
+        let should_echo = request.prepare.destination() == self.get_own_address().as_bytes();
+        if !should_echo {
+            return Box::new(self.next_incoming.handle_request(request));
+        }
+
+        let suffices_min_length = MIN_DATA_SIZE <= request.prepare.data().len();
```
emschwartz: Nitpick: you can get rid of this check because the next one (checking whether the data starts with the right prefix) will catch the same case

crates/interledger-packet/src/packet.rs
```diff
@@ -102,6 +124,15 @@ pub struct Prepare {
     data_offset: usize,
 }
 
+/// The struct when updating Prepare with multiple parameters efficiently
+pub struct PrepareUpdateParams<'a> {
```
emschwartz: Instead of enabling all of the fields to be updated, I think we should just create a new `Prepare` packet for the Echo handler. This is the only case where we want to change the `Prepare` details (beyond the amount and expiry) and we know the data will be short so copying it over or serializing a new packet will not be much overhead

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,149 @@
...
+        if !starts_with_echo_prefix {
+            return self.get_invalid_prefix_error();
+        }
+
+        let mut reader = Vec::<u8>::with_capacity(request.prepare.data().len());
+        reader.extend_from_slice(request.prepare.data());
+        let mut reader: &[u8] = reader.borrow();
+        reader.skip(ECHO_PREFIX.len()).ok();
+        let echo_packet_type = reader.read_u8().unwrap();
+        let source_address = reader.read_var_octet_string().unwrap();
```
emschwartz: You shouldn't unwrap here because we don't know if this is a properly formatted packet and we don't want to panic if not

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,149 @@
...
+
+        let is_packet_type_request = echo_packet_type == EchoPacketType::Request as u8;
```
emschwartz: What if it's the response coming back? Right now we're not sending outgoing echo requests but we could add that in the future and wouldn't want this to reject those packets

crates/interledger-packet/src/packet.rs
dora-gt: ok!

crates/interledger-service-util/src/echo.rs
dora-gt: `MIN_DATA_SIZE` means echo prefix + 1 byte, so I thought this meant for ensuring next 1 byte is readable.

crates/interledger-service-util/src/echo.rs
dora-gt: ↑

dora-gt: isn't it?

crates/interledger-service-util/src/echo.rs
emschwartz: Got it. Instead of doing that, I would recommend handling the error case for when you try reading the `u8`, which will catch if there isn't another byte to read

crates/interledger-service-util/src/echo.rs
dora-gt: oops! I see. `read_var_octet_string` should not be unwrapped!

dora-gt:
Idk the exact word for “考慮漏れ” in English, it means "something beyond the consideration" or "something one did not consider", I'm sorry I have so many "考慮漏れ"😔 and thank you for your review.

emschwartz:
No problem at all! I'm very grateful that you're jumping in to contribute so there's no need to apologize. I'm more than happy to review PRs :)

dora-gt:
I'll take more time to learn Rust and Interledger.rs to write better code!

dora-gt:
I added some unit tests for the service and it looks working at least unit test level but I have some questions.

1. Currently, the service accepts `ilp_address` argument for its own ILP address (the server's address). This is because I thought there may be a case that the node needs to create echo packets not using ILDCP address but using a statically specified address. e.g. When a connector node needs to create echo packets, being fired from Admin APIs. Is that OK?
1. The service expects that expiry is shortened by another service like `ExpiryShortenerService`. Is that OK?

or do you see any to fix so far?

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,444 @@
...
+{
+    pub fn new(ilp_address: Bytes, next_incoming: S) -> Self {
+        EchoService {
+            ilp_address,
+            next_incoming,
+            account_type: PhantomData,
+        }
+    }
+
+    fn get_error_response(
```
emschwartz: Nitpick: I don't think this function is necessary. It's simple enough to use the RejectBuilder and return the boxed future in the other places the errors are created.

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,444 @@
...
+            data,
+        }
+        .build();
+        return Box::new(err(result));
+    }
+
+    fn get_invalid_prefix_error(&self) -> BoxedIlpFuture {
+        return self.get_error_response(
+            ErrorCode::F01_INVALID_PACKET,
+            "packet data does not start with ECHO prefix.",
```
emschwartz: Nitpick: as we discussed on slack, let's start the error messages with capital letters

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,444 @@
...
+
+impl<S, A> IncomingService<A> for EchoService<S, A>
+where
+    S: IncomingService<A>,
+    A: Account,
+{
+    type Future = BoxedIlpFuture;
+
+    fn handle_request(&mut self, mut request: IncomingRequest<A>) -> Self::Future {
+        let should_echo = request.prepare.destination() == self.ilp_address.borrow() as &[u8];
```
emschwartz: We should also look at the packet data to make sure this is an echo packet. There may be other reasons to send packets addressed to the node itself but right now those would be rejected

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,444 @@
...
+use interledger_service::*;
+use std::convert::TryFrom;
+use std::io::Read;
+use std::marker::PhantomData;
+use std::str;
+use std::time::SystemTime;
+
+/// A service that responds to the Echo Protocol.
+/// Currently, this service only supports bidirectional mode (unidirectional mode is not supported yet).
+/// The service doesn't shorten expiry as it expects the expiry to be shortened by another service
```
emschwartz: I agree with this decision

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,444 @@
...
+        let source_address = match reader.read_var_octet_string() {
+            Ok(value) => value,
+            Err(err) => {
+                eprintln!("could not read source address: {:?}", err);
+                return self.get_source_address_not_readable_error();
+            }
+        };
+
+        // create a new prepare packet to echo the prepare
+        let execution_condition = match <[u8; 32]>::try_from(request.prepare.execution_condition())
```
emschwartz: It shouldn't be possible for this to be malformed. The Interledger packet parser should make sure of that

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,444 @@
...
+    pub expires_at: SystemTime,
+    pub execution_condition: &'a [u8; 32],
+    /// The ILP address that the initiator wants to Ping
+    pub destination: &'a [u8],
+    /// The ILP address of the initiator
+    pub source_address: &'a [u8],
+}
+
+impl<'a> EchoRequestBuilder<'a> {
+    pub fn build_echo_request(&self) -> Prepare {
```
emschwartz: Nitpick: for consistency, just call it `build` or `to_prepare`

crates/interledger-service-util/src/echo.rs
dora-gt: Yeah... that was too much. I'll fix.

crates/interledger-service-util/src/echo.rs
dora-gt: I'm sorry, I forgot to fix.

crates/interledger-service-util/src/echo.rs
dora-gt: Indeed. I'll fix.

dora-gt: I'm wondering which is better:

1. Packets that have **echo prefix** should be handled as echo packets.
    - This could cause errors when it fails to parse in farther bytes.
1. Packets that are correctly parsed as echo packets should be handled as echo packets.
    - This never causes errors because when it fails to parse, it will be handled as non-echo packets.

dora-gt: I think (1) is better because if someone tries to send an echo packet and unfortunately the packet is malformed, the service returns an appropriate error. In addition, (2) is a bit non-explicit for a protocol. If we adopt (1), the prefix should be declared as special `data` prefix in RFCs.

crates/interledger-service-util/src/echo.rs
dora-gt: Indeed.

I think it would be better if `execution_condition` of `Prepare` returns `&[u8;32]` so we don't need to care about this. Is there any reason?

emschwartz: Actually I don't remember why we made it return a `&[u8]` instead of a `&[u8;32]`. I'd be up for changing that (@sentientwaffle any objections?)

crates/interledger-service-util/src/echo.rs
emschwartz: Agree with 1

dora-gt:
It seems ready to merge (except for `execution_condition` but it might be a different PR) so I squashed commits and fixed the commnet for it.

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,452 @@
...
+        let should_echo = request.prepare.destination() == self.ilp_address.borrow() as &[u8]
+            && request.prepare.data().starts_with(ECHO_PREFIX.as_bytes());
+        if !should_echo {
+            return Box::new(self.next_incoming.handle_request(request));
+        }
+
+        let mut reader = request.prepare.data();
+
+        // prefix is checked above, so skip prefix bytes.
+        reader.skip(ECHO_PREFIX_LEN).ok();
```
emschwartz: This should be an `unwrap` or `expect`, since you aren't actually using the Option return value

crates/interledger-service-util/src/echo.rs
```diff
@@ -0,0 +1,452 @@
...
+{
+    pub fn new(ilp_address: Bytes, next_incoming: S) -> Self {
+        EchoService {
+            ilp_address,
+            next_incoming,
+            account_type: PhantomData,
+        }
+    }
+
+    fn get_packet_type_not_readable_error(&self) -> BoxedIlpFuture {
```
emschwartz: Actually, sorry, another kind of nitpick: I don't think this needs to be a separate function if it's only called from one place in the code

crates/interledger-service-util/src/echo.rs
dora-gt: oops... I misunderstood `ok`. I'll fix.

crates/interledger-service-util/src/echo.rs
dora-gt: My intention was keeping the source code readable but yeah, it might be too much. I'll fix.

emschwartz: Gotcha. For what it's worth, my take on this kind of thing is that splitting code out into separate functions can help make certain things more readable but also sometimes makes it harder to understand because you need to bounce around the file to figure out what's going on. In this case, it's a pretty minor difference but I think returning the Reject packet right there is clear enough as to what's going on

dora-gt:
Fixed! Sorry for taking time.

emschwartz:
I'm gonna merge this even though the checks aren't passing right now. There's some unrelated issue with kcov that I need to figure out

PR #75, opened 2019-05-31
Address type for ILP-Addresses

### LINKED ISSUE #69: Create an IlpAddress type
This type should:
- [x] Wrap a [`Bytes`](https://github.com/carllerche/bytes)
- [x] Check that the address conforms to the [spec](https://interledger.org/rfcs/0015-ilp-addresses/)
- [x] Implement `Debug`, `Display`, `FromStr`
- [ ] Implement `Deref` for the targets `&str` and `&[u8]`
- [x] Implement serde's `Serialize`, `Deserialize`

This will let us efficiently use the addresses and avoid copying them but also enable printing them without constantly re-parsing the bytes as utf8.

Credit to @sentientwaffle for this idea.

gakonst:
Closes #69. Heavily based on the work done in https://github.com/emschwartz/interledger-rs/blob/dj-connector-relay2/crates/interledger-packet/src/address.rs by @sentientwaffle

Implements:
- Serde serialize/deserialize
- Display + Debug
- Default
- Deref<Target = [u8]> (easy to go to string from there)
- AsRef<Bytes> and AsRef<[u8]
- FromStr
- TryFrom (now that it's standardized)

A follow up PR will be building on top of this and replaces all instances of Bytes/Vec<u8>/[u8].

@emschwartz lmk if I should better organize/squash the commits. Are we missing any feature?


crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,366 @@
...
+        formatter
+            .debug_tuple("Address")
+            .field(&str::from_utf8(&self.0).unwrap())
+            .finish()
+    }
+}
+
+impl fmt::Display for Address {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str(str::from_utf8(&self.0).unwrap())
```
emschwartz: We should avoid re-parsing this every time we want to print it. I wonder if this should be backed by the [String crate](https://docs.rs/string/0.2.0/string/) instead of a Bytes or if we should use the unsafe [`str::from_utf8_unchecked`](https://doc.rust-lang.org/std/str/fn.from_utf8_unchecked.html) method

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,366 @@
...
+    pub unsafe fn new_unchecked(bytes: &[u8]) -> Self {
+        Address(Bytes::from(bytes))
+    }
+
+    /// ```text
+    /// scheme = "g" / "private" / "example" / "peer" / "self" /
+    ///          "test" / "test1" / "test2" / "test3" / "local"
+    /// ```
+    #[inline]
+    pub fn scheme(&self) -> &[u8] {
```
emschwartz: How about making this return a `&str` instead?

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,366 @@
...
+        self.0.split(|&byte| byte == b'.').next().unwrap()
+    }
+
+    /// Returns the local part (right-most '.' separated segment) of the ILP Address.
+    pub fn local(&self) -> &[u8] {
```
emschwartz: Instead of this, how about having a method that returns an Iterator over the segments?

I'm not a fan of calling this method "local", because local-ness is relative and there could be different segments attached by different applications, each of which considers a different part "local".

crates/interledger-packet/src/errors.rs
```diff
@@ -30,14 +32,15 @@ quick_error! {
             description(descr)
             display("Wrong Type {}", descr)
         }
+        AddressError(err: AddressError) {
+            from()
+            description(err.description())
+            cause(err)
+        }
+        // TODO &'static str instead of String?
```
emschwartz: FWIW, `String` allows us to give more specific context as to why this error was triggered

emschwartz:
Also, it would be good to squash the commits before merging, but you can do that once we're finished reviewing it.

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,366 @@
...
+        if is_valid && segments > 1 {
+            Ok(Address(Bytes::from(bytes)))
+        } else {
+            Err(AddressError)
+        }
+    }
+}
+
+impl std::ops::Deref for Address {
+    type Target = [u8];
```
emschwartz: It might be more helpful to deref as an `&str`, because it's easy to go from that to `&[u8]` but requires checking the string again to go from `&[u8]` to `&str`

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,366 @@
...
+}
+
+impl Address {
+    /// Returns the length of the ILP Address.
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Returns an empty ILP Address with "0" as its only byte.
```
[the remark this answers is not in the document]
gakonst: So regarding this, I agree, but in multiple other places of the codebase there's the `triggered_by` field which is filled with an empty slice `&[]`, e.g. https://github.com/emschwartz/interledger-rs/blob/master/crates/interledger-btp/src/lib.rs#L175. The [`RejectBuilder`](https://github.com/emschwartz/interledger-rs/blob/master/crates/interledger-packet/src/packet.rs#L349) and other similar structures take that type of argument. How should that field be handled in your opinion?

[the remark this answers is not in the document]
emschwartz: We _could_ make that an `Option<Address>`. It really should be filled in everywhere that it's created in this codebase but we also can't guarantee that it's set by everyone else

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,366 @@
...
+
+impl std::str::FromStr for Address {
+    type Err = AddressError;
+
+    fn from_str(src: &str) -> Result<Self, Self::Err> {
+        Address::try_from(Bytes::from(src))
+    }
+}
+
+impl Default for Address {
```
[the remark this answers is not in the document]
gakonst: Required for using `ilp_address: Address` here: https://github.com/emschwartz/interledger-rs/blob/master/crates/interledger-store-memory/src/account.rs#L86.

Maybe we can define a default address and use that? Not sure.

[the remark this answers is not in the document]
emschwartz: I'd be okay getting rid of that Default as well

crates/interledger-packet/src/address.rs
gakonst: Yeah so we can use a method that just splits the address in '.' separated segments. `scheme` would return the first element of the split. I preferred to define the `local` method in order to avoid having to go through the unwrapping process here: https://github.com/emschwartz/interledger-rs/blob/master/crates/interledger-stream/src/server.rs#L67

gakonst:
What do you think about moving the value inside of the struct instead of using a reference in `PrepareBuilder.destination`, `RejectBuilder.triggered_by`? Moving the value makes the code easier to write, but maybe a reference makes more sense, since you shouldn't need to `clone()` the address that you are passing into a struct... I think?

crates/interledger-packet/src/address.rs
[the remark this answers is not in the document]
gakonst: I like the Option approach, let's go with that.

crates/interledger-packet/src/address.rs
gakonst: fixed in https://github.com/emschwartz/interledger-rs/pull/75/commits/fde033c11670b0c232a7d77c870a0b402ea3eb76

crates/interledger-packet/src/address.rs
gakonst: fixed in https://github.com/emschwartz/interledger-rs/pull/75/commits/fde033c11670b0c232a7d77c870a0b402ea3eb76

crates/interledger-packet/src/address.rs
gakonst: Fixed in fixed in https://github.com/emschwartz/interledger-rs/pull/75/commits/fde033c11670b0c232a7d77c870a0b402ea3eb76

crates/interledger-packet/src/address.rs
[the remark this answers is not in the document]
gakonst: fixed in https://github.com/emschwartz/interledger-rs/pull/75/commits/089cc54d156b9c07e9137243cec458fe190163f8

crates/interledger-packet/src/address.rs
gakonst: ref: https://github.com/emschwartz/interledger-rs/pull/75/commits/fde033c11670b0c232a7d77c870a0b402ea3eb76#diff-8c5852cf578f26f54fc6aa011fb65509R156

crates/interledger-packet/src/errors.rs
sentientwaffle: (fyi looks like this TODO was imported from my branch. It's totally unrelated to the rest of this patch).

That is true. My intent was to make the errors more lightweight so that they didn't require allocations. Rather than doing `ParseError::SomeError(format!("some info {} {} {}", ..., ..., ...))` it could have a static message along with some extra fields.

crates/interledger-packet/src/packet.rs
```diff
@@ -761,6 +824,11 @@ mod test_reject {
     fn test_data() {
         assert_eq!(REJECT.data(), fixtures::DATA);
     }
+
+    #[test]
+    fn test_into_data() {
+        assert_eq!(REJECT.clone().into_data(), BytesMut::from(REJECT.data()),);
```
sentientwaffle:
```suggestion
        assert_eq!(REJECT.clone().into_data(), BytesMut::from(REJECT.data()));
```

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,374 @@
...
+        Address::from_str(string).map_err(serde::de::Error::custom)
+    }
+}
+
+impl serde::Serialize for Address {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_newtype_struct("Address", &self.0)
```
sentientwaffle: Does this still serialize to a plan string in e.g. json?

crates/interledger-packet/Cargo.toml
```diff
@@ -5,18 +5,19 @@ authors = ["Evan Schwartz <evan@ripple.com>"]
...
-repository = "https://github.com/emschwartz/interledger-rs"
 
 [dependencies]
 byteorder = "1.3.1"
-bytes = "0.4.12"
+bytes = { version = "0.4.12", features = ["serde"] }
 chrono = "0.4.6"
 hex = "0.3.2"
 quick-error = "1.2.2"
+serde = { version = "1.0", features = ["derive"] }
```
sentientwaffle: `serde` should be an optional dependency; hide the `Serialize`/`Deserialize` `impl`s behind a `cfg` attribute.

crates/interledger-packet/Cargo.toml
```diff
@@ -5,18 +5,19 @@ authors = ["Evan Schwartz <evan@ripple.com>"]
 description = "Interledger packet serialization/deserialization"
 license = "Apache-2.0"
 edition = "2018"
-repository = "https://github.com/emschwartz/interledger-rs"
```
sentientwaffle: Why was this removed?

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        new_address.put_slice(self.0.as_ref());
+        new_address.put(b'.');
+        new_address.put_slice(suffix);
+
+        Address::try_from(new_address.freeze())
+    }
+
+    /// Appends the ILP Address with the provided suffix without a separator.
+    pub fn append(&self, suffix: &[u8]) -> Result<Address, AddressError> {
```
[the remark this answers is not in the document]
gakonst: https://github.com/emschwartz/interledger-rs/blob/master/crates/interledger-stream/src/server.rs#L48-L55

This would be converted to:

`destination_account.with_suffix(...)`
followed by 
`destination_account.append(...)`

crates/interledger-packet/Cargo.toml
gakonst: Done, why should it be optional though?

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,374 @@
...
+        self.0.clone()
+    }
+
+    /// Creates an ILP address without validating the bytes.
+    ///
+    /// # Safety
+    ///
+    /// The given bytes must be a valid ILP address.
+    #[inline]
+    pub unsafe fn new_unchecked(bytes: &[u8]) -> Self {
```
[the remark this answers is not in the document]
gakonst: Can you elaborate how that makes it more obvious to the caller? Happy to make the change nonetheless.

crates/interledger-packet/benches/packets.rs
```diff
@@ -20,7 +22,7 @@ lazy_static! {
             \x74\xe1\x13\x6d\xc7\x1c\x9e\x5f\x28\x3b\xec\x83\x46\x1c\xbf\x12\
             \x61\xc4\x01\x4f\x72\xd4\x8f\x8d\xd6\x54\x53\xa0\xb8\x4e\x7d\xe1\
         ",
-        destination: b"example.alice",
+        destination: Address::from_str("example.alice").unwrap(),
```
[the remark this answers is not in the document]
gakonst: Not sure if we should be adding a method that can panic just for the sake of making tests authoring easier, and besides we can define helpers inside the test that has the same functionality

crates/interledger-packet/src/address.rs
gakonst: Indeed, before it'd serialize as an array of u8s. 

What do you think about https://github.com/emschwartz/interledger-rs/pull/75/commits/e9a0a7315231d0f2b8bc168a37cd931d471f5333?

crates/interledger-packet/benches/packets.rs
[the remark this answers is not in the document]
emschwartz: I think this should use the `new_unchecked` function

crates/interledger-packet/src/address.rs
sentientwaffle: Seems ok, though I'm not sure why it needs the `serde_json` dependency.

crates/interledger-packet/Cargo.toml
sentientwaffle: Making them optional is a common convention I've noticed in many crates.

It keeps the extra impls (and serde itself) from being compiled if the dependent crate doesn't actually need serde.

crates/interledger-packet/src/address.rs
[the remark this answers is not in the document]
sentientwaffle: I see. The method name `append` is misleading here, since `self` isn't modified (and it can't be, since its based on `Bytes`), rather a new deep copy is allocated, modified, and returned.

This is probably not a common case, and could be handled by the caller doing `Bytes::from(address)`, converting to a `BytesMut` (which deep copies), then `.freeze`ing and converting back to an address.

crates/interledger-packet/src/address.rs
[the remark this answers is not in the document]
sentientwaffle: Many useful types deref to `&[u8]`. If you have a `String` and call `Address::new_unchecked(&my_string_address)` it not obvious that `new_unchecked` is creating a copy of `my_string_address`'s bytes. `Address:new_unchecked(Bytes::from(&my_string_address))` is more transparent about the extra copy.

More specifically, methods like `ilp::Prepare::destination` should use `new_unchecked`, since a valid ILP Prepare has a valid destination address. And the `Prepare` has a `Bytes`, so no copies are needed to return the destination address.

crates/interledger-packet/src/address.rs
```diff
@@ -106,18 +106,16 @@ impl AsRef<Bytes> for Address {
...
+            .debug_tuple("Address")
+            .field(&**self.to_string())
+            .finish()
     }
 }
 
 impl fmt::Display for Address {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        unsafe { write!(f, "{}", str::from_utf8_unchecked(&self.0)) }
+        write!(f, "{}", &**self)
```
sentientwaffle: <https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.write_str>

crates/interledger-packet/src/errors.rs
```diff
@@ -37,10 +37,14 @@ quick_error! {
             description(err.description())
             cause(err)
         }
-        // TODO &'static str instead of String?
         InvalidPacket(descr: String) {
             description(descr)
             display("Invalid Packet {}", descr)
         }
+        Other(err: Box<std::error::Error>) {
```
sentientwaffle: Is this variant ever used?

crates/interledger-packet/Cargo.toml
```diff
@@ -22,6 +22,7 @@ lazy_static = "1.3"
 # testing, but optional otherwise.
 serde = { version = "1.0", features = ["derive"]  }
 serde_test = "1.0"
+serde_json = "1.0"
```
sentientwaffle: Why is this needed?

emschwartz:
@sentientwaffle can you comment on the pros and cons of having or not having the additional `Addr` type? It doesn't seem like so much overhead to have a second type if it provides some substantial benefit

crates/interledger-packet/src/errors.rs
gakonst: It was there in master https://github.com/emschwartz/interledger-rs/blob/master/crates/interledger-packet/src/errors.rs#L37-L41

crates/interledger-packet/src/address.rs
gakonst: It was checked in by accident. Removed now.

sentientwaffle:
@emschwartz 

A separate `Addr` type is useful for getting an address from an existing `Bytes` (e.g. `Prepare::destination` or `Reject::triggered_by`).

The alternative -- only 1 `Address` type and using `Bytes::clone` is still fairly cheap (`Bytes::clone` doesn't deep copy the bytes) but does have a little more overhead (an `Arc::clone`). At least in the absence of benchmarks indicating otherwise, that probably doesn't qualify as a "substantial" benefit.

I don't have a strong opinion in favor of one or the other.

gakonst:
> A separate Addr type is useful for getting an address from an existing Bytes (e.g. Prepare::destination or Reject::triggered_by).

My understanding is that the goal is to replace the `destination` and `triggered_by` fields with `Address` and `Option<Address>` (since `triggered_by` can be empty) fields. 

We should avoid using the raw `Bytes` and go for the `Address` type in as many places as possible to ensure that the correct format is enforced

crates/interledger-packet/src/address.rs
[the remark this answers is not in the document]
gakonst: WDYT https://github.com/emschwartz/interledger-rs/pull/75/commits/247cf79bf23b6036fddfe29ee7bb748f47a9cfb7

crates/interledger-packet/src/address.rs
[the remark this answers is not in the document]
gakonst: OK, removed.

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,363 @@
...
+
+impl Address {
+    /// Returns the length of the ILP Address.
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Returns the `Bytes` conversion of the ILP Address
+    pub fn as_bytes(&self) -> Bytes {
```
emschwartz: After looking through https://github.com/emschwartz/interledger-rs/pull/77#discussion_r290397008, I think we should call this `to_bytes`. "as" seems to be used for references or very minor transformations. `Arc::clone` seems to be on the border between the two but I _think_ it's slightly more overhead than calling `as_ref`

gakonst: I agree with this, good point.

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,373 @@
...
+        self.0.clone()
+    }
+
+    /// Creates an ILP address without validating the bytes.
+    ///
+    /// # Safety
+    ///
+    /// The given bytes must be a valid ILP address.
+    #[inline]
+    pub unsafe fn new_unchecked(bytes: Bytes) -> Self {
```
sentientwaffle: Add a `debug_assert` to verify that `bytes` is a valid address. The original `Address` type omitted it so that it could be a `const fn`, but that isn't an issue here.

gakonst: https://github.com/emschwartz/interledger-rs/pull/75/commits/05120f8b8c57a36121738a4d66acba09332bdc53

added a try from with a reference, is that what you had in mind? This will get optimized away in release compilation so it shouldn't be a problem

crates/interledger-packet/src/packet.rs
```diff
@@ -577,24 +645,36 @@ mod test_prepare {
     use super::*;
     use crate::fixtures::{self, PREPARE, PREPARE_BUILDER, PREPARE_BYTES};
 
+    #[test]
+    #[should_panic]
```
sentientwaffle: The purpose of this test is to verify that parsing an ILP Prepare packet that has an invalid `destination` errors. It builds the ILP Prepare using `Address::new_unchecked`, which fails due to the `debug_assert`.

Marking the test as `should_panic` is misleading. The test shouldn't panic.

There are a couple possible solutions:
1. Remove the `debug_assert` so this test works.
2. Instead of constructing a packet, build `with_bad_address`'s bytes by (e.g.) cloning `fixtures::PREPARE_BYTES` and manually one of the `destination`'s bytes to  a space.
3. Remove the test.

crates/interledger-packet/src/address.rs
sentientwaffle: I think `local` is a confusing name for the last segment of the ilp address.

`local.` is an [address scheme](https://github.com/interledger/rfcs/blob/master/0015-ilp-addresses/0015-ilp-addresses.md#allocation-schemes), unrelated to the last segment.

:+1: for the reverse iterator suggested by @emschwartz (maybe `segments_rev`)

crates/interledger-packet/src/packet.rs
gakonst: Thanks for the observation. I preferred option 2, fixed in https://github.com/emschwartz/interledger-rs/pull/75/commits/c21de7d2f484fbaebd9e8a70c8ea556158a4715b

crates/interledger-service-util/src/echo.rs
```diff
@@ -111,21 +113,23 @@ where
...
                     message: b"Could not read source address.",
-                    triggered_by: self.ilp_address.borrow(),
+                    triggered_by: Some(&self.ilp_address),
                     data: &[],
                 }
                 .build()));
             }
         };
 
+        let source_address = Address::try_from(source_address).unwrap();
```
sentientwaffle: Is it safe to `unwrap` here? If the `source_address: &[u8]` is an invalid address this will panic.

crates/interledger-packet/src/address.rs
gakonst: We can use `segments.rev().next()` then.

crates/interledger-ccp/src/packet.rs
```diff
@@ -117,7 +121,7 @@ impl RouteControlRequest {
         }
 
         PrepareBuilder {
-            destination: CCP_CONTROL_DESTINATION,
+            destination: (*CCP_CONTROL_DESTINATION).clone(),
```
sentientwaffle: The deref can be implicit.
```suggestion
            destination: CCP_CONTROL_DESTINATION.clone(),
```

crates/interledger-packet/src/address.rs
```diff
@@ -327,6 +327,25 @@ mod test_address {
         );
     }
 
+    #[test]
+    fn test_segments() {
+        let addr = Address::from_str("test.alice.1234.5789").unwrap();
+        let expected = ["test", "alice", "1234", "5789"];
+        let segments = addr.segments();
+        for (a, b) in segments.zip(&expected) {
```
sentientwaffle: This test may pass even if `segments` and `expected` have different numbers of elements (`zip` short-circuits).

crates/interledger-packet/src/address.rs
```diff
@@ -157,18 +166,9 @@ impl Address {
...
-                .map(|s| str::from_utf8_unchecked(&s))
-        }
-    }
-
-    /// Returns the local part (right-most '.' separated segment) of the ILP Address.
-    pub fn local(&self) -> &[u8] {
-        self.0.rsplit(|&byte| byte == b'.').next().unwrap()
+    /// Returns the last part (right-most '.' separated segment) of the ILP Address.
+    pub fn last(&self) -> &str {
+        self.segments().rev().next().unwrap_or("")
```
sentientwaffle: Why `unwrwap_or("")` instead of `unwrap()`? Addresses should be guaranteed at least one segment (in addition to the scheme).

crates/interledger-service-util/src/echo.rs
gakonst: Yeah I was going to fix it, https://github.com/emschwartz/interledger-rs/pull/75/commits/e56d53d3477571f78a5a3d9977621f7c954764fd is my initial approach, following how the rest of the file handles errors, but I do not really like it. Do you perhaps have any suggestions for that

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+    type Error = ParseError;
+
+    fn try_from(bytes: Bytes) -> Result<Self, Self::Error> {
+        if bytes.len() > MAX_ADDRESS_LENGTH {
+            return Err(ParseError::InvalidAddress(AddressError::InvalidLength(
+                bytes.len(),
+            )));
+        }
+
+        let mut segments = 0;
```
emschwartz: I find this code a bit difficult to follow. It would be helpful to add comments and possibly bring the `is_scheme` and `is_segment_byte` functionality up into this function so it's more obvious what it's actually checking for

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+
+impl TryFrom<&[u8]> for Address {
+    type Error = ParseError;
+
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        Self::try_from(Bytes::from(bytes))
+    }
+}
+
+impl std::ops::Deref for Address {
```
emschwartz: Nit: import this at the top of the file

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+impl AsRef<Bytes> for Address {
+    #[inline]
+    fn as_ref(&self) -> &Bytes {
+        &self.0
+    }
+}
+
+impl fmt::Debug for Address {
```
emschwartz: Nit: import at the top of the file

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        Self::try_from(Bytes::from(bytes))
+    }
+}
+
+impl std::ops::Deref for Address {
+    type Target = str;
+
+    fn deref(&self) -> &str {
+        unsafe { str::from_utf8_unchecked(self.0.as_ref()) }
```
emschwartz: Might be worth adding a comment here that we know this is safe because of the parsing done in `try_from`

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+    ///          "test" / "test1" / "test2" / "test3" / "local"
+    /// ```
+    #[inline]
+    pub fn scheme(&self) -> &str {
+        self.segments().next().unwrap()
+    }
+
+    /// Returns the last part (right-most '.' separated segment) of the ILP Address.
+    /// Addresses are guaranteed at least one segment (in addition to the scheme).
+    pub fn last(&self) -> &str {
```
emschwartz:
```suggestion
    pub fn last_segment(&self) -> &str {
```

It's not clear what `address.last()` would return just from the name.

Also, if this function is only going to be used in a single place in STREAM, I'm not sure it's worth including

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+        self.segments().rev().next().unwrap()
+    }
+
+    /// Suffixes the ILP Address with the provided suffix. Includes a '.' separator
+    pub fn with_suffix(&self, suffix: &[u8]) -> Result<Address, ParseError> {
```
emschwartz: Would it be worth checking to make sure the `suffix` doesn't start with a `.`? I don't think we currently have any use case for addresses with `..` in them, but I'm also not sure we want to ban it.

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+
+        new_address.put_slice(self.0.as_ref());
+        new_address.put(b'.');
+        new_address.put_slice(suffix);
+
+        Address::try_from(new_address.freeze())
+    }
+}
+
+impl<'a> PartialEq<[u8]> for Address {
```
emschwartz: Should it also implement `PartialEq<Address>`?

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+    fn eq(&self, other: &[u8]) -> bool {
+        self.0 == other
+    }
+}
+
+static SCHEMES: &'static [&'static [u8]] = &[
```
emschwartz: Statics and constants should be declared at the top of the file after dependency imports

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+        b"test.alice.XYZ.1234.-_~",
+        b"g.us-fed.ach.0.acmebank.swx0a0.acmecorp.sales.199.~ipr.cdfa5e16-e759-4ba3-88f6-8b9dc83c1868.2",
+
+        b"g.A", b"private.A", b"example.A", b"peer.A", b"self.A",
+        b"test.A", b"test1.A", b"test2.A", b"test3.A", b"local.A",
+    ];
+
+    static INVALID_ADDRESSES: &'static [&'static [u8]] = &[
+        b"", // empty
+        // Invalid characters.
```
emschwartz: Can you also add a non-utf8 character?

crates/interledger-packet/src/error.rs
```diff
@@ -64,15 +65,19 @@ impl From<ErrorCode> for [u8; 3] {
...
 }
 
-impl fmt::Display for ErrorCode {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", str::from_utf8(&self.0[..]).unwrap())
+impl fmt::Debug for ErrorCode {
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter
+            .debug_tuple("ErrorCode")
+            .field(&str::from_utf8(&self.0[..]).map_err(|_| fmt::Error)?)
```
emschwartz: I wonder if we should also include the standard code description here

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+pub enum AddressError {
+    InvalidLength(usize),
+    InvalidFormat,
+}
+
+use std::error::Error;
+impl Error for AddressError {
+    fn description(&self) -> &str {
+        match *self {
+            AddressError::InvalidLength(length) => "invalid address length",
```
emschwartz:
```suggestion
            AddressError::InvalidLength(length) => "Invalid address length",
```

crates/interledger-packet/src/address.rs
```diff
@@ -0,0 +1,385 @@
...
+            AddressError::InvalidFormat => "invalid address format",
```
emschwartz:
```suggestion
            AddressError::InvalidFormat => "Invalid address format",
```

Do you think it would be worth separating `InvalidCharacter` and `InvalidScheme` errors?

crates/interledger-packet/src/packet.rs
```diff
@@ -127,10 +148,10 @@ impl Prepare {
 
         // Skip execution condition.
         content.skip(CONDITION_LEN)?;
-        // Skip the data.
-        // TODO make sure address is only ASCII characters
-        content.skip_var_octet_string()?;
+        // Validate and skip the destination.
+        Address::try_from(content.read_var_octet_string()?)?;
```
emschwartz: If we're parsing this anyway here, we should just include the `Address` as a field on the `Prepare` struct so that we don't need to create it again when `.destination()` is called

crates/interledger-packet/src/packet.rs
```diff
@@ -182,9 +205,10 @@ impl Prepare {
     }
 
     #[inline]
-    pub fn destination(&self) -> &[u8] {
+    pub fn destination(&self) -> Address {
```
emschwartz: This should just clone an `Address` that's saved as part of the `Prepare`. If we have to copy the bytes of the address out of the buffer, we should only do it once

crates/interledger-packet/src/packet.rs
```diff
@@ -360,7 +410,7 @@ impl Reject {
         let code = ErrorCode::new(code);
 
         let triggered_by_offset = content_offset + content_len - content.len();
-        content.skip_var_octet_string()?;
+        Address::try_from(content.read_var_octet_string()?)?;
```
emschwartz: Same as above, if we're doing this copying here, we should store the `Address` on the struct

crates/interledger-stream/src/client.rs
```diff
@@ -177,7 +181,7 @@ where
         // Create the ILP Prepare packet
         let data = stream_packet.into_encrypted(&self.shared_secret);
         let prepare = PrepareBuilder {
-            destination: &self.destination_account[..],
+            destination: Address::try_from(self.destination_account.clone()).unwrap(),
```
emschwartz: The `source_account` and `destination_account` should be `Address`es instead of `Bytes` so we don't need any parsing or unwrapping here

crates/interledger-stream/src/server.rs
```diff
@@ -496,9 +503,11 @@ mod stream_receiver_service {
         let execution_condition = generate_condition(&shared_secret[..], &data);
 
         data.extend_from_slice(b"extra");
+        // destination_account.extend_from_slice(b"extra");
```
```suggestion
```

crates/interledger-packet/src/address.rs
sentientwaffle: Addresses containing `..` are already banned.

> Each segment MUST be separated from other segments by a period character (`.`).

and

> Each segment MUST contain one or more of the following characters: 
> ...

from https://github.com/interledger/rfcs/blob/master/0015-ilp-addresses/0015-ilp-addresses.md#address-requirements

crates/interledger-packet/src/packet.rs
sentientwaffle: I don't think that would work. The `Prepare` wraps a `BytesMut`, and a `BytesMut` can't have a `Bytes` referencing it since it might change under them.

sentientwaffle: Nevermind, it would work, it just allocates. I got mixed up since the original used `Addr` and this wasn't an issue.

crates/interledger-packet/src/address.rs
gakonst: This has been replaced with pattern matching

crates/interledger-packet/src/address.rs
gakonst: Yeah `last_segment` is a good idea. 

If there's no point in adding the function, then we can also remove `scheme` and just use the `segments` iterator to access any segment of the address.

crates/interledger-packet/src/address.rs
gakonst: This variable is no longer present

crates/interledger-packet/src/address.rs
gakonst: Added a test. We already derive PartialEq

crates/interledger-packet/src/error.rs
gakonst: Not sure what you mean?

crates/interledger-packet/src/address.rs
gakonst: If the address is malformed, the InvalidFormat error is sufficient for debugging I think. 
I doubt that more granularity is needed

crates/interledger-packet/src/address.rs
emschwartz: I think there aren't so many places we need to access it so removing the methods seems fine

crates/interledger-store-redis/src/account.rs
```diff
@@ -121,7 +124,7 @@ impl ToRedisArgs for Account {
         self.id.write_redis_args(&mut rv);
         if !self.ilp_address.is_empty() {
             "ilp_address".write_redis_args(&mut rv);
-            rv.push(self.ilp_address.to_vec());
+            rv.push(self.ilp_address.to_bytes().to_vec()); // TODO: Can we go directly to Vec<u8>?
```
emschwartz: This avoids the `Bytes.clone()` but I'm not sure how much that really makes a difference since we're copying the address bytes anyway
```suggestion
            rv.push((self.ilp_address.as_ref() as &[u8]).to_vec());
```

crates/interledger-store-redis/src/account.rs
```diff
@@ -191,6 +194,8 @@ impl FromRedisValue for Account {
     fn from_redis_value(v: &Value) -> Result<Self, RedisError> {
         let hash: HashMap<String, Value> = HashMap::from_redis_value(v)?;
         let ilp_address: String = get_value("ilp_address", &hash)?;
+        let ilp_address = Address::from_str(&ilp_address)
+            .map_err(|_| RedisError::from((ErrorKind::TypeError, "invalid ILP address")))?;
```
emschwartz:
```suggestion
            .map_err(|_| RedisError::from((ErrorKind::TypeError, "Invalid ILP address")))?;
```

crates/interledger-stream/src/client.rs
```diff
@@ -113,7 +112,8 @@ where
             })];
             if self.should_send_source_account {
                 frames.push(Frame::ConnectionNewAddress(ConnectionNewAddressFrame {
-                    source_account: &self.source_account[..],
+                    // How can we make this take a reference?
```
emschwartz: `Address` is _basically_ a reference so I wouldn't worry about it too much

crates/interledger-stream/src/server.rs
```diff
@@ -40,48 +41,46 @@ impl ConnectionGenerator {
...
-            base64::URL_SAFE_NO_PAD,
-        ));
-        let auth_tag = &hmac_sha256(&shared_secret[..], &destination_account[..])[..14];
-        destination_account.put(base64::encode_config(auth_tag, base64::URL_SAFE_NO_PAD));
-        debug!(
-            "Generated address: {}",
-            str::from_utf8(&destination_account[..]).unwrap_or("<not utf8>"),
-        );
-        (destination_account.freeze(), shared_secret)
+        let destination_account = base_address
```
```suggestion
        // Note that the unwrap here is safe because we know the base_address
        // is valid and adding base64-url characters will always be valid
        let destination_account = base_address
```

crates/interledger-service-util/src/echo.rs
gakonst: The first check actually just reads it from the buffer, but the actual checking that it conforms to the spec is not done until the second try_from, so I do not think it's unnecessary

crates/interledger-store-memory/src/store.rs
```diff
@@ -172,16 +174,17 @@ impl BtpOpenSignupStore for InMemoryStore {
...
             .id(account_id)
-            .ilp_address(account.ilp_address)
             .btp_incoming_token(account.auth_token.to_string())
             .asset_code(account.asset_code.to_string())
             .asset_scale(account.asset_scale)
             .build();
 
         (*self.accounts.write()).insert(account_id, account.clone());
-        (*self.routing_table.write()).insert(account.inner.ilp_address.clone(), account_id);
+        // Can we avoid the clone to get out of the Arc?
```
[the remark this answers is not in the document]
gakonst:
```suggestion
```

crates/interledger-service-util/src/echo.rs
emschwartz: Ah, sorry. Misread the match statement. If there's an error here it should return an F01 error just like the above check. Instead of using the matches

crates/interledger-stream/src/client.rs
gakonst:
```suggestion
```

crates/interledger-service-util/src/echo.rs
gakonst:
```rust
        let source_address = match reader.read_var_octet_string() {
            Ok(value) => Address::try_from(value).map_err(|_| {
                Box::new(err(RejectBuilder {
                    code: ErrorCode::F01_INVALID_PACKET,
                    message: b"Could not parse source address from Echo packet",
                    triggered_by: Some(&self.ilp_address),
                    data: &[],
                }
                .build()))
            })?,
            Err(error) => {
                eprintln!("Could not read source address: {:?}", error);
                return Box::new(err(RejectBuilder {
                    code: ErrorCode::F01_INVALID_PACKET,
                    message: b"Could not read source address.",
                    triggered_by: Some(&self.ilp_address),
                    data: &[],
                }
                .build()));
            }
        };
```

This was what I had done initially, but the question mark operator doesn't behave well with futures it seems

```
rror[E0277]: the `?` operator can only be used in a function that returns `Result` or `Option` (or another type that implements `std::ops::Try`)
   --> crates/interledger-service-util/src/echo.rs:110:26
    |
110 |               Ok(value) => Address::try_from(value).map_err(|_| {
    |  __________________________^
111 | |                 Box::new(err(RejectBuilder {
112 | |                     code: ErrorCode::F01_INVALID_PACKET,
113 | |                     message: b"Could not parse source address from Echo packet",
...   |
117 | |                 .build()))
118 | |             })?,
    | |_______________^ cannot use the `?` operator in a function that returns `std::boxed::Box<dyn futures::future::Future<Error = interledger_packet::packet::Reject, Item = interledger_packet::packet::Fulfill> + std::marker::Send>`
    |
    = help: the trait `std::ops::Try` is not implemented for `std::boxed::Box<dyn futures::future::Future<Error = interledger_packet::packet::Reject, Item = interledger_packet::packet::Fulfill> + std::marker::Send>`
    = note: required by `std::ops::Try::from_error`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: Could not compile `interledger-service-util`.
```

emschwartz: Right. Ahh the pleasure of working with Future types. Looking forward to async await

PR #77, opened 2019-06-03
Use IlpAddress in rest of codebase and fix tests

gakonst:
Stacked PR on top of #75, should be merged on it, and then we merge to master.

### DIFF
--- a/crates/interledger-ccp/src/packet.rs
+++ b/crates/interledger-ccp/src/packet.rs
@@ -30,9 +30,9 @@ lazy_static! {
     }
     .build();
     pub static ref CCP_CONTROL_DESTINATION: Address =
-        unsafe { Address::new_unchecked(b"peer.route.control") };
+        unsafe { Address::new_unchecked(Bytes::from("peer.route.control")) };
     pub static ref CCP_UPDATE_DESTINATION: Address =
-        unsafe { Address::new_unchecked(b"peer.route.update") };
+        unsafe { Address::new_unchecked(Bytes::from("peer.route.update")) };
 }
 
 #[derive(Clone, Copy, PartialEq, Debug)]
--- a/crates/interledger-ccp/src/packet.rs
+++ b/crates/interledger-ccp/src/packet.rs
@@ -9,6 +9,7 @@ use std::{
     convert::TryFrom,
     io::Read,
     str,
+    str::FromStr,
     time::{Duration, SystemTime},
 };
 
@@ -29,8 +30,8 @@ lazy_static! {
         data: &[],
     }
     .build();
-    pub static ref CCP_CONTROL_DESTINATION: Address = Address::from_str("peer.route.control").unwrap()
-    pub static ref CCP_UPDATE_DESTINATION: Address = Address::from_str("peer.route.update").unwrap()
+    pub static ref CCP_CONTROL_DESTINATION: Address = Address::from_str("peer.route.control").unwrap();
+    pub static ref CCP_UPDATE_DESTINATION: Address = Address::from_str("peer.route.update").unwrap();
 }
 
 #[derive(Clone, Copy, PartialEq, Debug)]
--- a/crates/interledger-ccp/src/packet.rs
+++ b/crates/interledger-ccp/src/packet.rs
@@ -12,8 +12,6 @@ use std::{
     time::{Duration, SystemTime},
 };
 
-pub const CCP_CONTROL_DESTINATION: &[u8] = b"peer.route.control";
-pub const CCP_UPDATE_DESTINATION: &[u8] = b"peer.route.update";
 // pub const PEER_PROTOCOL_FULFILLMENT: [u8; 32] = [0; 32];
 pub const PEER_PROTOCOL_CONDITION: [u8; 32] = [
     102, 104, 122, 173, 248, 98, 189, 119, 108, 143, 193, 139, 142, 159, 142, 32, 8, 151, 20, 133,
@@ -31,6 +29,10 @@ lazy_static! {
         data: &[],
     }
     .build();
+    pub static ref CCP_CONTROL_DESTINATION: Address =
+        unsafe { Address::new_unchecked(b"peer.route.control") };
+    pub static ref CCP_UPDATE_DESTINATION: Address =
+        unsafe { Address::new_unchecked(b"peer.route.update") };
 }
 
 #[derive(Clone, Copy, PartialEq, Debug)]
@@ -118,7 +120,7 @@ impl RouteControlRequest {
         }
 
         PrepareBuilder {
-            destination: Address::try_from(CCP_CONTROL_DESTINATION).unwrap(),
+            destination: (*CCP_CONTROL_DESTINATION).clone(),
             amount: 0,
             expires_at: SystemTime::now() + Duration::from_millis(PEER_PROTOCOL_EXPIRY_DURATION),
             execution_condition: &PEER_PROTOCOL_CONDITION,
@@ -321,7 +323,7 @@ impl RouteUpdateRequest {
         }
 
         PrepareBuilder {
-            destination: Address::try_from(CCP_UPDATE_DESTINATION).unwrap(),
+            destination: (*CCP_UPDATE_DESTINATION).clone(),
             amount: 0,
             expires_at: SystemTime::now() + Duration::from_millis(PEER_PROTOCOL_EXPIRY_DURATION),
             execution_condition: &PEER_PROTOCOL_CONDITION,
--- a/crates/interledger-ccp/src/packet.rs
+++ b/crates/interledger-ccp/src/packet.rs
@@ -3,9 +3,10 @@ use bytes::{BufMut, Bytes};
 use hex;
 use interledger_packet::{
     oer::{BufOerExt, MutBufOerExt},
-    Fulfill, FulfillBuilder, ParseError, Prepare, PrepareBuilder,
+    Address, Fulfill, FulfillBuilder, ParseError, Prepare, PrepareBuilder,
 };
 use std::{
+    convert::TryFrom,
     io::Read,
     str,
     time::{Duration, SystemTime},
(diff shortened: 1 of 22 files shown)


gakonst:
There's a few places where I do: `Address::try_from(CCP_UPDATE_DESTINATION).unwrap()`, where `CCP_UPDATE_DESTINATION` is a static. Can we somehow move the conversion to the declaration of the variable?

crates/interledger-stream/src/server.rs
```diff
@@ -167,13 +170,14 @@ fn receive_money(
...
     let prepare_amount = prepare.amount();
+    let client_addr = Address::try_from(client_address).ok();
     let stream_packet =
         StreamPacket::from_encrypted(shared_secret, prepare.into_data()).map_err(|_| {
             debug!("Unable to parse data, rejecting Prepare packet");
             RejectBuilder {
                 code: ErrorCode::F06_UNEXPECTED_PAYMENT,
                 message: b"Could not decrypt data",
-                triggered_by: client_address,
+                triggered_by: client_addr.clone(),
```
emschwartz: To avoid needing to clone it before passing it in, we could make `RejectBuilder` take an `&Address` and have the builder to do the clone (although it might not even need to do that)

emschwartz:
> Do you think it's worth it to create a const function to convert from str to Address for such occasions (this would require adding the const_fn feature)?

I would prefer to keep using stable instead of nightly. For now, I think I would recommend using [`lazy_static`](https://docs.rs/lazy_static/1.3.0/lazy_static/) and the `new_unchecked` method instead of a const fn

crates/interledger-router/src/router.rs
```diff
@@ -36,16 +36,15 @@ where
...
 
     fn handle_request(&mut self, request: IncomingRequest<T::Account>) -> Self::Future {
-        let destination = Bytes::from(request.prepare.destination());
+        let destination = request.prepare.destination();
         let mut next_hop: Option<<T::Account as Account>::AccountId> = None;
         let routing_table = self.store.routing_table();
 
         // Check if we have a direct path for that account or if we need to scan through the routing table
-        if let Some(account_id) = routing_table.get(&destination) {
+        if let Some(account_id) = routing_table.get(&destination.as_bytes()) {
```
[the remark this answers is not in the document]
gakonst: Yeah so my plan was to minimize any interface changes for this PR, to get tests to pass. Then, make another PR which starts changing all function interfaces to use `Address` instead of the found `&[u8]`, `Vec<u8>`, `String, or `Bytes`.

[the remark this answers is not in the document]
emschwartz: I would include those changes in the same PR but maybe in a different commit

crates/interledger-stream/src/server.rs
gakonst: https://github.com/emschwartz/interledger-rs/pull/77/commits/de8dd8dc42b1c3bf5c1b3eee0ee7f747d0d1c636

crates/interledger-packet/src/address.rs
```diff
@@ -163,7 +173,7 @@ impl Address {
     }
 
     /// Suffixes the ILP Address with the provided suffix. Includes a '.' separator
-    pub fn with_suffix(&self, suffix: &[u8]) -> Result<Address, AddressError> {
+    pub fn with_suffix(&self, suffix: &[u8]) -> Result<Address, ParseError> {
```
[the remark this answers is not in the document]
gakonst: https://github.com/emschwartz/interledger-rs/pull/77/commits/4f74cf1f086f15d3f85cfa2f44263338f3ca8765

crates/interledger-ccp/src/server.rs
```diff
@@ -195,15 +197,16 @@ where
             };
 
             if !self.spawn_tasks {
-                let ilp_address = self.ilp_address.clone();
+                // TODO: Can we safely comment this out?
```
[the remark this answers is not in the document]
gakonst: https://github.com/emschwartz/interledger-rs/pull/77/commits/4f74cf1f086f15d3f85cfa2f44263338f3ca8765

crates/interledger-ccp/src/packet.rs
```diff
@@ -117,7 +118,7 @@ impl RouteControlRequest {
         }
 
         PrepareBuilder {
-            destination: CCP_CONTROL_DESTINATION,
+            destination: Address::try_from(CCP_CONTROL_DESTINATION).unwrap(),
```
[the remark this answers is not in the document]
gakonst: https://github.com/emschwartz/interledger-rs/pull/77/commits/4f74cf1f086f15d3f85cfa2f44263338f3ca8765

crates/interledger-ccp/src/packet.rs
```diff
@@ -30,6 +29,10 @@ lazy_static! {
         data: &[],
     }
     .build();
+    pub static ref CCP_CONTROL_DESTINATION: Address =
+        unsafe { Address::new_unchecked(Bytes::from("peer.route.control")) };
```
emschwartz: Nitpick but I think the way you have it in the tests with `Address::from_str(...).unwrap()` is the slightly more common way of doing this

crates/interledger-ccp/src/server.rs
```diff
@@ -252,11 +253,12 @@ where
     /// then check whether those routes are better than the current best ones we have in the
     /// Local Routing Table.
     fn handle_route_update_request(&self, request: IncomingRequest<A>) -> BoxedIlpFuture {
+        let ilp_address = Address::try_from(self.ilp_address.clone()).ok();
```
emschwartz: This shouldn't be parsed in the handler. The CCP Manager should be instantiated with an `Address` instead of `Bytes`

crates/interledger-ccp/src/server.rs
```diff
@@ -300,15 +302,16 @@ where
                     spawn(future);
                     Box::new(ok(CCP_RESPONSE.clone()))
                 } else {
-                    let ilp_address = self.ilp_address.clone();
+                    // Can we safely remove this re-setting of the ilp_address?
+                    // let ilp_address = self.ilp_address.clone();
```
emschwartz: Should remove these comments

crates/interledger-ccp/src/test_helpers.rs
```diff
@@ -7,12 +7,13 @@ use futures::{
...
+use interledger_packet::{Address, ErrorCode, RejectBuilder};
 use interledger_service::{
     incoming_service_fn, outgoing_service_fn, BoxedIlpFuture, IncomingService, OutgoingRequest,
     OutgoingService,
 };
 use parking_lot::Mutex;
+use std::str::FromStr;
 use std::{iter::FromIterator, sync::Arc};
 
 lazy_static! {
```
emschwartz: I think these test accounts should use `Address` too

crates/interledger-packet/src/address.rs
```diff
@@ -39,19 +43,21 @@ impl fmt::Display for AddressError {
 pub struct Address(Bytes);
 
 impl FromStr for Address {
-    type Err = AddressError;
+    type Err = ParseError;
```
emschwartz: Why isn't this an `AddressError`?

crates/interledger-packet/src/address.rs
```diff
@@ -69,13 +75,13 @@ impl TryFrom<Bytes> for Address {
         if is_valid_scheme && segments > 1 {
             Ok(Address(bytes))
         } else {
-            Err(AddressError::InvalidFormat)
+            Err(ParseError::InvalidAddress(AddressError::InvalidFormat))
```
emschwartz: Any reason not to include a more specific description of what's wrong with it?

crates/interledger-router/src/router.rs
```diff
@@ -57,7 +56,7 @@ where
                     route.1
                 );
                 // Check if the route prefix matches or is empty (meaning it's a catch-all address)
-                if (route.0.is_empty() || destination.starts_with(&route.0[..]))
+                if (route.0.is_empty() || destination.as_bytes().starts_with(&route.0[..]))
```
emschwartz: Using `as_ref` would work too and wouldn't trigger an `Arc::clone`

crates/interledger-stream/src/server.rs
```diff
@@ -44,6 +46,7 @@ impl ConnectionGenerator {
         let random_bytes = generate_token();
         // base_address + "." + 32-bytes encoded as base64url
         let shared_secret = hmac_sha256(&self.secret_generator[..], &random_bytes[..]);
+        // TODO: Use Address field here.
```
emschwartz: Any reason not to change this now?

crates/interledger-stream/src/server.rs
```diff
@@ -134,11 +137,12 @@ where
         if request
             .prepare
             .destination()
+            .as_bytes()
```
emschwartz: I think this should use `as_ref`

crates/interledger-packet/src/address.rs
gakonst: My thinking was that following the style in the rest of the codebase, the error everywhere is

`ParseError::InvalidAddress` from `errors.rs` which itself takes an `AddressError`. 
We could perhaps make all `InvalidAddress` take a `String` (similar to how `WrongType` is declared), and then specify the error string

crates/interledger-packet/src/address.rs
gakonst: Not sure what further information we should give here. Could maybe include the invalid address that was passed in, and some regex or reference to the expected format?

crates/interledger-packet/src/address.rs
sentientwaffle: I agree with evan, this should be an `AddressError`.

`ParseError::AddressError` indicates that parsing a packet failed due to a problem parsing an address.
The `AddressError` type indicates a problem parsing an address.

PR #83, opened 2019-06-07
Use IlpAddress in rest of codebase

gakonst:
Final refactor of the rest of the codebase to support `Address`.


crates/interledger-stream/src/packet.rs
```diff
@@ -429,28 +429,30 @@ impl<'a> SerializableFrame<'a> for ConnectionCloseFrame<'a> {
 }
 
 #[derive(PartialEq, Clone)]
-pub struct ConnectionNewAddressFrame<'a> {
-    pub source_account: &'a [u8],
+pub struct ConnectionNewAddressFrame {
+    pub source_account: Address,
```
gakonst: Wonder if this should be made back into a reference

crates/interledger/src/cli.rs
```diff
@@ -25,11 +25,16 @@ use interledger_store_redis::{connect as connect_redis_store, IntoConnectionInfo
...
 use ring::rand::{SecureRandom, SystemRandom};
-use std::{convert::TryFrom, net::SocketAddr, str, sync::Arc, u64};
+use std::{convert::TryFrom, net::SocketAddr, str, str::FromStr, sync::Arc, u64};
 use tokio::{self, net::TcpListener};
 use tower_web::ServiceBuilder;
 use url::Url;
 
+lazy_static! {
+    pub static ref LOCAL_ILP_ADDRESS: Address = 
+        Address::from_str("local.host").unwrap();
```
gakonst: Placeholder address as discussed with Evan. We'll have to make it so that an account builder always takes an ilp address as argument, or makes an ILDCP request to get it from a connected node

crates/interledger-api/src/lib.rs
```diff
@@ -60,6 +60,9 @@ pub trait NodeStore: Clone + Send + Sync + 'static {
 /// The Account type for the RedisStore.
 #[derive(Debug, Extract, Response, Clone)]
 pub struct AccountDetails {
+    // Error when converting this to Option<Address>
+    // `_IMPL_SERIALIZE_FOR_AccountsResponse::_serde::Serialize`
+    // is not implemented for `interledger_packet::address::Address`
```
emschwartz: That might be because you need to include `features = ["serde"]` when importing the `interledger-packet` module in `Cargo.toml`

crates/interledger-ildcp/src/packet.rs
```diff
@@ -94,8 +95,9 @@ impl IldcpResponse {
         })
     }
 
-    pub fn client_address(&self) -> &[u8] {
-        (&self.buffer[..]).peek_var_octet_string().unwrap()
+    pub fn client_address(&self) -> Result<Address, ParseError> {
```
emschwartz: Instead of returning a `Result` here, it should parse it when `try_from` is called and error if it is invalid

crates/interledger-btp/src/server.rs
```diff
@@ -171,24 +171,21 @@ where
...
-                    ))
+                    )
                 };
-                let mut ilp_address = BytesMut::with_capacity(
-                    ildcp_info.client_address().len() + 1 + local_part.len(),
-                );
-                ilp_address.put(ildcp_info.client_address());
-                ilp_address.put(&b"."[..]);
-                ilp_address.put(local_part);
+                let ilp_address = ildcp_info.client_address().unwrap();
```
emschwartz: There shouldn't be an unwrap here

crates/interledger-ccp/src/server.rs
```diff
@@ -109,13 +110,10 @@ where
...
+        let ilp_address = account.client_address().clone();
         // The global prefix is the first part of the address (for example "g." for the global address space, "example", "test", etc)
-        let ilp_address = Bytes::from(account.client_address());
-        let global_prefix: Bytes = ilp_address
-            .iter()
-            .position(|c| c == &b'.')
-            .map(|index| ilp_address.slice_to(index + 1))
-            .unwrap_or_else(|| ilp_address.clone());
+        let mut global_prefix: Bytes = ilp_address.scheme().into();
+        global_prefix.extend(b"."); // append a separator with the scheme for the router
```
emschwartz: Didn't we say we weren't going to have both `extend` and `with_suffix`? It's unclear what the difference between those would be from the names

crates/interledger-ildcp/src/server.rs
```diff
@@ -35,15 +35,13 @@ where
...
-                client_address: &request.from.client_address(),
+                client_address: &from,
                 asset_code: request.from.asset_code(),
                 asset_scale: request.from.asset_scale(),
             };
-            debug!(
-                "Responding to query for ILDCP info by account: {:?}",
-                str::from_utf8(&request.from.client_address()[..]).unwrap_or("<not utf8>")
-            );
+            debug!("responding to query for ildcp info by account: {:?}", from);
```
emschwartz: Ahh so much cleaner! :smile_cat:

crates/interledger-packet/src/address.rs
```diff
@@ -54,25 +55,18 @@ impl TryFrom<Bytes> for Address {
...
-                segments += 1;
-                let scheme_ok = i != 0 || is_scheme(segment);
-                scheme_ok
-                    && !segment.is_empty()
-                    && segment.iter().all(|&byte| is_segment_byte(byte))
-            });
-
-        if is_valid_scheme && segments > 1 {
+        let pattern =
+            Regex::new(r"^(g|private|example|peer|self|test[1-3]?|local)([.][a-zA-Z0-9_~-]+)+$")
```
emschwartz: This pattern should be put at the top of the file in a `lazy_static` block to avoid recompiling the Regex each time it's used

crates/interledger-packet/src/address.rs
```diff
@@ -54,25 +55,18 @@ impl TryFrom<Bytes> for Address {
...
+                .unwrap();
+
+        if pattern.is_match(str::from_utf8(&bytes).unwrap()) {
```
emschwartz: Why is this being unwrapped? It will panic if the bytes are not valid UTF8 and we have no guarantee that they are

crates/interledger-spsp/src/client.rs
```diff
@@ -41,7 +43,7 @@ where
         send_money(
             service,
             &from_account,
-            spsp.destination_account.as_bytes(),
+            Address::from_str(&spsp.destination_account).unwrap(), // todo: Figure out how to handle error
```
emschwartz: You can separate parsing this into another `.and_then` block where you map the SPSP response to `(destination_account, shared_secret)` and return an error if it doesn't parse

crates/interledger-spsp/src/server.rs
```diff
@@ -26,9 +27,12 @@ impl SpspResponder {
     pub fn generate_http_response(&self) -> Response<Body> {
         let (destination_account, shared_secret) = self
             .connection_generator
-            .generate_address_and_secret(&self.ilp_address[..]);
-        let destination_account = String::from_utf8(destination_account.to_vec()).unwrap();
-        debug!("Generated address and secret for: {}", destination_account);
+            .generate_address_and_secret(&self.ilp_address);
+        let destination_account = destination_account.to_string();
```
emschwartz: Nitpick: I don't think this needs to be turned into a String. The `SpspResponse` should be changed to be an `Address` instead of a `String`

crates/interledger-store-redis/src/account.rs
```diff
@@ -121,7 +125,7 @@ impl ToRedisArgs for Account {
         self.id.write_redis_args(&mut rv);
         if !self.ilp_address.is_empty() {
             "ilp_address".write_redis_args(&mut rv);
-            rv.push(self.ilp_address.to_vec());
+            rv.push(self.ilp_address.to_bytes().to_vec()); // TODO: Can we go directly to Vec<u8>?
```
emschwartz: I think you can use `as_ref` instead (because [`ToRedisArgs`](https://docs.rs/redis/0.11.0-beta.1/redis/trait.ToRedisArgs.html) is implemented for slices)

crates/interledger-store-redis/src/store.rs
```diff
@@ -496,7 +498,7 @@ impl NodeStore for RedisStore {
                     }
 
                     // Add route to routing table
-                    pipe.hset(ROUTES_KEY, account.ilp_address.to_vec(), account.id)
+                    pipe.hset(ROUTES_KEY, account.ilp_address.to_bytes().to_vec(), account.id) // TODO: Can we go directly to Vec<u8>?
```
emschwartz: Again, should be able to use `as_ref`

crates/interledger-stream/src/client.rs
```diff
@@ -43,8 +42,8 @@ where
             state: SendMoneyFutureState::SendMoney,
             next: Some(service),
             from_account,
-            source_account: Bytes::from(account_details.client_address()),
-            destination_account,
+            source_account: account_details.client_address().unwrap(),
+            destination_account: destination_account,
```
emschwartz: Nit:
```suggestion
            destination_account,
```

crates/interledger-stream/src/client.rs
```diff
@@ -115,7 +114,8 @@ where
             })];
             if self.should_send_source_account {
                 frames.push(Frame::ConnectionNewAddress(ConnectionNewAddressFrame {
-                    source_account: &self.source_account[..],
+                    // How can we make this take a reference?
```
emschwartz: What do you mean? Can't it just take `&Address` instead?

crates/interledger-stream/src/packet.rs
emschwartz: Seems like it would be fine to do that

crates/interledger-stream/src/server.rs
```diff
@@ -42,31 +42,32 @@ impl ConnectionGenerator {
...
-        debug!(
-            "Generated address: {}",
-            str::from_utf8(&destination_account[..]).unwrap_or("<not utf8>"),
-        );
-        (destination_account.freeze(), shared_secret)
+        let destination_account = base_address
+            .with_suffix(
+                &base64::encode_config(&random_bytes[..], base64::URL_SAFE_NO_PAD).as_ref(),
+            )
+            .unwrap(); // How should we handle an invalid suffix here? Can we assume this call will never fail?
```
emschwartz: If we are creating the suffix, we can assume it will never fail. Base64-url characters are all valid in an address segment

crates/interledger-stream/src/server.rs
```diff
@@ -42,31 +42,32 @@ impl ConnectionGenerator {
...
+        debug!("Generated address: {}", destination_account,);
+        (destination_account, shared_secret)
     }
 
     /// Rederive the `shared_secret` from a `destination_account`. This will return an
     /// error if the address has been modified in any way or if the packet was not generated
     /// with the same server secret.
-    pub fn rederive_secret(&self, destination_account: &[u8]) -> Result<[u8; 32], ()> {
+    pub fn rederive_secret(&self, destination_account: &Address) -> Result<[u8; 32], ()> {
+        let destination_account = destination_account.to_bytes();
```
emschwartz: Does this need to be a `Bytes`?

crates/interledger-stream/src/server.rs
```diff
@@ -136,17 +137,14 @@ where
     fn send_request(&mut self, request: OutgoingRequest<A>) -> Self::Future {
         let dest = request.prepare.destination();
         let dest: &[u8] = dest.as_ref();
-        if dest.as_ref().starts_with(request.to.client_address()) {
+        let to = request.to.client_address();
+        if dest.as_ref().starts_with(to.as_ref()) {
             if let Ok(shared_secret) = self
                 .connection_generator
-                .rederive_secret(request.prepare.destination().as_ref())
+                .rederive_secret(&request.prepare.destination())
```
emschwartz: Shouldn't that method return a reference?

crates/interledger/src/cli.rs
```diff
@@ -25,11 +25,15 @@ use interledger_store_redis::{connect as connect_redis_store, IntoConnectionInfo
...
 use parking_lot::RwLock;
 use ring::rand::{SecureRandom, SystemRandom};
-use std::{convert::TryFrom, net::SocketAddr, str, sync::Arc, u64};
+use std::{convert::TryFrom, net::SocketAddr, str, str::FromStr, sync::Arc, u64};
 use tokio::{self, net::TcpListener};
 use tower_web::ServiceBuilder;
 use url::Url;
 
+lazy_static! {
+    pub static ref LOCAL_ILP_ADDRESS: Address = Address::from_str("local.host").unwrap();
```
emschwartz: :grin:

crates/interledger/src/cli.rs
```diff
@@ -172,7 +176,8 @@ pub fn run_spsp_server_btp(
 ) -> impl Future<Item = (), Error = ()> {
     debug!("Starting SPSP server");
     let ilp_address = Arc::new(RwLock::new(Bytes::new()));
-    let incoming_account: Account = AccountBuilder::new()
+    // TODO Make this happen in opposite order so that ilp address is obtained otherwise
```
emschwartz: What does that mean?

crates/interledger-ccp/src/server.rs
gakonst: `extend` is a method of `Bytes`. The method which was suffixing without a separator has been removed from `Address`.

crates/interledger-ildcp/src/packet.rs
gakonst: What do you mean? Always return an `Address` and panic if not? We could do `try_from(..).expect('...')` but not sure if that's desirable

crates/interledger-btp/src/server.rs
gakonst: It is there due to https://github.com/emschwartz/interledger-rs/pull/83#discussion_r292792738

crates/interledger-store-redis/src/account.rs
gakonst: I had done that previously, however runs into some overflows, which weren't solved even after setting `#![recursion_limit=SOME_HIGH_VALUE]`

ref:https://pastebin.com/2mvqGw6w

crates/interledger-store-redis/src/store.rs
gakonst: as https://github.com/emschwartz/interledger-rs/pull/83#discussion_r292816541

crates/interledger-stream/src/server.rs
gakonst: Since `Address` owns the `Bytes`, the function here https://github.com/emschwartz/interledger-rs/blob/use-address-everywhere/crates/interledger-packet/src/packet.rs#L210-L211 cannot return a reference as it would return a reference to data owned by the current function

crates/interledger-stream/src/client.rs
gakonst: I attempted making the `ConnectionNewAddressFrame` accept a reference to `Address`, as shown in this branch:
https://github.com/emschwartz/interledger-rs/commit/925c79d845c21302bb3d05715432d39db8b894da

However, the `try_from` in https://github.com/emschwartz/interledger-rs/blob/use-address-everywhere/crates/interledger-stream/src/packet.rs#L439-L441, makes `read_contents` now own `source_account.` As a result, returning it by reference is not possible since we'd be "returning a value referencing data owned by the current function". 

A solution to this would be to attempt to make `Address` accept `&Bytes` inside of it (and use TryFrom<&Bytes> to avoid taking ownership), but it runs into [similar problems](https://github.com/emschwartz/interledger-rs/commit/e4ebae5451c9e166140e28b9989e388be6b3a048).  

Generally, this situation may arise anytime a function:
1. takes a buffer as an argument
2. extracts a few bytes (takes ownership) 
3. and then creates an address out of these bytes

so maybe we could leave the NewAddressFrame as is, without making it take a reference?

gakonst:
@sentientwaffle would be great if you could also take a look

crates/interledger-api/src/lib.rs
sentientwaffle: Can this comment be removed now?

crates/interledger-ildcp/src/packet.rs
```diff
@@ -113,8 +118,8 @@ impl fmt::Debug for IldcpResponse {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "IldcpResponse {{ client_address: \"{}\", asset_code: \"{}\", asset_scale: {} }}",
-            str::from_utf8(self.client_address()).unwrap_or("<not utf8>"),
+            "IldcpResponse {{ client_address: \"{:?}\", asset_code: \"{}\", asset_scale: {} }}",
```
sentientwaffle: Why `Debug` instead of `Display`?

PR #91, opened 2019-06-14
[donotmerge] testing if CI will show same linking error

gakonst:


emschwartz:
Looks like the answer is no

gakonst:
Yeah pretty odd. I'll need to check if the problem started appearing in my branch, or if CI can't build the HEAD at `es-settlement`

PR #96, opened 2019-06-17
Propagate packet to next peer instead of waiting for store operation to finish

### LINKED ISSUE #92: Directly propagate packet instead of waiting for store to update
It's more important to propagate the reject to the peer ASAP than wait for the disk write to complete and then pass the packet. 

As a result, we can `spawn` instead of `and_then`

Based on discussion from: 
https://github.com/emschwartz/interledger-rs/pull/90/files#r293530578
https://github.com/emschwartz/interledger-rs/pull/90/files#r293530234

> A comment from @cjimison also made me realize another reason to send the fulfill on even if we aren't able to save to the DB: if our database connection is dropped, we shouldn't wait until it's reconnected to pass on the fulfillment since we're already on the hook for the money

gakonst:
Fixes #92. The log in case the store operation failed should be moved in the respective implementation of each store.

### DIFF
--- a/crates/interledger-service-util/src/balance_service.rs
+++ b/crates/interledger-service-util/src/balance_service.rs
@@ -4,6 +4,7 @@ use interledger_ildcp::IldcpAccount;
 use interledger_packet::{ErrorCode, Fulfill, Reject, RejectBuilder};
 use interledger_service::*;
 use std::marker::PhantomData;
+use tokio_executor::spawn;
 
 pub trait BalanceStore: AccountStore {
     /// Fetch the current balance for the given account.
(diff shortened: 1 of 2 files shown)


crates/interledger-service-util/src/balance_service.rs
```diff
@@ -105,30 +111,31 @@ where
...
-                            // TODO should we spawn a task to update the balances instead of doing it before returning the fulfill?
-                            // do we always return the fulfill, even if the state update fails? If so, I think we can.
-                            store.update_balances_for_fulfill(from.clone(), incoming_amount, to.clone(), outgoing_amount)
-                                .then(move |result| {
-                                    if result.is_err() {
-                                        error!("Error applying balance changes for fulfill from account: {} to account: {}. Incoming amount was: {}, outgoing amount was: {}", from.id(), to.id(), incoming_amount, outgoing_amount);
-                                    }
-                                    Ok(fulfill)
-                                })
+                            let fulfill_balance_update = store.update_balances_for_fulfill(
```
emschwartz:
```suggestion
                            // We will spawn a task to update the balances in the database
                            // so that we DO NOT wait for the database before sending the
                            // Fulfill packet back to our peer. Due to how the flow of ILP
                            // packets work, once we get the Fulfill back from the next node
                            // we need to propagate it backwards ASAP. If we do not give the
                            // previous node the fulfillment in time, they won't pay us back
                            // for the packet we forwarded. Note this means that we will
                            // relay the fulfillment _even if saving to the DB fails._
                            let fulfill_balance_update = store.update_balances_for_fulfill(
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -105,30 +111,31 @@ where
...
+                                from.clone(),
+                                incoming_amount,
+                                to.clone(),
+                                outgoing_amount,
+                            );
```
emschwartz: It's important to log if this doesn't work because otherwise we'll have no idea it didn't

```suggestion
                            ).map_err(|_| error!("Error applying balance changes for fulfill from account: {} to account: {}. Incoming amount was: {}, outgoing amount was: {}", from.id(), to.id(), incoming_amount, outgoing_amount));
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -105,30 +111,31 @@ where
...
                         .or_else(move |reject| {
-                            // can spawn here, if it is the case that the returned packet is not dependent on the outcome of the store state update
-                            store_clone.update_balances_for_reject(from_clone.clone(), incoming_amount, to_clone.clone(), outgoing_amount)
-                                .then(move |result| {
-                                    if result.is_err() {
-                                        error!("Error rolling back balance change for accounts: {} and {}. Incoming amount was: {}, outgoing amount was: {}", from_clone.id(), to_clone.id(), incoming_amount, outgoing_amount);
-                                    }
-                                    Err(reject)
-                                })
+                            let reject_balance_update = store_clone.update_balances_for_reject(
```
emschwartz:
```suggestion
                            // Similar to the logic for handling the Fulfill packet above, we
                            // spawn a task to update the balance for the Reject in parallel
                            // rather than waiting for the database to update before relaying
                            // the packet back. In this case, the only substantive difference
                            // would come from if the DB operation fails or takes too long.
                            // The packet is already rejected so it's more useful for the sender
                            // to get the error message from the original Reject packet rather
                            // than a less specific one saying that this node had an "internal
                            // error" caused by a database issue.
                            let reject_balance_update = store_clone.update_balances_for_reject(
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -105,30 +111,31 @@ where
...
+                                from_clone.clone(),
+                                incoming_amount,
+                                to_clone.clone(),
+                                outgoing_amount,
+                            );
```
emschwartz:
```suggestion
                            ).map_err(|_| error!("Error rolling back balance change for accounts: {} and {}. Incoming amount was: {}, outgoing amount was: {}", from_clone.id(), to_clone.id(), incoming_amount, outgoing_amount));
```

PR #107, opened 2019-06-18
Update Settlement Engine to RFC536

gakonst:
Latest RFC: https://github.com/interledger/rfcs/pull/536

Features:
- [x] `/accounts/:id/messages` endpoint on connector
- [x] `/accounts/:id/settlement` endpoint on connector
- [x] incoming service which checks if the packet being received should be forwarded to the settlement engine, and calls `/accounts/:id/messages` on the SE
- [x] client function which gets called by the store and calls `/accounts/:id/settlement` on the SE
- [x] idempotency - (2 types: SET/GET key to true with expiry for settlement requests. HGETALL/HSET for caching the Status Code and Data in messages)
- [x] verbose errors

### DIFF
--- a/crates/interledger-api/src/lib.rs
+++ b/crates/interledger-api/src/lib.rs
@@ -12,9 +12,9 @@ use interledger_http::{HttpAccount, HttpStore};
 use interledger_ildcp::IldcpAccount;
 use interledger_packet::Address;
 use interledger_router::RouterStore;
-use interledger_service::{Account as AccountTrait, IncomingService, OutgoingService};
+use interledger_service::{Account as AccountTrait, IncomingService};
 use interledger_service_util::{BalanceStore, ExchangeRateStore};
-use interledger_settlement::{SettlementAccount, SettlementApi, SettlementStore};
+use interledger_settlement::{SettlementAccount, SettlementStore};
 use serde::Serialize;
 use std::str;
 use tower_web::{net::ConnectionStream, ServiceBuilder};
@@ -79,25 +79,23 @@ pub struct AccountDetails {
     pub settlement_engine_ilp_address: Option<Address>,
 }
 
-pub struct NodeApi<T, S, U> {
-    store: T,
+pub struct NodeApi<S, I> {
+    store: S,
     admin_api_token: String,
     default_spsp_account: Option<String>,
-    incoming_handler: S,
-    outgoing_handler: U,
+    incoming_handler: I,
     server_secret: Bytes,
 }
 
-impl<T, S, U, A> NodeApi<T, S, U>
+impl<S, I, A> NodeApi<S, I>
 where
-    T: NodeStore<Account = A>
+    S: NodeStore<Account = A>
         + HttpStore<Account = A>
         + BalanceStore<Account = A>
         + SettlementStore<Account = A>
         + RouterStore
         + ExchangeRateStore,
-    S: IncomingService<A> + Clone + Send + Sync + 'static,
-    U: OutgoingService<A> + Clone + Send + Sync + 'static,
+    I: IncomingService<A> + Clone + Send + Sync + 'static,
     A: AccountTrait
         + HttpAccount
         + IldcpAccount
@@ -110,9 +108,8 @@ where
     pub fn new(
         server_secret: Bytes,
         admin_api_token: String,
-        store: T,
-        outgoing_handler: U,
-        incoming_handler: S,
+        store: S,
+        incoming_handler: I,
     ) -> Self {
         NodeApi {
             store,
@@ -120,7 +117,6 @@ where
             default_spsp_account: None,
             incoming_handler,
             server_secret,
-            outgoing_handler,
         }
     }
 
@@ -129,10 +125,10 @@ where
         self
     }
 
-    pub fn serve<I>(&self, incoming: I) -> impl Future<Item = (), Error = ()>
+    pub fn serve<T>(&self, incoming: T) -> impl Future<Item = (), Error = ()>
     where
-        I: ConnectionStream,
-        I::Item: Send + 'static,
+        T: ConnectionStream,
+        T::Item: Send + 'static,
     {
         ServiceBuilder::new()
             .resource(IlpApi::new(
@@ -150,10 +146,6 @@ where
                 }
                 spsp
             })
-            .resource(SettlementApi::new(
-                self.store.clone(),
-                self.outgoing_handler.clone(),
-            ))
             .resource(AccountsApi::new(
                 self.admin_api_token.clone(),
                 self.store.clone(),
--- a/crates/interledger-service-util/src/expiry_shortener_service.rs
+++ b/crates/interledger-service-util/src/expiry_shortener_service.rs
@@ -1,7 +1,7 @@
 use interledger_service::{Account, OutgoingRequest, OutgoingService};
 use std::time::Duration;
 
-pub const DEFAULT_ROUND_TRIP_TIME: u64 = 500; // milliseconds?
+pub const DEFAULT_ROUND_TRIP_TIME: u64 = 500;
 
 pub trait RoundTripTimeAccount: Account {
     fn round_trip_time(&self) -> u64 {
@@ -36,7 +36,7 @@ where
     type Future = O::Future;
 
     /// On send request:
-    /// 1. Get the sender and receiver's roundtrip time (default 500ms)
+    /// 1. Get the sender and receiver's roundtrip time (default 1000ms)
     /// 2. Reduce the packet's expiry by that amount
     /// 3. Forward the request
     fn send_request(&mut self, mut request: OutgoingRequest<A>) -> Self::Future {
--- a/crates/interledger-service-util/src/lib.rs
+++ b/crates/interledger-service-util/src/lib.rs
@@ -24,3 +24,135 @@ pub use self::rate_limit_service::{
     RateLimitAccount, RateLimitError, RateLimitService, RateLimitStore,
 };
 pub use self::validator_service::ValidatorService;
+
+pub struct ConvertDetails {
+    pub from: u8,
+    pub to: u8,
+}
+
+/// Traits for u64 and f64 asset code conversions for amounts and rates
+pub trait Convert {
+    fn normalize_scale(&self, details: ConvertDetails) -> Self;
+}
+
+impl Convert for u64 {
+    fn normalize_scale(&self, details: ConvertDetails) -> Self {
+        let from_scale = details.from;
+        let to_scale = details.to;
+        if from_scale >= to_scale {
+            self * 10u64.pow(u32::from(from_scale - to_scale))
+        } else {
+            self / 10u64.pow(u32::from(to_scale - from_scale))
+        }
+    }
+}
+
+impl Convert for f64 {
+    fn normalize_scale(&self, details: ConvertDetails) -> Self {
+        let from_scale = details.from;
+        let to_scale = details.to;
+        if from_scale >= to_scale {
+            self * 10f64.powi(i32::from(from_scale - to_scale))
+        } else {
+            self / 10f64.powi(i32::from(to_scale - from_scale))
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    /// Tests for the asset conversion
+    use super::{Convert, ConvertDetails};
+
+    #[test]
+    fn u64_test() {
+        // 1 unit with base 1, is 1 unit with base 1
+        assert_eq!(1u64.normalize_scale(ConvertDetails { from: 1, to: 1 }), 1);
+        // 1 unit with base 10, is 10 units with base 1
+        assert_eq!(1u64.normalize_scale(ConvertDetails { from: 2, to: 1 }), 10);
+        // 1 wei is 1e9 sats (multiplied by rate)
+        assert_eq!(
+            1u64.normalize_scale(ConvertDetails { from: 18, to: 9 }),
+            1_000_000_000
+        );
+
+        // there's leftovers for all number slots which do not increase in
+        // increments of 10^{to_scale-from_scale}
+        assert_eq!(1u64.normalize_scale(ConvertDetails { from: 1, to: 2 }), 0);
+        assert_eq!(10u64.normalize_scale(ConvertDetails { from: 1, to: 2 }), 1);
+        // 100 units with base 2 is 10 units with base 3
+        assert_eq!(
+            100u64.normalize_scale(ConvertDetails { from: 2, to: 3 }),
+            10
+        );
+        // 299 units with base 2 is 10 units with base 3 plus 99 leftovers
+        assert_eq!(
+            100u64.normalize_scale(ConvertDetails { from: 2, to: 3 }),
+            10
+        );
+
+        assert_eq!(999u64.normalize_scale(ConvertDetails { from: 6, to: 9 }), 0);
+        assert_eq!(
+            1000u64.normalize_scale(ConvertDetails { from: 6, to: 9 }),
+            1
+        );
+        assert_eq!(
+            1999u64.normalize_scale(ConvertDetails { from: 6, to: 9 }),
+            1
+        ); // 5 is leftovers, maybe we should return it?
+
+        // allow making sub-sat micropayments
+        assert_eq!(1u64.normalize_scale(ConvertDetails { from: 9, to: 18 }), 0);
+        assert_eq!(
+            1_000_000_000u64.normalize_scale(ConvertDetails { from: 9, to: 18 }),
+            1
+        );
+    }
+
+    #[allow(clippy::float_cmp)]
+    #[test]
+    fn f64_test() {
+        // 1 unit with base 1, is 1 unit with base 1
+        assert_eq!(1f64.normalize_scale(ConvertDetails { from: 1, to: 1 }), 1.0);
+        // 1 unit with base 10, is 10 units with base 1
+        assert_eq!(
+            1f64.normalize_scale(ConvertDetails { from: 2, to: 1 }),
+            10.0
+        );
+        // 1 wei is 1e9 sats (multiplied by rate)
+        assert_eq!(
+            1f64.normalize_scale(ConvertDetails { from: 18, to: 9 }),
+            1_000_000_000.0
+        );
+
+        // 1.0 unit with base 1 is 0.1 unit with base 2
+        assert_eq!(1f64.normalize_scale(ConvertDetails { from: 1, to: 2 }), 0.1);
+        assert_eq!(
+            10f64.normalize_scale(ConvertDetails { from: 1, to: 2 }),
+            1.0
+        );
+        // 100 units with base 2 is 10 units with base 3
+        assert_eq!(
+            100f64.normalize_scale(ConvertDetails { from: 2, to: 3 }),
+            10.0
+        );
+        // 299 units with base 2 is 10 units with base 3 plus 99 leftovers
+        assert_eq!(
+            100f64.normalize_scale(ConvertDetails { from: 2, to: 3 }),
+            10.0
+        );
+
+        assert_eq!(
+            999f64.normalize_scale(ConvertDetails { from: 6, to: 9 }),
+            0.999
+        );
+        assert_eq!(
+            1000f64.normalize_scale(ConvertDetails { from: 6, to: 9 }),
+            1.0
+        );
+        assert_eq!(
+            1999f64.normalize_scale(ConvertDetails { from: 6, to: 9 }),
+            1.999
+        );
+    }
+}
--- a/crates/interledger-settlement/src/api.rs
+++ b/crates/interledger-settlement/src/api.rs
@@ -1,161 +1,707 @@
-use crate::{SettlementAccount, SettlementStore};
+use super::{SettlementAccount, SettlementStore};
+use bytes::Bytes;
 use futures::{
-    future::{err, result, Either},
+    future::{ok, result, Either},
     Future,
 };
-use hyper::Response;
+use hyper::{Response, StatusCode};
 use interledger_ildcp::IldcpAccount;
 use interledger_packet::PrepareBuilder;
 use interledger_service::{AccountStore, OutgoingRequest, OutgoingService};
-use serde_json::Value;
+use interledger_service_util::{Convert, ConvertDetails};
+use ring::digest::{digest, SHA256};
 use std::{
     marker::PhantomData,
     str::{self, FromStr},
     time::{Duration, SystemTime},
 };
+use tower_web::{net::ConnectionStream, ServiceBuilder};
 
 static PEER_PROTOCOL_CONDITION: [u8; 32] = [
     102, 104, 122, 173, 248, 98, 189, 119, 108, 143, 193, 139, 142, 159, 142, 32, 8, 151, 20, 133,
     110, 226, 51, 179, 144, 42, 89, 29, 13, 95, 41, 37,
 ];
 
-pub struct SettlementApi<S, T, A> {
-    outgoing_handler: S,
-    store: T,
-    account_type: PhantomData<A>,
-}
-
-#[derive(Extract)]
-#[serde(rename_all = "camelCase")]
-struct SettlementDetails {
-    account_id: String,
+#[derive(Debug, Clone, Extract)]
+struct SettlementData {
     amount: u64,
 }
 
-#[derive(Response)]
-#[web(status = "200")]
-struct Success;
-
-// TODO add authentication
+pub struct SettlementApi<S, O, A> {
+    outgoing_handler: O,
+    store: S,
+    account_type: PhantomData<A>,
+}
 
 impl_web! {
-    impl<S, T, A> SettlementApi<S, T, A>
+    impl<S, O, A> SettlementApi<S, O, A>
     where
-        S: OutgoingService<A> + Clone + Send + Sync + 'static,
-        T: SettlementStore<Account = A> + AccountStore<Account = A> + Clone + Send + Sync + 'static,
+        S: SettlementStore<Account = A> + AccountStore<Account = A> + Clone + Send + Sync + 'static,
+        O: OutgoingService<A> + Clone + Send + Sync + 'static,
         A: SettlementAccount + IldcpAccount + Send + Sync + 'static,
     {
-        pub fn new(store: T, outgoing_handler: S) -> Self {
+        pub fn new(store: S, outgoing_handler: O) -> Self {
             SettlementApi {
-                outgoing_handler,
                 store,
+                outgoing_handler,
                 account_type: PhantomData,
             }
         }
 
-        #[post("/settlements/receiveMoney")]
-        fn receive_settlement(&self, body: SettlementDetails) -> impl Future<Item = Success, Error = Response<()>> {
-            let amount = body.amount;
-            let store = self.store.clone();
-            let store_clone = store.clone();
-            let account_id = body.account_id;
-            result(A::AccountId::from_str(account_id.as_str())
-                .map_err(move |_err| {
-                    error!("Unable to parse account id: {}", account_id);
-                    Response::builder().status(400).body(()).unwrap()
-                }))
-                .and_then(move |account_id| store.get_accounts(vec![account_id]).map_err(move |_| {
-                    error!("Error getting account: {}", account_id);
-                    Response::builder().status(404).body(()).unwrap()
-                }))
-                .and_then(move |mut accounts| {
-                    let account = accounts.pop().unwrap();
-                    if let Some(settlement_engine) = account.settlement_engine_details() {
-                        Ok((account, settlement_engine))
-                    } else {
-                        error!("Account {} does not have settlement engine details configured. Cannot handle incoming settlement", account.id());
-                        Err(Response::builder().status(500).body(()).unwrap())
+        fn check_idempotency(
+            &self,
+            idempotency_key: Option<String>,
+            input_hash: [u8; 32],
+        ) -> impl Future<Item = Option<(StatusCode, Bytes)>, Error = (StatusCode, String)> {
+            if let Some(idempotency_key) = idempotency_key {
+                Either::A(self.store.load_idempotent_data(idempotency_key.clone())
+                .map_err(move |err| {
+                    let err = format!("Couldn't connect to store {:?}", err);
+                    error!("{}", err);
+                    (StatusCode::from_u16(500).unwrap(), err)
+                }).and_then(move |ret: Option<(StatusCode, Bytes, [u8; 32])>| {
+                        if let Some(d) = ret {
+                            if d.2 != input_hash {
+                                // Stripe CONFLICT status code
+                                return Err((StatusCode::from_u16(409).unwrap(), "Provided idempotency key is tied to other input".to_string()))
+                            }
+                            if d.0.is_success() {
+                                return Ok(Some((d.0, d.1)))
+                            } else {
+                                return Err((d.0, String::from_utf8_lossy(&d.1).to_string()))
+                            }
+                        }
+                        Ok(None)
                     }
-                })
-                .and_then(move |(account, settlement_engine)| {
-                    let account_id = account.id();
-
-                    let amount = if account.asset_scale() >= settlement_engine.asset_scale {
-                        amount
-                            * 10u64.pow(u32::from(
-                                account.asset_scale() - settlement_engine.asset_scale,
-                            ))
-                    } else {
-                        amount
-                            / 10u64.pow(u32::from(
-                                settlement_engine.asset_scale - account.asset_scale(),
-                            ))
-                    };
-
-                    store_clone.update_balance_for_incoming_settlement(account_id, amount)
-                        .map_err(move |_| {
-                            error!("Error updating balance of account: {} for incoming settlement of amount: {}", account_id, amount);
-                            Response::builder().status(500).body(()).unwrap()
-                        })
-                })
-                .and_then(|_| Ok(Success))
+                ))
+            } else {
+                Either::B(ok(None))
+            }
         }
 
-        #[post("/settlements/sendMessage")]
-        fn send_outgoing_message(&self, body: Value)-> impl Future<Item = Value, Error = Response<()>> {
-            if let Value::Object(json) = &body {
-                if let Some(account_id) = json.get("accountId").and_then(|a| a.as_str()) {
-                    if let Ok(account_id) = A::AccountId::from_str(account_id) {
-                        let mut outgoing_handler = self.outgoing_handler.clone();
-                        return Either::A(self.store.get_accounts(vec![account_id])
-                            .map_err(move |_| {
-                                error!("Account {} not found", account_id);
-                                Response::builder().status(404).body(()).unwrap()
-                            })
-                            .and_then(|accounts| {
-                                let account = &accounts[0];
-                                if let Some(settlement_engine) = account.settlement_engine_details() {
-                                    Ok((account.clone(), settlement_engine))
-                                } else {
-                                    error!("Account {} has no settlement engine details configured, cannot send a settlement engine message to that account", accounts[0].id());
-                                    Err(Response::builder().status(404).body(()).unwrap())
-                                }
-                            })
-                            .and_then(move |(account, settlement_engine)| {
-                                // Send the message to the peer's settlement engine.
-                                // Note that we use dummy values for the `from` and `original_amount`
-                                // because this `OutgoingRequest` will bypass the router and thus will not
-                                // use either of these values. Including dummy values in the rare case where
-                                // we do not need them seems easier than using
-                                // `Option`s all over the place.
-                                outgoing_handler.send_request(OutgoingRequest {
-                                    from: account.clone(),
-                                    to: account.clone(),
-                                    original_amount: 0,
-                                    prepare: PrepareBuilder {
-                                        destination: settlement_engine.ilp_address,
-                                        amount: 0,
-                                        expires_at: SystemTime::now() + Duration::from_secs(30),
-                                        data: body.to_string().as_bytes().as_ref(),
-                                        execution_condition: &PEER_PROTOCOL_CONDITION,
-                                    }.build()
-                                })
-                                .map_err(|reject| {
-                                    error!("Error sending message to peer settlement engine. Packet rejected with code: {}, message: {}", reject.code(), str::from_utf8(reject.message()).unwrap_or_default());
-                                    // TODO should we respond with different HTTP error codes based on the ILP error codes?
-                                    Response::builder().status(502).body(()).unwrap()
-                                })
-                            })
-                            .and_then(|fulfill| {
-                                serde_json::from_slice(fulfill.data()).map_err(|err| {
-                                    error!("Error parsing response from peer settlement engine as JSON: {:?}", err);
-                                    Response::builder().status(502).body(()).unwrap()
-                                })
-                            }));
-                    }
+        #[post("/accounts/:account_id/settlement")]
+        fn receive_settlement(&self, account_id: String, body: SettlementData, idempotency_key: Option<String>) -> impl Future<Item = Response<Bytes>, Error = Response<String>> {
+            let amount = body.amount;
+            let store = self.store.clone();
+
+            let input = format!("{}{:?}", account_id, body);
+            let input_hash = get_hash_of(input.as_ref());
+
+            self.check_idempotency(idempotency_key.clone(), input_hash).map_err(|res| {
+                Response::builder().status(res.0).body(res.1).unwrap()
+            })
+            .and_then(move |ret: Option<(StatusCode, Bytes)>| {
+                if let Some(d) = ret {
+                    return Either::A(ok(Response::builder().status(d.0).body(d.1).unwrap()));
                 }
-            }
-            Either::B(err(Response::builder().status(400).body(()).unwrap()))
+                Either::B(
+                    result(A::AccountId::from_str(&account_id)
+                    .map_err({
+                        let store = store.clone();
+                        let idempotency_key = idempotency_key.clone();
+                        move |_err| {
+                        let error_msg = format!("Unable to parse account id: {}", account_id);
+                        error!("{}", error_msg);
+                        let status_code = StatusCode::from_u16(400).unwrap();
+                        let data = Bytes::from(error_msg.clone());
+                        if let Some(idempotency_key) = idempotency_key {
+                            store.save_idempotent_data(idempotency_key, input_hash, status_code, data);
+                        }
+                        Response::builder().status(400).body(error_msg).unwrap()
+                    }}))
+                    .and_then({
+                        let store = store.clone();
+                        let idempotency_key = idempotency_key.clone();
+                        move |account_id| {
+                        store.get_accounts(vec![account_id])
+                        .map_err({
+                            let store = store.clone();
+                            let idempotency_key = idempotency_key.clone();
+                            move |_err| {
+                            let error_msg = format!("Error getting account: {}", account_id);
+                            error!("{}", error_msg);
+
+                            let status_code = StatusCode::from_u16(404).unwrap();
+                            let data = Bytes::from(error_msg.clone());
+                            if let Some(idempotency_key) = idempotency_key {
+                                store.save_idempotent_data(idempotency_key.clone(), input_hash, status_code, data);
+                            }
+                            Response::builder().status(404).body(error_msg).unwrap()
+                        }})
+                    }})
+                    .and_then({
+                        let store = store.clone();
+                        let idempotency_key = idempotency_key.clone();
+                        move |accounts| {
+                        let account = &accounts[0];
+                        if let Some(settlement_engine) = account.settlement_engine_details() {
+                            Ok((account.clone(), settlement_engine))
+                        } else {
+                            let error_msg = format!("Account {} does not have settlement engine details configured. Cannot handle incoming settlement", account.id());
+                            error!("{}", error_msg);
+                            if let Some(idempotency_key) = idempotency_key {
+                                store.save_idempotent_data(idempotency_key, input_hash, StatusCode::from_u16(404).unwrap(), Bytes::from(error_msg.clone()));
+                            }
+                            Err(Response::builder().status(404).body(error_msg).unwrap())
+                        }
+                    }})
+                    .and_then({
+                        let store = store.clone();
+                        let idempotency_key = idempotency_key.clone();
+                        move |(account, settlement_engine)| {
+                        let account_id = account.id();
+                        let amount = amount.normalize_scale(ConvertDetails {
+                            from: account.asset_scale(),
+                            to: settlement_engine.asset_scale
+                        });
+                        store.update_balance_for_incoming_settlement(account_id, amount, idempotency_key)
+                            .map_err(move |_| {
+                                let err = format!("Error updating balance of account: {} for incoming settlement of amount: {}", account_id, amount);
+                                error!("{}", err);
+                                Response::builder().status(500).body(err).unwrap()
+                        })
+                    }}).and_then({
+                        let store = store.clone();
+                        let idempotency_key = idempotency_key.clone();
+                        move |_| {
+                        let ret = Bytes::from("Success");
+                        if let Some(idempotency_key) = idempotency_key {
+                            store.save_idempotent_data(idempotency_key, input_hash, StatusCode::OK, ret.clone());
+                        }
+                        ok(())
+                        .and_then(|_| Ok(Response::builder().status(StatusCode::OK).body(ret).unwrap()))
+                    }}))
+                })
+        }
+
+        // Gets called by our settlement engine, forwards the request outwards
+        // until it reaches the peer's settlement engine. Extract is not
+        // implemented for Bytes unfortunately.
+        #[post("/accounts/:account_id/messages")]
+        fn send_outgoing_message(&self, account_id: String, body: Vec<u8>, idempotency_key: Option<String>)-> impl Future<Item = Response<Bytes>, Error = Response<String>> {
+            let store = self.store.clone();
+            let mut outgoing_handler = self.outgoing_handler.clone();
+
+            let input = format!("{}{:?}", account_id, body);
+            let input_hash = get_hash_of(input.as_ref());
+
+            self.check_idempotency(idempotency_key.clone(), input_hash).map_err(|res| {
+                Response::builder().status(res.0).body(res.1).unwrap()
+            })
+            .and_then(move |ret: Option<(StatusCode, Bytes)>| {
+                if let Some(d) = ret {
+                    return Either::A(ok(Response::builder().status(d.0).body(d.1).unwrap()));
+                }
+                Either::B(
+                result(A::AccountId::from_str(&account_id)
+                .map_err({
+                    let store = store.clone();
+                    let idempotency_key = idempotency_key.clone();
+                    move |_err| {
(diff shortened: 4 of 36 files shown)


crates/interledger-settlement/src/message_service.rs
```diff
@@ -77,13 +79,14 @@ where
...
                                     RejectBuilder {
-                                    code: ErrorCode::T00_INTERNAL_ERROR,
-                                    message: b"Error getting settlement engine response",
-                                    data: &[],
-                                    triggered_by: Some(&ilp_address),
-                                }.build()
+                                        code: ErrorCode::T00_INTERNAL_ERROR,
+                                        message: b"Error getting settlement engine response",
+                                        data: &[],
+                                        triggered_by: Some(&engine_address),
```
emschwartz: This shouldn't be the `engine_address` because that's going to be something hard-coded for each engine like `peer.settle.xrp-paychan`, which isn't helpful for debugging which peer's system actually produced the error. Also, the `ilp_address` is that of the connector, which in this case is the entity that produced the reject packet (even though they did it based on an error produced upstream)

crates/interledger-settlement/src/message_service.rs
```diff
@@ -148,3 +152,285 @@ where
...
+    use mockito::Matcher;
+
+    use std::str::FromStr;
+    use tokio::runtime::Runtime;
+    use url::Url;
+
+    use std::str;
+
+    #[test]
+    fn settlement_ok_valid_json() {
```
emschwartz: It shouldn't expect the body to be JSON anymore

crates/interledger-settlement/src/api.rs
```diff
@@ -92,70 +102,148 @@ impl_web! {
...
+                        }.build()
+                    })
+                    .map_err(|reject| {
+                        error!("Error sending message to peer settlement engine. Packet rejected with code: {}, message: {}", reject.code(), str::from_utf8(reject.message()).unwrap_or_default());
+                        // spec: "Could not process sending of the message -> 400"
+                        Response::builder().status(400).body(()).unwrap()
+                    })
+                })
+                .and_then(|fulfill| {
+                    serde_json::from_slice(fulfill.data()).map_err(|err| {
```
gakonst: Can we just return `fulfill.data()`? I think the engine returns raw data as per the latest spec

emschwartz: Yes

crates/interledger-settlement/src/api.rs
```diff
@@ -19,67 +19,77 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
                 store,
+                outgoing_handler,
                 account_type: PhantomData,
             }
         }
 
-        #[post("/settlements/receiveMoney")]
-        fn receive_settlement(&self, body: SettlementDetails) -> impl Future<Item = Success, Error = Response<()>> {
+
+        // TODO: The SE should retry until this is ACK’d so it needs to be idempotent,
```
emschwartz: :+1:

crates/interledger-settlement/src/api.rs
```diff
@@ -19,67 +19,77 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+        // https://stripe.com/docs/api/idempotent_requests?lang=curl
+        // TODO: Can we make account_id: A::AccountId somehow?
+        // derive(Extract) is not possible since it's inside a trait.
```
emschwartz: I can't think of a way to do it but using `AccountId::from_str` seems okay

crates/interledger-settlement/src/api.rs
```diff
@@ -19,67 +19,77 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+        // TODO: Can the Response<()> be converted to a Response<String>? It'd
```
emschwartz: Yes, that was just me being lazy on the first pass. We might want to return JSON errors rather than just the error message as plain text

crates/interledger-settlement/src/api.rs
```diff
@@ -19,67 +19,77 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+        // be nice if we could include the full error message body (currently
+        // it's just the header)
+        #[post("/accounts/:account_id/settlement")]
+        fn receive_settlement(&self, account_id: String, body: SettlementDetails) -> impl Future<Item = Success, Error = Response<()>> {
             let amount = body.amount;
+            let _scale = body.scale; // todo: figure out how to use this, is it really necessary? should we check if it matches the SE details?
```
emschwartz: If the scale is going to be sent on every request, we shouldn't need the scale included in the `SettlementEngineDetails`

crates/interledger-settlement/src/api.rs
```diff
@@ -19,67 +19,77 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
             let store = self.store.clone();
             let store_clone = store.clone();
-            let account_id = body.account_id;
-            result(A::AccountId::from_str(account_id.as_str())
+            result(A::AccountId::from_str(&account_id)
                 .map_err(move |_err| {
                     error!("Unable to parse account id: {}", account_id);
-                    Response::builder().status(400).body(()).unwrap()
+                    Response::builder().status(404).body(()).unwrap()
```
emschwartz: If the account ID is invalid I don't think it's a 404

crates/interledger-settlement/src/api.rs
```diff
@@ -92,70 +102,148 @@ impl_web! {
                             ))
                     };
 
+                    // TODO Idempotency header!
```
emschwartz: :+1:

crates/interledger-settlement/src/api.rs
```diff
@@ -92,70 +102,148 @@ impl_web! {
...
                     store_clone.update_balance_for_incoming_settlement(account_id, amount)
                         .map_err(move |_| {
                             error!("Error updating balance of account: {} for incoming settlement of amount: {}", account_id, amount);
-                            Response::builder().status(500).body(()).unwrap()
+                            Response::builder().status(201).body(()).unwrap() // Request was sent, but SE operation have failed.
```
emschwartz: We shouldn't ack this if we haven't been able to process it. In this case, we know the processing failed so we should definitely return an error that will cause the SE to retry

```suggestion
                            Response::builder().status(500).body(()).unwrap() // Request was received but we were unable to process it (the settlement engine should retry this request)
```

crates/interledger-settlement/src/api.rs
```diff
@@ -92,70 +102,148 @@ impl_web! {
...
-                            .and_then(|fulfill| {
-                                serde_json::from_slice(fulfill.data()).map_err(|err| {
-                                    error!("Error parsing response from peer settlement engine as JSON: {:?}", err);
-                                    Response::builder().status(502).body(()).unwrap()
-                                })
-                            }));
+        // Gets called by our settlement engine, forwards the request outwards
+        // until it reaches the peer's settlement engine
+        #[post("/accounts/:account_id/messages")]
+        fn send_outgoing_message(&self, account_id: String, body: String)-> impl Future<Item = Value, Error = Response<()>> {
```
emschwartz: Shouldn't the body be `Bytes` or `Vec<u8>`?

crates/interledger-settlement/src/api.rs
```diff
@@ -92,70 +102,148 @@ impl_web! {
...
+                    // we do not need them seems easier than using
+                    // `Option`s all over the place.
+                    outgoing_handler.send_request(OutgoingRequest {
+                        from: account.clone(),
+                        to: account.clone(),
+                        original_amount: 0,
+                        prepare: PrepareBuilder {
+                            destination: settlement_engine.ilp_address,
+                            amount: 0,
+                            expires_at: SystemTime::now() + Duration::from_secs(30),
```
emschwartz: I wonder if we should set this shorter to make sure the HTTP request doesn't time out

crates/interledger-settlement/src/api.rs
```diff
@@ -92,70 +102,148 @@ impl_web! {
...
+                            data: body.as_ref(),
+                            execution_condition: &PEER_PROTOCOL_CONDITION,
+                        }.build()
+                    })
+                    .map_err(|reject| {
+                        error!("Error sending message to peer settlement engine. Packet rejected with code: {}, message: {}", reject.code(), str::from_utf8(reject.message()).unwrap_or_default());
+                        // spec: "Could not process sending of the message -> 400"
+                        Response::builder().status(400).body(()).unwrap()
```
emschwartz: We should probably look at what the `Reject`'s error code was to determine what HTTP error code to respond with. A 502 might be more appropriate (400 definitely isn't right, because the problem wasn't the SE request)

crates/interledger-settlement/src/api.rs
emschwartz: Yes

crates/interledger-settlement/src/fixtures.rs
```diff
@@ -0,0 +1,24 @@
...
+
+pub static DATA: &str = "DATA_FOR_SETTLEMENT_ENGINE";
+pub static BODY: &str = "hi";
+pub static SETTLEMENT_BODY: SettlementDetails = SettlementDetails {
+    amount: 100,
+    scale: 9,
+};
+
+lazy_static! {
+    pub static ref TEST_MUTEX: Mutex<()> = Mutex::new(());
```
emschwartz: Is this needed for these tests? The main reason to use this for the redis integration tests is that it seems to have a problem with multiple threads accessing it simultaneously from the tests (see https://github.com/mitsuhiko/redis-rs/issues/138)

crates/interledger-settlement/src/message_service.rs
```diff
@@ -49,96 +48,69 @@ where
             if request.prepare.destination() == settlement_engine_details.ilp_address {
                 let ilp_address_clone = self.ilp_address.clone();
                 let mut settlement_engine_url = settlement_engine_details.url;
+                // it is expected that the request's data is already in a proper
```
emschwartz:
```suggestion
                // The `Prepare` packet's data was sent by the peer's settlement
                // engine so we assume it is in a format that our settlement engine
                // will understand
```

crates/interledger-settlement/src/message_service.rs
```diff
@@ -49,96 +48,69 @@ where
...
+                // format. `to_vec()` needed to work around lifetime error
```
emschwartz: Copying the data isn't a big issue if the message sizes are small. That said, we might want to think about adding a `Prepare.freeze()` method, or something like that that would turn it into an `ImmutablePrepare` backed by a `Bytes` instead of a `BytesMut`. In general, using a `BytesMut` enables us to do in-place updating of the `amount` and `expires_at`, but in this case it means we can't have a method that returns a slice of the data as a `Bytes` (that wouldn't require copying but would solve the lifetime issue)

cc @sentientwaffle

crates/interledger-settlement/src/message_service.rs
```diff
@@ -49,96 +48,69 @@ where
...
-                        .send()
-                        .map_err(move |error| {
-                            error!("Error sending message to settlement engine: {:?}", error);
+                // TODO add auth
+                settlement_engine_url
+                    .path_segments_mut()
+                    .expect("Invalid settlement engine URL")
+                    .push("accounts")
+                    .push(&request.from.id().to_string())
+                    .push("messages"); // Maybe set the idempotency flag here in the headers
```
emschwartz: This should also be idempotent (in addition to the header, we also need the client to retry it accordingly)

crates/interledger-settlement/src/message_service.rs
```diff
@@ -49,96 +48,69 @@ where
...
+                        message: b"Error sending message to settlement engine",
+                        data: &[],
+                        triggered_by: Some(&ilp_address_clone),
+                    }.build()
+                })
+                .and_then(move |response| {
+                    let status = response.status();
+                    if status.is_success() {
+                        Either::A(response.into_body().concat2().map_err(move |err| {
+                            // When can this case be reached?  Unclear when `concat2` fails
```
emschwartz: From the [docs](https://docs.rs/futures/0.1.27/futures/stream/trait.Stream.html#method.concat2):
> If an error occurs, all the results will be dropped and the error will be returned.

crates/interledger-settlement/src/api.rs
gakonst: Based on : https://interledger.slack.com/files/UHPJB89SL/FJYEX3WCA/updated_spec.yaml
```
Thanks for the comments on the PR. Regarding the HTTP Error Codes, I was based on https://interledger.slack.com/files/UHPJB89SL/FJYEX3WCA/updated_spec.yaml.

>   /accounts/{accountId}/settlement:
> responses:
        200:
          description: Acknowledgement the endpoint recieved the request and settlement
            was successful.
        201:
          description: Acknowledgement the endpoint recieved the request. NOTE, for
            the Settlement Engine this does not mean settlement has taken place
        404:
          description: Account was not found for given AccountId
```

Since the 400 error code was not specified, I decided to go with a 404.

crates/interledger-settlement/src/api.rs
gakonst: Agreed, however: https://forum.interledger.org/t/settlement-architecture/545/22?u=gakonst



```
201:
          description: Acknowledgement the endpoint recieved the request. NOTE, for
            the Settlement Engine this does not mean settlement has taken place
```

crates/interledger-settlement/src/api.rs
gakonst: If it becomes shorter isn't it more likely we'll send an expired packet? Did you mean longer?

crates/interledger-settlement/src/api.rs
emschwartz: We should define 400 and use it

crates/interledger-settlement/src/api.rs
gakonst: We call `body.as_ref()`, whcih does a cheap pointer conversion, don't think it should matter?

crates/interledger-settlement/src/message_service.rs
gakonst: Yeah the idempotency features are going to be the last part of the PR

crates/interledger-settlement/src/message_service.rs
gakonst: I read that, was having a hard time writing a test case which gets into this execution path

crates/interledger-settlement/src/fixtures.rs
gakonst: It is not, I copied the `block_on` helper which used it without much thought. Removed, thanks

crates/interledger-settlement/src/api.rs
gakonst: Agreed.

crates/interledger-settlement/src/message_service.rs
emschwartz: I wouldn't worry about writing a test case for that

crates/interledger-settlement/src/api.rs
emschwartz: It should only respond with a 201 if it saved the request to the store such that it does not need to be retried again and will be processed later. If we got an error and don't retry it on our side, we definitely shouldn't respond with a 2xx code

crates/interledger-settlement/src/api.rs
emschwartz: I meant shorter, because I think the normal HTTP request timeout is 30 seconds. It's possible that we send the request and the request from the SE times out. You're right that decreasing the timeout increases the chance that the packet ends up being rejected because it hits the timeout, though it really should not take anywhere near 30 seconds

crates/interledger-settlement/src/api.rs
emschwartz: This will make sure it's valid utf8. We do not want it to check for that

emschwartz: `Vec<u8>` will have the same performance as `String` but will allow any bytes

crates/interledger-settlement/src/api.rs
gakonst: That was my bad, I thought this was a SE operation, while it is a store operation. Definitely a `500` here.

crates/interledger-settlement/src/api.rs
```diff
@@ -19,47 +19,48 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
-    account_id: String,
-    amount: u64,
-}
-
-#[derive(Response)]
+#[derive(Debug, Response)]
 #[web(status = "200")]
 struct Success;
 
 // TODO add authentication
```
emschwartz: It seems like we may not end up adding authentication to this API and instead assuming that it's run inside the firewall. If that's the case, we will need to run this API on a different port than the rest of the API and stop including it in the `InterledgerApi` `tower-web` service.

crates/interledger-store-redis/src/store.rs
```diff
@@ -105,8 +105,21 @@ return balance";
 static PROCESS_INCOMING_SETTLEMENT: &str = "
 local account = 'accounts:' .. ARGV[1]
 local amount = tonumber(ARGV[2])
+local idempotency_key = ARGV[3]
+
 local balance, prepaid_amount = unpack(redis.call('HMGET', account, 'balance', 'prepaid_amount'))
 
+-- If idempotency key has been used, then do not perform any operations
+local exists = redis.call('HGET', idempotency_key, 'set')
```
emschwartz: What is the `'set'` part about?

crates/interledger-store-redis/src/store.rs
```diff
@@ -1134,16 +1147,93 @@ impl RateLimitStore for RedisStore {
...
+                        idempotency_key_clone, data_clone, err
+                    )
+                })
+                .and_then(move |(connection, _balance): (_, i64)| {
+                    trace!(
+                        "Cached idempotency key {} to value {:?}",
+                        idempotency_key,
+                        data
+                    );
+                    cmd("EXPIRE")
```
emschwartz: Instead of issuing two separate commands, you should use `pipe` ([pipelines](https://redis.io/topics/pipelining)), as is done here: https://github.com/emschwartz/interledger-rs/blob/8ac5eb74a246e6b88b3b722426ed7a618d2b7a9a/crates/interledger-store-redis/src/store.rs#L322-L323

crates/interledger-settlement/src/api.rs
```diff
@@ -19,47 +19,48 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
-            let amount = body.amount;
+        pub fn store(&self) -> S {
+            self.store.clone()
+        }
+
+        // TODO: Can the Response<()> be converted to a Response<String>? It'd
+        // be nice if we could include the full error message body (currently
+        // it's just the header)
+        // TODO: Verify that the idempotency_key is seen as a "Idempotency-Key"
+        // header by impl_web!
```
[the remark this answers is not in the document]
gakonst: Yeah I read that, hence why I wrote "verify" in the comment

crates/interledger-store-redis/src/store.rs
gakonst: Actually I just realized that redis' set command can specify expiry in seconds as:

`SET <K> <V> EX <T>`

crates/interledger-store-redis/src/store.rs
gakonst: Was part of using the HGET/HSET commands, but looks like GET/SET will do the trick. I had thought to use HGET/HSET due to this blogpost by Instagram Eng: https://instagram-engineering.com/storing-hundreds-of-millions-of-simple-key-value-pairs-in-redis-1091ae80f74c

crates/interledger-settlement/src/api.rs
```diff
@@ -68,31 +69,20 @@ impl_web! {
...
                     if let Some(settlement_engine) = account.settlement_engine_details() {
-                        Ok((account, settlement_engine))
+                        Ok((account.clone(), settlement_engine))
                     } else {
                         error!("Account {} does not have settlement engine details configured. Cannot handle incoming settlement", account.id());
-                        Err(Response::builder().status(500).body(()).unwrap())
+                        Err(Response::builder().status(404).body(()).unwrap())
                     }
                 })
                 .and_then(move |(account, settlement_engine)| {
```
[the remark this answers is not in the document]
gakonst: That's checked inside the settlement script. There's no state mutation until it reaches https://github.com/emschwartz/interledger-rs/blob/a1e0b5c1389a94084bcdec69f0bf7c793e375000/crates/interledger-settlement/src/api.rs#L85, but I see what you're saying: The API itself isn't idempotent as it's coded here, the store is

crates/interledger-service-util/src/lib.rs
```diff
@@ -24,3 +24,84 @@ pub use self::rate_limit_service::{
...
 };
 pub use self::validator_service::ValidatorService;
+
+/// Traits for u64 and f64 asset code conversions for amounts and rates
+pub trait Convert {
+    fn normalize_scale(&self, from_scale: u8, to_scale: u8) -> Self;
+}
+
+impl Convert for u64 {
+    fn normalize_scale(&self, from_scale: u8, to_scale: u8) -> Self {
```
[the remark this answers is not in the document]
gakonst: We could add a `ConvertDetails { from: u8, to: u8 }` struct and make it take that as an argument, that way the params would need to be named and order wouldn't matter.

crates/interledger-settlement/src/api.rs
```diff
@@ -68,31 +69,20 @@ impl_web! {
...
-                    } else {
-                        amount
-                            / 10u64.pow(u32::from(
-                                settlement_engine.asset_scale - account.asset_scale(),
-                            ))
-                    };
-
-                    store_clone.update_balance_for_incoming_settlement(account_id, amount)
+                    store_clone.update_balance_for_incoming_settlement(account_id, amount, idempotency_key)
                         .map_err(move |_| {
```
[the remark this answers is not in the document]
gakonst: It already is in the function, what do you mean?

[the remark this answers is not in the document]
emschwartz: Missed that, sorry

gakonst:
Flaky test (locally):

```
cargo test

<redacted>

running 1 test
[2019-06-23T18:14:41Z ERROR interledger_http::client] Error sending HTTP request: Error(Hyper(Error(Connect, Os { code: 61, kind: ConnectionRefused, message: "Connection refused" })), "http://localhost:3010/ilp")
Error sending from node 1 to node 3: Error(Hyper(Error(Connect, Os { code: 61, kind: ConnectionRefused, message: "Connection refused" })), "http://localhost:3010/pay")
test three_nodes ... FAILED

failures:

---- three_nodes stdout ----
thread 'three_nodes' panicked at 'called `Result::unwrap()` on an `Err` value: Error(Hyper(Error(Connect, Os { code: 61, kind: ConnectionRefused, message: "Connection refused" })), "http://localhost:3010/pay")', src/libcore/result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.


failures:
    three_nodes
```

crates/interledger-store-redis/src/store.rs
```diff
@@ -1134,16 +1148,81 @@ impl RateLimitStore for RedisStore {
...
+        idempotency_key: String,
+        status_code: StatusCode,
+        data: Bytes,
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        let mut pipe = redis::pipe();
+        pipe.atomic()
+            .hset(idempotency_key.clone(), "status_code", status_code.as_u16())
+            .ignore()
+            .hset(idempotency_key.clone(), "data", data.as_ref())
+            .ignore() // Can we assume the data being passed is utf8? If so, maybe the funciton can take utf8 as argument.
```
gakonst:
```suggestion
```

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +22,516 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+                Response::builder().status(500).body(err).unwrap()
+            }).and_then(move |data: Option<(StatusCode, Bytes)>| {
+                if let Some(d) = data {
+                    if d.0.is_success() {
+                        let data = Response::builder().status(d.0).body(d.1).unwrap();
+                        return Either::A(Either::A(ok(data)));
                     } else {
-                        error!("Account {} does not have settlement engine details configured. Cannot handle incoming settlement", account.id());
-                        Err(Response::builder().status(500).body(()).unwrap())
+                        let body = String::from_utf8(d.1.to_vec()).unwrap(); /// if success return bytes otherwise return string
```
gakonst: This can crash if non-utf8 characters are cached, what's a good way to go from Bytes to String safely?

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +22,516 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
-                    if let Some(settlement_engine) = account.settlement_engine_details() {
-                        Ok((account, settlement_engine))
+        #[post("/accounts/:account_id/settlement")]
+        fn receive_settlement(&self, account_id: String, body: u64, idempotency_key: String) -> impl Future<Item = Response<Bytes>, Error = Response<String>> {
+            let amount = body;
+            // These clones are very ugly (but cheap since this is an Arc). They
+            // are required due to the multiple moves inside the futures chain.
+            // Can we remove them in some smart way?
+            // https://github.com/rust-lang/rfcs/issues/2407
+            // https://users.rust-lang.org/t/automatic-cloning-for-closures/2578
```
gakonst: Chose to go with clone_all, `enclose` was not sophisticated enough work with `and_then` chains

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +22,516 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
-            Either::B(err(Response::builder().status(400).body(()).unwrap()))
+
+                Either::B(
+                result(A::AccountId::from_str(&account_id)
+                .map_err({ clone_all!(store, idempotency_key); move |_err| {
+                    let error_msg = format!("Unable to parse account id: {}", account_id);
+                    error!("{}", error_msg);
+                    let status_code = StatusCode::from_u16(400).unwrap();
+                    let data = Bytes::from(error_msg.clone());
+                    // save the idempotency data, can do .wait() here? .then /
```
gakonst: Unclear if `.wait`ing is a good approach here

emschwartz: `.wait` should only be used in tests if at all since it blocks and removes the point of using async io in the first place

crates/interledger-settlement/src/api.rs
emschwartz: The only way is to do what you're doing but handle the error case so it doesn't panic

gakonst:
Merging master seems to have stopped the flaky tests from happening. 

Maybe https://github.com/emschwartz/interledger-rs/pull/107/commits/ea2963d0b4c0a1046bcc614f3cda41e63bfcc945 fixed it, or the removal of `.wait()` from store operations?

It failed here only: https://circleci.com/gh/emschwartz/interledger-rs/553. 

These 3 builds reproduce the expired packet error (#109):
https://circleci.com/gh/emschwartz/interledger-rs/554
https://circleci.com/gh/emschwartz/interledger-rs/551
https://circleci.com/gh/emschwartz/interledger-rs/553

crates/interledger-settlement/src/client.rs
```diff
@@ -4,15 +4,8 @@ use futures::{
...
-
-#[derive(Serialize)]
-#[serde(rename_all = "camelCase")]
-struct SendSettlement {
-    account_id: String,
-    amount: String,
-}
 
 #[derive(Clone)]
 pub struct SettlementClient {
```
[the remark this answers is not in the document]
gakonst: To rephrase what you wrote:

Instead of doing https://github.com/emschwartz/interledger-rs/blob/gakonst-settlement-update/crates/interledger-store-redis/src/store.rs#L541-L561, right after the `PROCESS_FULFILL` script gets run in redis, you're saying that we should have an extra service which will be querying the database (so it should take a `SettlementStore`) and also act as a settlement engine client (so it would include a `SettlementClient`)?

I like the idea of separating the settlement client from the main store implementation. How would it know when the database operation is complete? Can we emit some event in place of the `spawn` call, and make the client subscribe to that event?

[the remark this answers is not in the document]
emschwartz: Yeah, instead of doing the check for settlement call at the same time as the process fulfill, there would be a separate service that does the check for settlement call to the DB. That separate service would have a Settlement Store like you say.

All we need to do is spawn a future that first checks the DB `and_then` sends the send settlement message if necessary (no events needed).

crates/interledger-api/src/lib.rs
```diff
@@ -164,4 +162,17 @@ where
...
     }
+
+    pub fn serve_settlement<I>(&self, incoming: I) -> impl Future<Item = (), Error = ()>
+    where
+        I: ConnectionStream,
+        I::Item: Send + 'static,
+    {
+        ServiceBuilder::new()
+            .resource(SettlementApi::new(
+                Arc::new(RwLock::new(self.store.clone())),
```
emschwartz: This shouldn't need to be wrapped in a `RwLock` because the `store` itself should implement `Clone`

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +22,557 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+struct SettlementData {
     amount: u64,
 }
 
-#[derive(Response)]
-#[web(status = "200")]
-struct Success;
+pub struct SettlementApi<S, O, A> {
+    outgoing_handler: O,
+    store: Arc<RwLock<S>>,
```
emschwartz:
```suggestion
    store: S,
```

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +22,557 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+    account_type: PhantomData<A>,
+}
 
+macro_rules! clone_all {
```
emschwartz: What is this for? I don't think we need to be writing macros for this

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +22,557 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
-                    if let Some(settlement_engine) = account.settlement_engine_details() {
-                        Ok((account, settlement_engine))
+            // These clones are very ugly (but cheap since this is an Arc). They
+            // are required due to the multiple moves inside the futures chain.
+            // Can we remove them in some smart way?
+            // https://github.com/rust-lang/rfcs/issues/2407
+            // https://users.rust-lang.org/t/automatic-cloning-for-closures/2578
+            // Tried `enclose` macro, doesn't behave well with future chains
+
+            let store = Arc::clone(&self.store);
```
emschwartz: FWIW you can also do `self.store.clone()`

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +22,557 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+                    result(A::AccountId::from_str(&account_id)
+                    .map_err({ clone_all!(store, idempotency_key); move |_err| {
+                        let error_msg = format!("Unable to parse account id: {}", account_id);
+                        error!("{}", error_msg);
+                        let status_code = StatusCode::from_u16(400).unwrap();
+                        let data = Bytes::from(error_msg.clone());
+                        store.write().save_idempotent_data(idempotency_key, status_code, data);
+                        Response::builder().status(400).body(error_msg).unwrap()
+                    }}))
+                    .and_then({clone_all!(store, idempotency_key); move |account_id| {
```
emschwartz: Not sure I'm a fan of using the macro here, is this just to avoid putting the clones up top?

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +22,557 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+                let err = format!("Couldn't connect to store {:?}", err);
+                error!("{}", err);
+                Response::builder().status(500).body(err).unwrap()
+            }).and_then(move |data: Option<(StatusCode, Bytes)>| {
+                if let Some(d) = data {
+                    if d.0.is_success() {
+                        let data = Response::builder().status(d.0).body(d.1).unwrap();
+                        return Either::A(Either::A(ok(data)));
+                    } else {
+                        let body = String::from_utf8(d.1.to_vec()).unwrap(); /// if success return bytes otherwise return string
```
emschwartz: That's not great. Ideally the response type is always the same

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +22,557 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+        // until it reaches the peer's settlement engine. Extract is not
+        // implemented for Bytes unfortunately.
+        #[post("/accounts/:account_id/messages")]
+        fn send_outgoing_message(&self, account_id: String, body: Vec<u8>, idempotency_key: String)-> impl Future<Item = Response<Bytes>, Error = Response<String>> {
+            let store = self.store.clone();
+            let mut outgoing_handler = self.outgoing_handler.clone();
+
+            // Check store for idempotency key. If exists, return cached data
+            let s = self.store.read();
+            s.load_idempotent_data(idempotency_key.clone())
```
emschwartz: Can the idempotency functionality be extracted into its own function?

crates/interledger-settlement/src/test_helpers.rs
```diff
@@ -0,0 +1,239 @@
...
+            .build()))
+        }),
+    )
+}
+
+pub fn test_store(store_fails: bool, account_has_engine: bool) -> Arc<RwLock<TestStore>> {
+    let mut acc = TEST_ACCOUNT_0.clone();
+    acc.no_details = !account_has_engine;
+
+    Arc::new(RwLock::new(TestStore::new(vec![acc], store_fails)))
```
emschwartz: The `TestStore` should just wrap an `Arc<RwLock>`

crates/interledger-store-redis/src/store.rs
```diff
@@ -105,8 +107,20 @@ return balance";
 static PROCESS_INCOMING_SETTLEMENT: &str = "
 local account = 'accounts:' .. ARGV[1]
 local amount = tonumber(ARGV[2])
+local idempotency_key = ARGV[3]
+
 local balance, prepaid_amount = unpack(redis.call('HMGET', account, 'balance', 'prepaid_amount'))
 
+-- If idempotency key has been used, then do not perform any operations
+local exists = redis.call('GET', idempotency_key)
```
emschwartz: Use the command `EXISTS` instead

crates/interledger-store-redis/src/store.rs
```diff
@@ -105,8 +107,20 @@ return balance";
...
+if exists == '1' then
+    return balance + prepaid_amount
+end
+
+-- Otherwise, set it to true (there's no value to cached as a response)
+-- and make it expire after 24h (86400 sec)
+redis.call('SET', idempotency_key, '1', 'EX', 86400)
```
emschwartz: Any reason not to use `true` instead?

crates/interledger-store-redis/src/store.rs
```diff
@@ -1134,16 +1148,81 @@ impl RateLimitStore for RedisStore {
...
+                .map_err(move |err| {
+                    error!(
+                        "Error loading idempotency key {}: {:?}",
+                        idempotency_key_clone, err
+                    )
+                })
+                .and_then(move |(_connection, ret): (_, Vec<String>)| {
+                    // even elements represent the key
+                    // odd elements represent the value
+                    // there must be 4 elements, a status code and a data value
```
emschwartz: You could replace `Vec<String>` with a `HashMap<String, String>` or a type that represents exactly what you're saving to the DB so that the redis library does the parsing and we don't need to have a comment like this

crates/interledger-store-redis/src/store.rs
```diff
@@ -1134,16 +1148,81 @@ impl RateLimitStore for RedisStore {
...
+        &self,
+        idempotency_key: String,
+        status_code: StatusCode,
+        data: Bytes,
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        let mut pipe = redis::pipe();
+        pipe.atomic()
+            .hset(idempotency_key.clone(), "status_code", status_code.as_u16())
+            .ignore()
+            .hset(idempotency_key.clone(), "data", data.as_ref())
```
emschwartz: For multiple `HSET`s you can use `HMSET`

crates/interledger-settlement/src/api.rs
gakonst: The thing is, the clones are 1 per closure that moves the context, so we end up needing to clone 5/6 times at the top of the file, which I did not like, hence the macro.

Take a look here also:
https://github.com/rust-lang/rfcs/issues/2407
https://users.rust-lang.org/t/automatic-cloning-for-closures/2578

crates/interledger-settlement/src/api.rs
gakonst: Yeah, I just went with the [docs](https://doc.rust-lang.org/std/sync/struct.Arc.html#cloning-references) suggestion

> The Arc::clone(&from) syntax is the most idiomatic because it conveys more explicitly the meaning of the code. In the example above, this syntax makes it easier to see that this code is creating a new reference rather than copying the whole content of foo.

crates/interledger-settlement/src/api.rs
gakonst: The clone macro usage was in order to avoid cloning in multiple places manually, we can instead use the approach done in this commit: https://github.com/emschwartz/interledger-rs/pull/107/commits/e0a32959ede8c72a3518b4dcf1eb29d4a75c465a

crates/interledger-settlement/src/api.rs
gakonst: Fixed in https://github.com/emschwartz/interledger-rs/pull/107/commits/b3444a8df1c01ae3dfb58a8c0109f15145b87b9c, what do you think?

crates/interledger-settlement/src/api.rs
gakonst: WDYT https://github.com/emschwartz/interledger-rs/pull/107/commits/a02578bc251acd68cfcb1ad172b07c4ec5ca901a?

crates/interledger-api/src/lib.rs
```diff
@@ -164,4 +160,17 @@ where
             ))
             .serve(incoming)
     }
+
+    pub fn serve_settlement<I>(&self, incoming: I) -> impl Future<Item = (), Error = ()>
```
emschwartz: Might be worth putting this in the settlement crate instead

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +20,565 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+        ) -> impl Future<Item = Option<(StatusCode, Bytes)>, Error = (StatusCode, Bytes)> {
+            self.store.load_idempotent_data(idempotency_key.clone())
+            .map_err(move |err| {
+                let err = format!("Couldn't connect to store {:?}", err);
+                error!("{}", err);
+                (StatusCode::from_u16(500).unwrap(), Bytes::from(err))
+            }).and_then(move |ret: Option<(StatusCode, Bytes)>| {
+                    if let Some(d) = ret {
+                        if d.0.is_success() {
+                            return Either::A(Either::A(ok(Some((d.0, d.1)))))
```
emschwartz: nit: you don't need the `Either`, `ok`, and `err` here. You can just return a `Result` from this closure (it implements `IntoFuture`)

crates/interledger-settlement/src/api.rs
```diff
@@ -19,143 +20,565 @@ static PEER_PROTOCOL_CONDITION: [u8; 32] = [
...
+                    }}).and_then({
+                        let store = store.clone();
+                        let idempotency_key = idempotency_key.clone();
+                        move |_| {
+                        let ret = Bytes::from("Success");
+                        store.save_idempotent_data(idempotency_key, StatusCode::OK, ret.clone())
+                        .map_err(move |err| {
+                            let err = format!("Couldn't connect to store {:?}", err);
+                            error!("{}", err);
+                            Response::builder().status(500).body(Bytes::from(err)).unwrap()
```
emschwartz: I know I said we should have a consistent response type but now I think it might be worse. What is common practice if you want to return binary data on success but have a text or JSON error? It's less helpful to turn that text into JSON because then it a) cannot be read b) it's not clear that it's actually a text-based error message c) could be confused with whatever type of binary data would be sent back on success

crates/interledger-settlement/src/client.rs
```diff
@@ -29,39 +23,32 @@ impl SettlementClient {
     pub fn send_settlement<A: SettlementAccount + IldcpAccount>(
         &self,
         account: A,
-        amount: u64,
+        amount: u64, // TODO: Might change to the RFC's Quantity value which contains `scale` as well depending on the decision
```
emschwartz: I think you can take this TODO out and just change it depending on what we decide

crates/interledger-settlement/src/message_service.rs
```diff
@@ -49,96 +48,70 @@ where
...
-                        settlement_engine_url
-                            .path_segments_mut()
-                            .expect("Invalid settlement engine URL")
-                            .push("receiveMessage"); // Maybe set the idempotency flag here in the headers
-                        return Box::new(self.http_client.post(settlement_engine_url)
-                        .json(&message)
-                        .send()
-                        .map_err(move |error| {
-                            error!("Error sending message to settlement engine: {:?}", error);
+                // TODO add auth
```
emschwartz: I think we can remove these TODOs if we're not going to add auth

crates/interledger-settlement/src/message_service.rs
emschwartz: Looks like idempotency from the node -> SE still needs to be added

crates/interledger-store-redis/src/store.rs
```diff
@@ -1134,16 +1149,89 @@ impl RateLimitStore for RedisStore {
...
+                        "Error loading idempotency key {}: {:?}",
+                        idempotency_key_clone, err
+                    )
+                })
+                .and_then(
+                    move |(_connection, ret): (_, SlowHashMap<String, String>)| {
+                        // even elements represent the key
+                        // odd elements represent the value
+                        // there must be 4 elements, a status code and a data value
+                        // stored there
```
emschwartz: Remove this comment

crates/interledger-store-redis/src/store.rs
```diff
@@ -1134,16 +1149,89 @@ impl RateLimitStore for RedisStore {
...
+    fn save_idempotent_data(
+        &self,
+        idempotency_key: String,
+        status_code: StatusCode,
+        data: Bytes,
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        let mut pipe = redis::pipe();
+        pipe.atomic()
+            .cmd("HMSET") // cannot use hset_multiple since data and status_code have different types
+            .arg(idempotency_key.clone())
```
emschwartz: I think you can pass `&idempotency_key` or `idempotency_key.as_str()` instead of cloning it here

crates/interledger-settlement/src/api.rs
gakonst: I'd be happy to revert that commit. I believe that associating a Success status code with a `Bytes` and an Error status code with a `String` from the API consumer side is a fine approach.

crates/interledger-service-util/src/expiry_shortener_service.rs
```diff
@@ -1,7 +1,7 @@
 use interledger_service::{Account, OutgoingRequest, OutgoingService};
 use std::time::Duration;
 
-pub const DEFAULT_ROUND_TRIP_TIME: u64 = 500; // milliseconds?
+pub const DEFAULT_ROUND_TRIP_TIME: u64 = 1000;
```
emschwartz: Why change this? This means that each hop will subtract 2 seconds (1 second for each side)

crates/interledger-service/src/lib.rs
```diff
@@ -36,14 +36,16 @@ use std::{
...
+
 /// The base trait that Account types from other Services extend.
 /// This trait only assumes that the account has an ID that can be compared with others.
 ///
 /// Each service can extend the Account type to include additional details they require.
 /// Store implementations will implement these Account traits for a concrete type that
 /// they will load from the database.
 pub trait Account: Clone + Send + Sized + Debug {
-    type AccountId: Eq + Hash + Debug + Display + Default + FromStr + Send + Sync + Copy;
+    type AccountId: Eq + Hash + Debug + Display + Default + FromStr + Send + Sync + Copy + Serialize;
```
emschwartz: Why is Serialize needed?

crates/interledger-settlement/src/api.rs
```diff
@@ -1,161 +1,592 @@
...
+                            return Err((d.0, String::from_utf8_lossy(&d.1).to_string()))
+                        }
+                    }
+                    Ok(None)
+                }
+            )
+        }
+
+        #[post("/accounts/:account_id/settlement")]
+        fn receive_settlement(&self, account_id: String, body: SettlementData, idempotency_key: String) -> impl Future<Item = Response<Bytes>, Error = Response<String>> {
```
emschwartz: Is the `idempotency_key` a required header? This won't be called if that key isn't passed in

crates/interledger-store-redis/src/store.rs
```diff
@@ -6,6 +6,8 @@ use futures::{
     Future, Stream,
 };
 use hashbrown::{HashMap, HashSet};
+
+use hyper::StatusCode;
```
emschwartz: I think you could just use the `http` crate directly if it's just for the `StatusCode` export

crates/interledger-store-redis/src/store.rs
```diff
@@ -105,8 +109,19 @@ return balance";
...
+local idempotency_key = ARGV[3]
+
 local balance, prepaid_amount = unpack(redis.call('HMGET', account, 'balance', 'prepaid_amount'))
 
+-- If idempotency key has been used, then do not perform any operations
+if redis.call('EXISTS', idempotency_key) == 1 then
+    return balance + prepaid_amount
+end
+
+-- Otherwise, set it to true (there's no value to cached as a response)
```
emschwartz: What does "there's no value to cached as a response" mean?

crates/interledger-stream/src/server.rs
```diff
@@ -132,14 +132,12 @@ where
...
-        let dest: &[u8] = dest.as_ref();
+        let destination = request.prepare.destination();
         let to = request.to.client_address();
-        if dest.as_ref().starts_with(to.as_ref()) {
-            if let Ok(shared_secret) = self
-                .connection_generator
-                .rederive_secret(&request.prepare.destination())
-            {
+        let dest: &[u8] = destination.as_ref();
+        let to_ref: &[u8] = to.as_ref();
```
emschwartz: nit: don't need to bind a variable to this, `to.as_ref()` can just be used on the next line

crates/interledger/src/main.rs
```diff
@@ -111,6 +111,7 @@ pub fn main() {
                     .arg(Arg::with_name("config")
                         .long("config")
                         .short("c")
+                        .takes_value(true)
```
emschwartz: doh!

crates/interledger-service/src/lib.rs
gakonst: Serialize is needed because the account id is included in the body of the request to the settlement engine during registration: https://github.com/emschwartz/interledger-rs/pull/107/commits/e31df2ef32687d4b62ac6abd406fee44b79194b6#diff-b6ee23b4722a9bdbc19e7bc41b67c64cR89

crates/interledger-settlement/src/api.rs
gakonst: If a call MUST be idempotent, then you cannot afford it to be called without the key. I believe it should be required.

gakonst: > The API supports idempotency for safely retrying requests without accidentally performing the same operation twice. 

from Stripe docs: https://stripe.com/docs/api/idempotent_requests

I read this as optional, so let's go with an Option here.

crates/interledger-settlement/src/lib.rs
```diff
@@ -38,12 +47,33 @@ pub trait SettlementAccount: Account {
...
         amount: u64,
+        idempotency_key: Option<String>,
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
+
+    /// Returns the API response that was saved when the idempotency key was used
+    /// Also returns a hash of the input data which resulted in the response
+    /// If the key was not used, it must return None
+    fn load_idempotent_data(
+        &self,
+        idempotency_key: Option<String>,
```
emschwartz: This function doesn't make sense with `idempotency_key: None` so I don't think it should take an `Option`

crates/interledger-settlement/src/lib.rs
```diff
@@ -38,12 +47,33 @@ pub trait SettlementAccount: Account {
...
+    ) -> Box<dyn Future<Item = Option<IdempotentData>, Error = ()> + Send>;
+
+    /// Saves the data that was passed along with the api request for later
+    /// The store MUST also save a hash of the input, so that it errors out on requests
+    fn save_idempotent_data(
+        &self,
+        idempotency_key: Option<String>,
```
emschwartz: Same as above

crates/interledger-store-redis/src/store.rs
```diff
@@ -6,22 +6,28 @@ use futures::{
...
-use interledger_settlement::{SettlementAccount, SettlementClient, SettlementStore};
+use interledger_settlement::{
+    IdempotentData, SettlementAccount, SettlementClient, SettlementStore,
+};
 use parking_lot::RwLock;
 use redis::{
     self, cmd, r#async::SharedConnection, Client, ConnectionInfo, PipelineCommands, Value,
 };
 use ring::{aead, hmac};
+use std::collections::HashMap as SlowHashMap;
```
emschwartz: As of Rust 1.36, `std::collections::HashMap` == `hashbrown::HashMap` so we can just remove the `hashbrown` crate from the dependencies

https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html

crates/interledger-store-redis/src/store.rs
```diff
@@ -105,8 +111,18 @@ return balance";
 static PROCESS_INCOMING_SETTLEMENT: &str = "
 local account = 'accounts:' .. ARGV[1]
 local amount = tonumber(ARGV[2])
+local idempotency_key = ARGV[3]
+
 local balance, prepaid_amount = unpack(redis.call('HMGET', account, 'balance', 'prepaid_amount'))
 
+-- If idempotency key has been used, then do not perform any operations
+if redis.call('EXISTS', idempotency_key) == 1 then
```
emschwartz: Will this cause problems if the `idempotency_key` was `None`?

gakonst: The implemented store returns without touching redis if the key was none. Nonetheless, I'll take the above comment and make the idempotent calls be non-optional. It was done primarily because it was simpler to make the option call internally, than add another layer of error checking inside the top level API call.

emschwartz:
Might be worth squashing some of the commits so that they're grouped into commits that each have a specific feature

PR #112, opened 2019-06-19
Minor CI fixes

emschwartz:
The BTP change fixes the issue that was causing panics in this run: https://circleci.com/gh/emschwartz/interledger-rs/511

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,6 @@ jobs:
           command: cargo test --all --all-features
           environment:
             RUST_BACKTRACE: "1"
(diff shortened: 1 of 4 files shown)


emschwartz:
@gakonst It seems like Circle only allows you to configure a `no_output_timeout` rather than a fixed amount of time. Unfortunately adding logging to the tests means that there is constant output (from the route broadcaster) so the `no_output_timeout` won't be hit. I think we might need to disable the logging because of this and only enable it when we're specifically trying to figure something out

gakonst:
I'm fine with disabling the logging in the test

Issue #132, opened 2019-07-15
Remove `extern crate` directives since we are in Rust 2018


gakonst:
> I am pretty sure you don't need all these `extern crate` directives in rust 2018.

https://doc.rust-lang.org/nightly/edition-guide/rust-2018/module-system/path-clarity.html

_Originally posted by @zmanian in https://github.com/emschwartz/interledger-rs/pull/125_

bstrie:
This is almost done; tower-web is a sticking point due to some deficiencies in its procedural macro expansions. See also #146

emschwartz:
Can this be closed now?

gakonst:
Let's keep this open until we replace tower-web with warp in the Settlement and Settlement Engines crates, for some reason doing [`use tower_web::impl_web` does not work. ](https://github.com/interledger-rs/interledger-rs/blob/0835049bebcdd0a4aceda1a1eba378ddde03ff70/crates/interledger-settlement-engines/src/api.rs#L25)

bstrie:
This specific issue was fixed in #400, so I'll close this. Removing the tower_web dependency entirely can fall under #140.

PR #145, opened 2019-07-18
Remove all `extern crate` except for tower-web

bstrie:
Progress towards #132. It turns out that tower-web is afflicted by a bug that makes it compatible only with the old-style `#[macro_use]` import, since that approach brought everything imaginable into the root scope and so it only happened to work by serendipity. The fix on their end would probably look something like https://github.com/graphql-rust/graphql-client/pull/180 (in general I think this is related to the note in the migration guide at https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#local-helper-macros ). I considered making a PR to tower-web but it has a few idling PRs and hasn't been updated in a few months, making me wonder whether we should look towards migrating to something else in the future (I'll file a separate issue for that). In any case, this style looks much more uniform, makes it easier to tell where identifiers are defined, and also makes it possible for the unused imports warning to work, which allowed me to outright eliminate some crates from some files.

### DIFF
--- a/crates/interledger-api/src/lib.rs
+++ b/crates/interledger-api/src/lib.rs
@@ -1,10 +1,6 @@
 #![recursion_limit = "128"]
 #[macro_use]
 extern crate tower_web;
-#[macro_use]
-extern crate log;
-#[macro_use]
-extern crate serde_json;
 
 use bytes::Bytes;
 use futures::Future;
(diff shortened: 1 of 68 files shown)


emschwartz:
On the question of whether to switch away from `tower-web`, https://github.com/emschwartz/interledger-rs/issues/39 also lists another potential motivation for using a different framework.

That said, `tower-web` is fairly easy to understand and use so I'm hesitant to switch to another framework that's more opinionated. The main downsides are the use of macros (which makes the errors produced from inside the `impl_web` block completely unintelligible) and if it's not being super actively maintained.

carllerche:
I have opened https://github.com/carllerche/tower-web/issues/218. tl;dr, I have no time right now to focus on tower-web. If anyone wishes to join the effort of development and maintenance, I can assist in mentoring and small things here and there.

PR #148, opened 2019-07-19
Dev 0 - Easing Local Developer Setup

Neil-Jubinville:
Added a docker folder at the root for docker configurations.

Inside the docker folder will be a docker-compose file that stands up two ILP nodes.   Additionally traefik  is supplied as a reverse proxy allowing URL names to be routed to the nodes.  

Simply rebuild your docker image, then cd into the docker folder and type` docker-compose up`. 

When you want to blow it away and start with two new nodes simply type `docker-compose down`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,14 +16,22 @@ EXPOSE 7770
 
 VOLUME [ "/data" ]
 ENV REDIS_DIR=/data
+ENV DISABLE_LOCALTUNNEL=true
 
 # Install SSL certs and Redis
 RUN apk --no-cache add \
     ca-certificates \
-    redis
+    redis \
+    nano
+
+# Update the redis unix socket location
+RUN sed -i 's|unixsocket /run/redis/redis.sock|unixsocket /tmp/redis.sock|g' /etc/redis.conf
+
+# Start the Redis Server with specified conf
+RUN redis-server /etc/redis-conf &
 
 # Install localtunnel
-RUN npm install localtunnel request request-promise-native
+RUN npm install localtunnel request request-promise-native ip
 
 # Build run script
 WORKDIR /usr/src
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,47 @@
+version: '3'
+
+services:
+  reverse-proxy:
+    image: traefik 
+    command: --api --docker 
+    ports:
+      - "80:80"     # The HTTP port
+      - "8080:8080" # The Traefik Web UI (enabled by --api)
+      - "7770:7770" # The ILP HTTP API port
+      - "7768:7768" # The ILP BTP port
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock # So that Traefik can listen to the Docker events
+  
+  ilp-node-1:
+    image: neil/interledger-rs
+    env_file: 
+      - ilpnode1.conf
+    labels:
+      - "traefik.frontend.rule=Host: ilpnode1.local"
+      - "traefik.enable=true"
+      - "traefik.port=7770"
+      - "traefik.protocol=http"
+      - "traefik.backend=ilpnode1"
+    networks:
+      default:
+        ipv4_address: 172.25.0.3
+
+  ilp-node-2:
+    image: neil/interledger-rs
+    env_file: 
+      - ilpnode2.conf
+    labels:
+      - "traefik.frontend.rule=Host: ilpnode2.local"
+      - "traefik.enable=true"
+      - "traefik.port=7770"
+      - "traefik.protocol=http"
+      - "traefik.backend=ilpnode2"
+    networks:
+      default:
+        ipv4_address: 172.25.0.4
+
+# below only works with userdefined networks so define one:  docker network create --subnet=172.25.0.0/24 ilpnetwork  
+networks:
+  default:
+    external:
+      name: ilpnetwork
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -16,14 +16,22 @@ EXPOSE 7770
 
 VOLUME [ "/data" ]
 ENV REDIS_DIR=/data
+ENV DISABLE_LOCALTUNNEL=true
```
emschwartz: Instead of disabling localtunnel like this, we should just update how the node is run via Docker.

I made the `run-interledger-node.js` script because I wanted to set up some of the account stuff automatically and expose the node easily via localtunnel. I would be okay removing the whole script and instead providing instructions for how to set up the accounts you need on the node.

Dockerfile
```diff
@@ -16,14 +16,22 @@ EXPOSE 7770
...
 
 # Install SSL certs and Redis
 RUN apk --no-cache add \
     ca-certificates \
-    redis
+    redis \
+    nano
```
emschwartz: Why is nano needed in the docker container?

Dockerfile
```diff
@@ -16,14 +16,22 @@ EXPOSE 7770
...
+
+# Update the redis unix socket location
+RUN sed -i 's|unixsocket /run/redis/redis.sock|unixsocket /tmp/redis.sock|g' /etc/redis.conf
```
emschwartz: We shouldn't use `sed` to modify the configuration. Instead, we should pass a custom config file to redis or pass the arguments we want directly to redis

Dockerfile
```diff
@@ -16,14 +16,22 @@ EXPOSE 7770
...
+
+# Start the Redis Server with specified conf
+RUN redis-server /etc/redis-conf &
```
emschwartz: I don't think this will be re-run if the container is stopped and started again

docker/docker-compose.yml
```diff
@@ -0,0 +1,47 @@
+version: '3'
+
+services:
+  reverse-proxy:
```
emschwartz: Do we need to use a reverse proxy just to get a 2 node demo going locally? Seems like it would be better to keep it local and just talk to the node via localhost or the ports exposed on local IP for the container

docker/ilpnode1.conf
```diff
@@ -0,0 +1,3 @@
+ILP_ADDRESS=test.node1
+ILP_ADMIN_TOKEN=nodeadmintoken
+RUST_BACKTRACE=0
```
emschwartz: If you want helpful logs, you might also want to include `RUST_LOG=interledger/.*`

Dockerfile
Neil-Jubinville: Yes,  I think this should be the approach for getting started.

Dockerfile
Neil-Jubinville: Developers often need to go into the docker and tweak or test configurations or environments.   Nano is a popular lightweight simple editor for making such changes.    I would say this is acceptable for dev but CI/CD builds for release or production should be stripped down to only what is necessary.

Dockerfile
Neil-Jubinville: The socket location was otherwise crashing the startup.   I did not dig into why but it could have been related to the run-interlegder-node.js

Dockerfile
Neil-Jubinville: yeap.

Dockerfile
emschwartz: I would prefer not to ship with an editor, especially because everyone prefers something different. If you need to go into an image to edit something, you can just use `docker exec` to install your editor of choice.

Dockerfile
emschwartz: It's fine to change the socket location but it shouldn't be done with `sed`. The correct configuration should just be passed to redis directly

docker/docker-compose.yml
Neil-Jubinville: I am not thinking 2.   I am starting with 2 :).    

I think this is a very useful setup.   It is as you know optional in its use.   If one chooses to go by IP then that is obvious, use IPs.
   
If a dev wants name based resolution it is there and works integrated with docker. and more universal for discussion / community.   I am thinking of a developer experience.  Also https will need name based routing.

The alternative is devs will need to be more advanced in their networking and configure their computers,  we can avoid this baking in a simple reverse proxy .

docker/ilpnode1.conf
Neil-Jubinville: Agreed, should we consider passing a "DEFAULT_STORE"  ,  I am thinking of that Redis Config fix  .    Just wondering if we should plan some abstraction here.

Dockerfile
Neil-Jubinville: Certainly, I agree it should be removed once we target the underlying crash.

Dockerfile
Neil-Jubinville: Ok, sounds good.

emschwartz:
I added a simple 2-node example in https://github.com/emschwartz/interledger-rs/pull/157 that involves running the nodes directly from Rust. I think I'd prefer to leave out the Docker instructions for now until we have a proper plan for bundling the services and keeping the published Docker image up to date. Let me know what you think of that PR!

gakonst:
Is this still being worked on?

emschwartz:
I'm going to close this for now but we should revisit the docker and potentially docker compose deployment options later

PR #150, opened 2019-07-22
Ethereum Settlement Engine improvements

gakonst:
Fixes most of https://github.com/emschwartz/interledger-rs/issues/149

TODO: Add more tests for SPSP payments in `interledger` similar to `three_nodes.rs`

### DIFF
--- a/crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
+++ b/crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
@@ -454,6 +454,7 @@ where
         tx_hash: H256,
     ) -> impl Future<Item = (), Error = ()> {
         let mut url = self.connector_url.clone();
+        let account_id_clone = account_id.clone();
         url.path_segments_mut()
             .expect("Invalid connector URL")
             .push("accounts")
@@ -490,8 +491,8 @@ where
             ExponentialBackoff::from_millis(10).take(MAX_RETRIES),
             action,
         )
-        .map_err(|_| {
-            error!("Exceeded max retries when notifying connector. Please check your API.")
+        .map_err(move |_| {
+            error!("Exceeded max retries when notifying connector about account {:?} for amount {:?} and transaction hash {:?}. Please check your API.", account_id_clone, amount, tx_hash)
         })
     }
 
@@ -649,10 +650,7 @@ where
                 })
                 .and_then(move |resp| {
                     parse_body_into_payment_details(resp).and_then(move |payment_details| {
-                        let data = prefixed_mesage(
-                            payment_details.to.own_address.to_vec(),
-                            challenge_clone,
-                        );
+                        let data = prefixed_mesage(challenge_clone);
                         let challenge_hash = Sha3::digest(&data);
                         let recovered_address = payment_details.sig.recover(&challenge_hash);
                         debug!("Received payment details {:?}", payment_details);
@@ -712,7 +710,7 @@ where
             "Responding with our account's details {} {:?}",
             account_id, address
         );
-        let data = prefixed_mesage(address.own_address.to_vec(), body.clone());
+        let data = prefixed_mesage(body.clone());
         let signature = self.signer.sign_message(&data);
         let resp = {
             let ret = PaymentDetailsResponse::new(address, signature);
@@ -777,9 +775,8 @@ fn parse_body_into_payment_details(
         })
 }
 
-fn prefixed_mesage(data: Vec<u8>, challenge: Vec<u8>) -> Vec<u8> {
+fn prefixed_mesage(challenge: Vec<u8>) -> Vec<u8> {
     let mut ret = ETH_CREATE_ACCOUNT_PREFIX.to_vec();
-    ret.extend(data);
     ret.extend(challenge);
     ret
 }
@@ -915,7 +912,7 @@ mod tests {
         let bob: TestAccount = BOB.clone();
 
         let challenge = Uuid::new_v4().to_hyphenated().to_string().into_bytes();
-        let signed_challenge = prefixed_mesage(ALICE.address.to_vec(), challenge.clone());
+        let signed_challenge = prefixed_mesage(challenge.clone());
 
         let signature = ALICE_PK.clone().sign_message(&signed_challenge);
 
--- a/crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
+++ b/crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
@@ -51,7 +51,7 @@ impl PaymentDetailsResponse {
 /// functions are exposed via the Settlement Engine API.
 ///
 /// It requires a `confirmations` security parameter which is used to ensure
-/// that all transactions that get credited to the connector have sufficient
+/// that all transactions that get sent to the connector have sufficient
 /// confirmations (suggested value: >6)
 ///
 /// All settlements made with this engine make on-chain Layer 1 Ethereum
@@ -270,7 +270,7 @@ where
                 // get the safe number of blocks to avoid reorgs
                 let fetch_until = current_block - confirmations;
                 // U256 does not implement IntoFuture so we must wrap it
-                Ok((Ok(fetch_until), store.load_recently_observed_data()))
+                Ok((Ok(fetch_until), store.load_recently_observed_block()))
             })
             .flatten()
             .and_then(move |(fetch_until, last_observed_block)| {
@@ -316,7 +316,7 @@ where
                     debug!("Transactions settled {:?}", ret);
                     // now that all transactions have been processed successfully, we
                     // can save `fetch_until` as the latest observed block
-                    store_clone.save_recently_observed_data(fetch_until)
+                    store_clone.save_recently_observed_block(fetch_until)
                 })
             })
     }
@@ -337,10 +337,10 @@ where
         };
         let amount = transfer.amount;
         store
-            .check_tx_credited(transfer.tx_hash)
+            .check_if_tx_processed(transfer.tx_hash)
             .map_err(move |_| error!("Error when querying store about transaction: {:?}", tx_hash))
-            .and_then(move |credited| {
-                if !credited {
+            .and_then(move |processed| {
+                if !processed {
                     Either::A(
                         store
                             .load_account_id_from_address(addr)
@@ -350,7 +350,7 @@ where
                             .and_then(move |_| {
                                 // only save the transaction hash if the connector
                                 // was successfully notified
-                                store.credit_tx(tx_hash)
+                                store.mark_tx_processed(tx_hash)
                             }),
                     )
                 } else {
@@ -401,11 +401,11 @@ where
         let web3 = self.web3.clone();
         let store = self.store.clone();
         let self_clone = self.clone();
-        // Skip transactions which have already been credited to the connector
-        store.check_tx_credited(tx_hash)
+        // Skip transactions which have already been processed by the connector
+        store.check_if_tx_processed(tx_hash)
         .map_err(move |_| error!("Error when querying store about transaction: {:?}", tx_hash))
-        .and_then(move |credited| {
-            if !credited {
+        .and_then(move |processed| {
+            if !processed {
                 Either::A(
                 web3.eth().transaction(TransactionId::Hash(tx_hash))
                 .map_err(move |err| error!("Could not fetch transaction data from transaction hash: {:?}. Got error: {:?}", tx_hash, err))
@@ -433,7 +433,7 @@ where
                         .and_then(move |_| {
                             // only save the transaction hash if the connector
                             // was successfully notified
-                            store.credit_tx(tx_hash)
+                            store.mark_tx_processed(tx_hash)
                         }))
                     } else {
                         // otherwise return an empty future
--- a/crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
+++ b/crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
@@ -8,7 +8,7 @@ use std::iter::FromIterator;
 
 use ethereum_tx_sign::web3::{
     api::Web3,
-    futures::future::{err, ok, result, Either, Future},
+    futures::future::{err, join_all, ok, result, Either, Future},
     futures::stream::Stream,
     transports::Http,
     types::{Address, BlockNumber, TransactionId, H256, U256},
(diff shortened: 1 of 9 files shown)


gakonst:
```
#!/bin/bash -eo pipefail
cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 33 security advisories (from /home/circleci/.cargo/advisory-db)
    Scanning Cargo.lock for vulnerabilities (346 crate dependencies)
error: Vulnerable crates found!

ID:	 RUSTSEC-2019-0011
Crate:	 memoffset
Version: 0.2.1
Date:	 2019-07-16
URL:	 https://github.com/Gilnaa/memoffset/issues/9#issuecomment-505461490
Title:	 Flaw in offset_of and span_of causes SIGILL, drops uninitialized memory of arbitrary type on panic in client code
Solution: upgrade to: >= 0.5.0
```

This is not part of our deps

tarcieri:
@gakonst what version of `cargo-audit` is installed? You may be hitting an old bug: https://github.com/RustSec/cargo-audit/issues/88

crates/interledger-settlement-engines/src/engines/ethereum_ledger/types.rs
```diff
@@ -46,18 +46,15 @@ pub trait EthereumStore {
...
         block: U256,
-        balance: U256,
     ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
 
-    /// Loads the latest saved block number and account balance
-    fn load_recently_observed_data(
-        &self,
-    ) -> Box<dyn Future<Item = (U256, U256), Error = ()> + Send>;
+    /// Loads the latest saved block number
+    fn load_recently_observed_data(&self) -> Box<dyn Future<Item = U256, Error = ()> + Send>;
```
emschwartz: If this is only returning the latest saved block number, the method should probably be renamed to be more specific

crates/interledger-settlement-engines/src/engines/ethereum_ledger/types.rs
```diff
@@ -68,9 +65,11 @@ pub trait EthereumStore {
         eth_address: Addresses,
     ) -> Box<dyn Future<Item = <Self::Account as Account>::AccountId, Error = ()> + Send>;
 
-    /// Errors out if the transaction hash has already been stored before,
-    /// otherwise saves the transaction hash in the store.
+    /// Errors out if the transaction hash has already been stored before
```
emschwartz: Is that true, or will it return false? (Or true? -- based on the method name it's not immediately clear to me what the return type would mean)

crates/interledger-settlement-engines/src/engines/ethereum_ledger/types.rs
```diff
@@ -68,9 +65,11 @@ pub trait EthereumStore {
...
     fn check_tx_credited(&self, tx_hash: H256) -> Box<dyn Future<Item = bool, Error = ()> + Send>;
+
+    /// Saves the transaction hash in the store.
+    fn credit_tx(&self, tx_hash: H256) -> Box<dyn Future<Item = (), Error = ()> + Send>;
```
emschwartz: Don't think "credit" is the right verb. It's almost more like "mark tx processed" or something like that

crates/interledger-settlement-engines/src/engines/ethereum_ledger/utils.rs
```diff
@@ -29,26 +30,97 @@ pub fn make_tx(
         data.extend(ethabi::encode(&[Token::Address(to), Token::Uint(value)]));
         RawTransaction {
             to: Some(token_address),
-            nonce,
+            nonce: U256::from(0),
```
emschwartz: How will the Ethereum provider handle this? If it does something special automatically it might be worth including a comment to describe what happens

crates/interledger-settlement-engines/src/engines/ethereum_ledger/utils.rs
```diff
@@ -29,26 +30,97 @@ pub fn make_tx(
...
+    pub amount: U256,
+}
+
+/// Filters out transactions where the `from` and `to` fields match the provides
+/// addreses.
+pub fn transfer_logs(
+    web3: Web3<Http>,
+    contract_address: Address,
+    from: Option<Address>,
+    to: Option<Address>,
```
emschwartz: Would the `from` and `to` ever be different?

Also, if this function doesn't really do anything if you don't pass in `Some` values for both fields, it seems like the function shouldn't take an `Option` and the caller of this function should figure out (based on whether those values are `Some` or `None` whether to call this function)

crates/interledger-settlement-engines/src/engines/ethereum_ledger/utils.rs
```diff
@@ -29,26 +30,97 @@ pub fn make_tx(
...
+    // create a filter for Transfer events from `from_block` until `to_block
+    // that filters all events indexed by `from` and `to`.
+    let filter = FilterBuilder::default()
+        .from_block(from_block)
+        .to_block(to_block)
+        .address(vec![contract_address])
+        .topics(
+            Some(vec![H256::from(
+                // keccak256("transfer(address,address,to)")
+                "ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
```
emschwartz: Shouldn't be hard-coded, right?

crates/interledger-settlement-engines/src/engines/ethereum_ledger/utils.rs
```diff
@@ -29,26 +30,97 @@ pub fn make_tx(
...
+        )
+        .build();
+
+    // Make an eth_getLogs call to the Ethereum node
+    web3.eth()
+        .logs(filter)
+        .map_err(move |err| error!("Got error when fetching transfer logs{:?}", err))
+        .and_then(move |logs| {
+            let mut ret = Vec::new();
+            for l in logs {
```
emschwartz: nit: no single-letter variable names

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -18,6 +18,7 @@ use crate::stores::redis_store::{EngineRedisStore, EngineRedisStoreBuilder};
 // avoid double crediting transactions which have already been processed, and in
 // order to resume watching from the last observed point.
 static RECENTLY_OBSERVED_DATA_KEY: &str = "recently_observed_data";
+static SAVED_TRANSACTIONS_KEY: &str = "transactions";
 static SETTLEMENT_ENGINES_KEY: &str = "settlement";
 static LEDGER_KEY: &str = "ledger";
```
emschwartz: What's the point of including this in the prefix?

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -37,10 +38,17 @@ impl AccountTrait for Account {
...
+        "{}:{}:{}:{}",
+        ETHEREUM_KEY, LEDGER_KEY, SAVED_TRANSACTIONS_KEY, tx_hash,
+    )
+}
+
 fn ethereum_ledger_key(account_id: u64) -> String {
     format!(
         "{}:{}:{}:{}",
-        SETTLEMENT_ENGINES_KEY, LEDGER_KEY, ETHEREUM_KEY, account_id
+        ETHEREUM_KEY, LEDGER_KEY, SETTLEMENT_ENGINES_KEY, account_id
```
emschwartz: I don't think this key needs to be quite so verbose. I think we can assume for the moment that the ETH engine is going to talk to a different Redis DB instance than another settlement engine so we don't need to ensure that the schemas don't conflict.

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -240,21 +232,40 @@ impl EthereumStore for EthereumLedgerRedisStore {
...
 
     fn check_tx_credited(&self, tx_hash: H256) -> Box<dyn Future<Item = bool, Error = ()> + Send> {
+        Box::new(
+            cmd("EXISTS")
+                .arg(ethereum_transactions_key(tx_hash))
+                .arg(true)
+                .query_async(self.connection.clone())
+                .map_err(move |err| error!("Error loading account data: {:?}", err))
+                .and_then(
+                    move |(_conn, ret): (_, u64)| {
```
emschwartz:
```suggestion
                    move |(_conn, ret): (_, bool)| {
```

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -240,21 +232,40 @@ impl EthereumStore for EthereumLedgerRedisStore {
     }
 
     fn check_tx_credited(&self, tx_hash: H256) -> Box<dyn Future<Item = bool, Error = ()> + Send> {
+        Box::new(
+            cmd("EXISTS")
+                .arg(ethereum_transactions_key(tx_hash))
+                .arg(true)
```
emschwartz: What is this for?

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -240,21 +232,40 @@ impl EthereumStore for EthereumLedgerRedisStore {
...
+                            Ok(true)
+                        } else {
+                            Ok(false)
+                        }
+                    },
+                ),
+        )
+    }
+
+    fn credit_tx(&self, tx_hash: H256) -> Box<dyn Future<Item = (), Error = ()> + Send> {
```
emschwartz: Is there a possibility that separating `credit_tx` from `check_tx_credited` could cause a race condition? Is it possible that two threads or tasks could call `check_tx_credited` at the same time, both get a return value that indicates that it hasn't been credited, and both credit the balance accordingly?

tarcieri:
Re: RUSTSEC-2019-0011

What's actually happening is we retroactively updated the advisory to note that the older versions of `memoffset` are actually vulnerable too. It's been updated to note the older versions are also vulnerable now:

https://rustsec.org/advisories/RUSTSEC-2019-0011.html

It's been one of the more controversial vulns we've added to the database in that it falls into the "high severity / low exploitability" category. However, I would say the more we investigated what could go wrong based on particular patterns of misuse, the severity of the vulnerability continued to escalate, to first memory exposure and then to remove code execution.

If you feel comfortable with the current usage patterns, note that `cargo-audit` v0.7.0 added a `--ignore` option you can use to ignore particular advisories. This one is a bit annoying in that it has impacted a lot of people, and there hasn't been a particularly good resolution path for many.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/utils.rs
gakonst: Our engine does not actually care to filter for `from`, we only care that `to` was our address.

If the fields are left as `None`, then the filter will not limit itself to any parameters and will log all `Transfer` events. 

The arguments were made optional for the sake of flexibility, I'm happy to remove them if you think it's redundant. It could be extracted in `ethereum-utils` helpers in the future which would be useful for payment channels, where the optional parameters would make sense.

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
gakonst: I was not making that assumption, hence why I added all the prefixes. I also like it because if the connector uses the same redis instance for their connector store, these prefixes allow the engine keys to be nicely grouped together.

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
gakonst: We save transactions which have already been processed by the connector by their transaction hash. By prefixing the key with `transactions` we can easily inspect which transactions have already been processed, otherwise it'd be cumbersome.

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
gakonst: Typo, thanks.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/types.rs
gakonst: Now called `recently_observed_block`

crates/interledger-settlement-engines/src/engines/ethereum_ledger/types.rs
gakonst: Comment has been adjusted

crates/interledger-settlement-engines/src/engines/ethereum_ledger/types.rs
gakonst: Function has been renamed

crates/interledger-settlement-engines/src/engines/ethereum_ledger/utils.rs
gakonst: Added comments for clarity. The transaction's fields are populated by the client after probing the Ethereum node about what their values should be

crates/interledger-settlement-engines/src/engines/ethereum_ledger/utils.rs
gakonst: It should. This hash is the value of `keccak256("transfer(address,address,to)")`, which is how you install filters in Ethereum.

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
gakonst: It does not make sense to set `poll_frequency` to less than 60 seconds in Ethereum, since block time is targeted at 15 sec, and you want a buffer of some blocks to avoid reorgs. So I doubt it can happen in that case.

I think you are right though that if the function gets called frequently enough for the same range of blocks this can happen.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -259,160 +230,267 @@ where
...
-                            // are not fetched, OR process them serially, and on failure
-                            // save the block and quit.
-                            tokio::spawn(store.save_recently_observed_data(fetch_until, new_balance))
-                        }})
+            })
+    }
+
+    /// Submits an ERC20 transfer object's data to the connector
+    // todo: Try combining the body of this function with `submit_tx_to_connector`
+    fn submit_erc20_transfer(
```
emschwartz: This name seems ambiguous because this engine submits transactions to the ledger and notifies the connector of incoming transfers. This sounds like it would submit the transfer to the ledger. I would recommend using "notfiy", "notify connector of incoming X", or "process incoming X" as ways to describe what you do with incoming transfers and "send" or "submit" as words to describe outgoing transfers.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -616,7 +709,9 @@ where
             "Responding with our account's details {} {:?}",
             account_id, address
         );
-        let signature = self.signer.sign_message(&body.clone());
+        let mut data = body.clone();
+        data.extend(address.own_address.iter());
```
emschwartz: The magic bytes should come first and should include an identifier of the protocol this signature is being used for rather than just the account it corresponds to

gakonst:
re security advisory: https://github.com/tokio-rs/tokio/issues/1345

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -259,160 +231,267 @@ where
...
+        store
+            .check_if_tx_processed(transfer.tx_hash)
+            .map_err(move |_| error!("Error when querying store about transaction: {:?}", tx_hash))
+            .and_then(move |processed| {
+                if !processed {
+                    Either::A(
+                        store
+                            .load_account_id_from_address(addr)
+                            .and_then(move |id| {
+                                self_clone.notify_connector(id.to_string(), amount.low_u64())
```
emschwartz: It's pretty bad if this fails even after retrying multiple times. I would at least log a serious sounding error with the details of the incoming settlement that it was unable to credit

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -259,160 +231,267 @@ where
...
+    /// 5. Save (current block number - confirmations) to be used as
+    ///    last observed data for the next call of this function.
+    // Optimization: This will make 1 settlement API request to
+    // the connector per transaction that comes to us. Since
+    // we're scanning across multiple blocks, we want to
+    // optimize it so that: it gathers a mapping of
+    // (accountId=>amount) across all transactions and blocks, and
+    // only makes 1 transaction per accountId with the
+    // appropriate amount.
     pub fn notify(&self) -> impl Future<Item = (), Error = ()> + Send {
```
emschwartz: How about:
```suggestion
     pub fn handle_received_transactions(&self) -> impl Future<Item = (), Error = ()> + Send {
```

This is a minor point but "notify" makes me wonder what it's notifying about

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
emschwartz: Can we use the tx hash as the idempotency ID to ensure that the same transaction will never be processed by the connector twice? (As it is now, if the engine happened to crash after submitting the tx to the connector but before saving that fact to the DB, that transaction would be resent later)

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -259,160 +231,267 @@ where
...
+                                store.mark_tx_processed(tx_hash)
+                            }),
+                    )
+                } else {
+                    Either::B(ok(())) // return an empty future otherwise since we want to skip this transaction
+                }
+            })
+    }
+
+    fn submit_txs_in_block(&self, block_number: u64) -> impl Future<Item = (), Error = ()> {
```
emschwartz: nit: consistent naming with the other functions

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -616,7 +712,8 @@ where
             "Responding with our account's details {} {:?}",
             account_id, address
         );
-        let signature = self.signer.sign_message(&body.clone());
+        let data = prefixed_mesage(address.own_address.to_vec(), body.clone());
```
emschwartz: What's the point of including our address in here?

gakonst: Verifier sends `challenge` to Prover
Prover signs `unique_identifier` (+ `address`) + `challenge` and sends `signature` to prover along with `address`
Verifier checks that the sig on (`unique_identifier` (+ `address`) + `challenge`) is from `address`.
 
Does removing the address from the data that's being signed affect security? Doesn't look like it.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
gakonst: The current error in line 495 is: Exceeded max retries when notifying connector. Please check your API.

Do you think it is not enough?

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
emschwartz: I don't think so. Including the address shouldn't matter because the key used for that signature is linked to that address

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
emschwartz: I guess the effect of this would be that it would not update the last block seen so all of these transactions would be tried again. If that's the case, it's not the biggest deal. That said, it would probably be good to log some of the tx details so that if you're manually looking through to see what happened you can identify which transaction it failed at

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
gakonst: Done in #1f0d86c

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
gakonst: slightly more verbose error in https://github.com/emschwartz/interledger-rs/commit/1f0d86c3b94b30a74ee564b1848f7cd74e0b9be1

Issue #81, opened 2019-06-04
Rename Interledger.rs?


emschwartz:
This project could use a catchier name and one that rolls off the tongue better than "Interledger dot r s". Anyone have suggestions?

tarcieri:
First thing that came to my mind was some play on "Iron Bank" (e.g. Braavos)

ekrenzke:
I was calling my zmq implementation "Mantle" but I have not had the bandwidth to work on it much lately. 

I am a big fan of double meanings. From Google:

1. an important role or responsibility that passes from one person to another.

2. the region of the earth's interior between the crust and the core, believed to consist of hot, dense silicate rocks (mainly peridotite).

The downside being "Mantle" is arguably used too often in software engineering.

emschwartz:
Another option to throw into the mix: "Alloy" (metal theme + mixing different things and making them all stronger + flowing liquid to go along with streaming payments)

gakonst:
What stainless alloys exist?

Since the language is Rust, we may want some theme around not being oxidised/corroded (corroded = bad code/unmaintained), around a "stainless alloy + liquid" mix?

tarcieri:
Stainless steel is an alloy, FWIW...

matdehaast:
Forge

dora-gt:
`Poseidon` Inspired by...

- When once the ILP gets spread all over the world, money will be moving fast like a rapid current of water.
- It seems like a big ocean of money.
- Poseidon controls all the seas, `Interledger.rs` does.
- Crustaceans swim in the seas😂

bstrie:
Since Rust's name is, obviously, based on [the fungus](https://en.wikipedia.org/wiki/Rust_(fungus)), I propose we take a page from biology. In particular fungal colonies are connected by underground *networks* named [mycelium](https://en.wikipedia.org/wiki/Mycelium), which is a bit of a mouthful, but a punchy synonym for "mycelium" is **shiro**, which rolls off the tongue nicely. :)

emschwartz:
A friend made the point that Interledger.rs (or Interledger Rust) is the clearest regarding what it is. Other names might be more fun but don't contain as much information, especially for new people about what this project actually does. Worth considering

emschwartz:
[...]

PR #157, opened 2019-07-25
Update instructions for trying out Interledger.rs

emschwartz:
Resolves https://github.com/emschwartz/interledger-rs/issues/153

A couple of points worth mentioning:
- ~~I took out the `Dockerfile` because it's currently out of date and we need to rethink how we're bundling the various services (the `node`, Redis, settlement engines) and we need a CD pipeline to make sure the published Docker image stays up to date when we have it~~
- Instead of detailing all of the steps for the example in its README, I opted to break out the different steps into bash scripts. I thought this might make it easier to keep the example up to date (because we don't have code copied into the markdown file), but it might be slightly harder to read. Let me know what you think!
- We should get the other examples written up soon!

Also, the example setup was inspired by @gakonst's work in https://github.com/emschwartz/interledger-rs/pull/125


README.md
```diff
@@ -8,46 +8,30 @@
...
-3. Follow the instructions written to the terminal to try sending an SPSP payment
-
-
-### Manual Install
+## Installation
 
 Prerequisites:
 - Git
-- Rust (using [`rustup`](https://rustup.rs/) is recommended)
+- Rust, latest stable version (using [`rustup`](https://rustup.rs/) is recommended)
```
bstrie: Just a note that I have seen some people who were unfamiliar with Rust express wariness over whether rustup.rs is an official Rust site, which I empathize with given the command that it asks them to run. I don't think it really matters much in this case, but I do tend to instead link people to https://www.rust-lang.org/tools/install for that reason.

examples/simple/README.md
```diff
@@ -0,0 +1,21 @@
+# Simple Two-Node Payment
+
+This example sets up two Interledger.rs nodes, connects them together, and sends a payment from one to the other.
+
+To run the full example, use the [`run-all.sh`](./run-all.sh) script. Otherwise, you can walk through each step below.
+
+Each of the services write their logs to files found under the `logs` directory. You can run `tail -f logs/node-a.log`, for example, to watch the logs of Node A.
+
```
dora-gt: I'll insert a simple image of the overview here, making a pull request later so that the reader could understand what's happening with `run-all.sh`.

examples/simple/README.md
```diff
@@ -0,0 +1,21 @@
+# Simple Two-Node Payment
+
+This example sets up two Interledger.rs nodes, connects them together, and sends a payment from one to the other.
```
dora-gt:
```suggestion
This example sets up two Interledger.rs nodes, connects them together, and sends a payment from one to the other. It means that your nodes connect to neither the outer production Interledger network nor the test net.
```

Those who are not familiar with the ILP might be relieved with this, being noticed that this script never affects the ILP network and they don't need to contact with the other nodes.

examples/README.md
```diff
@@ -0,0 +1,10 @@
+# Interledger.rs Examples
+
+Here you can find various demos of Interledger.rs' functionality:
+
+1. [Simple Two-Node Payment](./simple/README.md)
+1. Two-Node Payment with On-Ledger Settlement (Coming Soon!)
+1. Three-Node Payment (Coming Soon!)
+1. Cross-Currency Payment and Settlement (Coming Soon!)
+
```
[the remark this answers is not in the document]
Neil-Jubinville: I like an "Integrate ILP with your app" section.   It will be good to eventually show simple language / sdk examples.   Clients can be nicely abstracted, again no touch on ILP infrastructure.    

Clients
- python  ILP client
- java ILP client
- go ILP client
etc...

Servers/ Connectors
- python  ILP server
- java ILP server
- go ILP server
etc...

Neil-Jubinville:
I would suggest we have two additional headings to help classify information.

One would be "**Quickstart**"  where it links to the shortest fastest steps to prove out an ILP transaction, probably the two node example.

The other section would be "Deploying ILP Infrastructure",  This would have three main categories right away, and speak more in a devops infrastructure language:

- Deploy a Test Network Locally (2 or more nodes)
- Deploy a Test Network in Docker with a Reverse Proxy (2 or more nodes)
- Deploy a Test Network in Kubernetes (2 or more nodes)

examples/simple/README.md
Neil-Jubinville: Could we clarify "connects them together"  to be "automatically peers them together".   This gives a clue to what is going on and we can hyper link them to the future CCP.

examples/simple/README.md
emschwartz: Sounds great! I agree that would make it much easier to understand

examples/simple/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+To run the full example, use the [`run-all.sh`](./run-all.sh) script. Otherwise, you can walk through each step below.
```
pgarg-ripple: To run this script, or other scripts mentioned below, do we need a Redis server running? If yes, then we must mention that.

dora-gt: I mentioned it in this PR: https://github.com/emschwartz/interledger-rs/pull/167

emschwartz: @dora-gt is right, you need it installed but that script will run an instance of it itself

Neil-Jubinville:
Ok so I am testing the examples today to see how they go, thank-you @emschwartz for putting it together.

The developer experience could be improved here a bit. When running the first script of start-services.sh , the developer is left wondering why nothing is running.

Looking in the logs you will see NodeA is busy chunking away compiling every crate :) - NodeB runtime thread is blocked waiting for the other thread to yield crate cache control.

This takes a long time on my machine.

A) We figure out how to speed it up.
B) We notify the developer that the compiling is underway. Or echo to check the logs for completion. We could loop a health check and have it exit when both nodes are up.

examples/simple/README.md
```diff
@@ -0,0 +1,23 @@
+# Simple Two-Node Payment
+
+This example sets up two Interledger.rs nodes, connects them together, and sends a payment from one to the other.
+
+To run the full example, install [Redis](https://redis.io/) and then use the [`run-all.sh`](./run-all.sh) script. Otherwise, you can walk through each step below.
```
pgarg-ripple: This sounds like you need Redis installed only if you run the full example. But if you run the individual scripts, you don't need Redis installed on your system. 
I would suggest rewriting that section, like so:

"Note: To run this example, you need to install Redis on your computer.

To run the full example, use the run-all.sh script. Otherwise, you can walk through each step below."

bstrie:
@neil-jubinville One of my goals right now is to see what I can do to reduce compilation times overall. That said, we probably also do want to notify the developer that compilation is underway regardless.

emschwartz:
Good points @Neil-Jubinville about the compilation time. Since we could ship a really small Docker image that only contains the Rust binary and `redis-server`, maybe it would be worth doing that and setting these scripts up such that you can either run the code directly or point the script to use the Docker images instead. That would move the compilation time to CI instead of on the developer's machine

examples/simple/README.md
Neil-Jubinville: Agreed - I had redis already on my local system and actually commented the redis parts out.

Neil-Jubinville:
@emschwartz @bstrie  I am still new rust project / compilation setups, but do you know if we could bake the dependencies and just link it like a dll?    

When I saw how long the compile was taking I started searching for optimizations and did not find a solution.   The low hanging fruit here is probably the docker route.

bstrie:
@Neil-Jubinville we absolutely could directly distribute a platform-specific binary in order to skip compilation entirely, although we might need some infrastructure if we wanted to do it well. Alternatively we could just compile it once for typical Linux targets, host just that binary, and expect everyone else to fall back on compiling themselves.

KevinWMatthews:
This branch has been very useful to me - I was able to get up and running with no prior experience with interledger and see what was going on. Thanks!

As suggested above, it might be worth letting the user know when Rust is compiling. Maybe run `cargo build` in a separate step in `start-services.sh` without redirecting the output into a logfile?

It also might be good to add a note in the README that reminds the user to run `./kill-services.sh` after they are done experimenting. It would be easy to execute `./run-all.sh` and then forget that there are services to stop.

emschwartz:
That's great to hear @KevinWMatthews!

Good points about building and `kill-services`, I'll make those changes

README.md
```diff
@@ -8,46 +8,31 @@
...
 Prerequisites:
 - Git
-- Rust (using [`rustup`](https://rustup.rs/) is recommended)
+- [Rust](https://www.rust-lang.org/tools/install) - latest stable version
 
 Install and Run:
 
 1. `git clone https://github.com/emschwartz/interledger-rs && cd interledger-rs`
 2. `cargo build` (add `--release` to compile the release version, which is slower to compile but faster to run)
 2. `cargo run --package interledger` (append command line options after a `--` to use the CLI)
```
bstrie: Purpose of the `--package interledger` part? Currently this is the same as `cargo run`.

examples/simple/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+This example sets up two Interledger.rs nodes, connects them together, and sends a payment from one to the other.
+
+To run the full example, use the [`run-all.sh`](./run-all.sh) script. Otherwise, you can walk through each step below.
+
+Each of the services write their logs to files found under the `logs` directory. You can run `tail -f logs/node-a.log`, for example, to watch the logs of Node A.
+
+## Running and Configuring the Nodes
+
+The [`start-services.sh`](./start-services.sh) script runs a Redis instance and the two nodes, Node A and Node B.
```
bstrie: Since this page is looking to be the "hello world" a.k.a. the simplest possible demo of functionality, could we use the in-memory store rather than redis?

README.md
bstrie: If the goal is to keep people from being confused if they try to go from `cargo run` to `cargo run foo`, then perhaps we could change it to `cargo run -- --help`.

examples/simple/start-services.sh
```diff
@@ -0,0 +1,17 @@
...
+redis-server --port 6379 &> logs/redis.log &
+
+# Turn on debug logging for all of the interledger.rs components and tower-web
+export RUST_LOG=tower-web,interledger/.*
+
+# Start both nodes
+cargo run --package interledger -- node --config ./node-a/config.yml &> logs/node-a.log &
+cargo run --package interledger -- node --config ./node-b/config.yml &> logs/node-b.log &
+
+sleep 1
```
bstrie: If this whole script is intended to be interactive, it would be friendly to give some output upon success.

examples/simple/start-services.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Start Redis
+redis-server --version > /dev/null || echo "Uh oh! You need to install redis-server before running this example"
+redis-server --port 6379 &> logs/redis.log &
+
+# Turn on debug logging for all of the interledger.rs components and tower-web
+export RUST_LOG=tower-web,interledger/.*
+
+# Start both nodes
+cargo run --package interledger -- node --config ./node-a/config.yml &> logs/node-a.log &
```
bstrie: Spitballing but I imagine this could similarly be encapsulated in a command like `interledger-node new node-a -c ./node-a/config.yml`

examples/simple/start-services.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Create the logs directory if it doesn't already exist
+mkdir -p logs
+
+# Start Redis
+redis-server --version > /dev/null || echo "Uh oh! You need to install redis-server before running this example"
+redis-server --port 6379 &> logs/redis.log &
```
bstrie: Is the `--port 6379` not redundant with the redis.conf? I'm actually not sure if the redis.conf is getting used here, since I would expect its location to either be passed in via a command line switch or for it to live in /etc/ or something, rather than implicitly looking for a conf file in the pwd.

examples/simple/create-accounts.sh
```diff
@@ -0,0 +1,37 @@
...
+printf "\nNode A's account on Node B:\n"
+curl \
+    -H "Content-Type: application/json" \
+    -H "Authorization: Bearer admin-b" \
+    -d @./node-b/node-a.json \
+    http://localhost:8770/accounts
+
+sleep 1
+
+printf "\n\n"
```
bstrie: Output to confirm success.

bstrie:
Ideally I think it would be useful for all the functionality implemented within these bash scripts to be exposed (and thereby documented) by the `interledger-node` command. What I have in mind would make the "Simple Two-Node Payment" document look like so:

# Simple two-node payment

This example sets up two Interledger.rs nodes, connects them together, and sends a payment from one to the other.

Each of the services write their logs to files found under the logs directory. You can run `tail -f logs/node-a.log`, for example, to watch the logs of Node A.

First, initialize a new memory store (Redis, for this example):

    interledger-node store init redis -l logs/redis.log

...then initialize two new Interledger nodes locally:

    interledger-node init node-a -c ./node-a/config.yaml -l logs/node-a.log
    interledger-node init node-b -c ./node-b/config.yaml -l logs/node-b.log

...then set up one new account on each node, for Alice on Node A and for Bob on Node B:

    interledger-node acct new alice@node-a -c ./node-a/alice.json
    interledger-node acct new bob@node-b -c ./node-b/bob.json

...then we create two more accounts, an account for Node B on Node A and an account for Node A on Node B, which will allow A to send funds to B and vice versa:

    interledger-node acct new node-b@node-a -c ./node-a/node-b.json
    interledger-node acct new node-a@node-b -c ./node-b/node-a.json

...and now that nodes A and B are linked, Alice and Bob are also automatically linked, and so Alice can send funds to Bob as follows:

    interledger-node send 500 from alice@node-a to bob@node-b

We can check the balances of each account as follows:

    interledger-node balance alice@node-a
    interledger-node balance node-b@node-a
    interledger-node balance node-a@node-b
    interledger-node balance bob@node-b

[This would be much nicer if we also took this opportunity to split out and rename the interledger-node command, see also #81 :) ]

bstrie:
Likely unrelated to this PR, but if I repeatedly run send-payment.sh and check the balances after, the first two runs of the script work as expected, then for subsequent runs Alice's balance reduces by 300 each time whereas all other balance stay the same.

examples/simple/README.md
emschwartz: Not a bad idea. Right now though, the "node" isn't set up to use the in-memory store so maybe we'll make that change separately

examples/simple/start-services.sh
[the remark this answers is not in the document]
emschwartz: I'm not sure. There are open questions about how good a choice Redis in particular is and how flexible we want to be in supporting different databases. Right now everything is abstracted behind the "store" concept, though it's possible that a better design would use both a SQL DB and a Redis instance for persistence and caching, respectively, rather than using one of the two for everything.

examples/simple/start-services.sh
emschwartz: What would the init part do?

bstrie: In my mind the parsing syntax is always `main-command sub-command [arguments]`, since the alternative would be `main-command sub-command|node-id [arguments]`, and this just keeps us from accidentally breaking users who happen to give a node an ID that we later introduce as a subcommand. Alternatively we could say that nodes must always be written `@ident` (and accounts could then always be `user@ident`), which would make `interledger-node @node-a` unambiguously future-proof, but I think I still prefer the original form because there's probably more than one operation you can do on a node--like say `init` to create a local node configuration and `load` to load a stopped node whose configuration already exists in your user profile or whatever--so either way you'd probably want subcommands somewhere in the invocation.

bstrie: Might as well also mention that I envision the node configuration could be passed as further arguments to this init subcommand, so that if it's possible for us to have sensible automatic defaults for node initialization (and if we had a shorter alias than interledger-node) then this could be reduced to just `ilp init @example.node-a`, which is lovely and short. :)

examples/simple/start-services.sh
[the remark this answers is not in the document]
bstrie: If we're looking to have the most inoffensive defaults possible then perhaps we could ship SQLite as the default persistence layer and the in-memory store as the caching layer (although this introduces a new dependency on SQLite, but if we're dockerizing everything anyway then maybe we don't mind?).

examples/simple/start-services.sh
emschwartz: Let's continue the discussion about what the CLI should do in another issue (want to start one?)

examples/simple/start-services.sh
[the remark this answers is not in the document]
emschwartz: Agreed that SQLite would be a fine default. We should do https://github.com/emschwartz/interledger-rs/issues/52 first and then maybe switch the default to use that instead of Redis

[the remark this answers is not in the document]
emschwartz: Agreed that SQLite would be a fine default. Let's do https://github.com/emschwartz/interledger-rs/issues/52 first and then think about switching the default to that instead of Redis

Issue #160, opened 2019-07-25
PROCESS_FULFILL script doesn't adjust REDIS properly


sappenin:
In [store.rs](https://github.com/emschwartz/interledger-rs/blob/master/crates/interledger-store-redis/src/store.rs#L95), the `PROCESS_FULFILL` script is computing the settlement amount, but is not properly updating the `balance` after computation.

There needs to be a `redis.call('HSET', account, 'balance', balance)` call at line 96.

sappenin:
We should probably add unit tests for this as well.

sappenin:
We should also add some extra documentation into the script near line 92 to describe the rationale behind this:

```
-- Update the balance _before_ sending the settlement so that we don't accidentally send
-- multiple settlements for the same balance. While there will be a small moment of time (the delta 
-- between this balance change and the moment that the settlement-engine accepts the request for
-- settlement payment) where the actual balance in Redis is less than it should be, this is tolerable
-- because this amount of time will always be small. This is because the design of the settlement 
-- engine API is asynchronous, meaning when a request is made to the settlement engine, it will 
-- accept the request and return (milliseconds) with a guarantee that the settlement payment will
--  _eventually_ be completed. Because of this settlement_engine guarantee, the Connector can 
-- operate as-if the settlement engine has completed. Finally, if the request to the settlement-engine 
-- fails, this amount will be re-added back to balance.
```

sappenin:
This issues is related to #162

PR #164, opened 2019-07-25
Fix/Refactor Redis Store bugs + Settlement E2E Test + Settlement Client refactor

### LINKED ISSUE #116: Extract settlement call to separate service
> One option we might want to consider as an alternative to having a "settlement client" that's passed into the store is moving the settlement check to a different service but spawning a task to check instead of waiting on it before forwarding the request. This would fit better with the service abstraction and wouldn't require every store implementation to take an instance of the settlement client but it would add another DB call.

_Originally posted by @emschwartz in https://github.com/emschwartz/interledger-rs/pull/107#discussion_r296609944_

gakonst:
Fixes: https://github.com/emschwartz/interledger-rs/issues/162, 
Fixes: https://github.com/emschwartz/interledger-rs/issues/160
Fixes: #168 (partially, removes redundant args but does not split the service in an outgoing/incoming)
Fixes #116 

Also adds an E2E test for settlement with the ethereum ledger engine which is the equivalent of running the following example:
```
./settlement_test.sh
./pay_dump bob 10
./pay_dump bob 20
./pay_dump bob 40
./pay_dump bob 1
```

This will perform a 61 Wei settlement transaction on the running ganache backend, and the end result is Alice owing 10 Wei to Bob  (which is below their settle_threshold)


crates/interledger-store-redis/src/store.rs
```diff
@@ -86,13 +86,14 @@ local prepaid_amount, settle_threshold, settle_to = unpack(redis.call('HMGET', t
 
 -- Check if we should send a settlement for this account
 local settle_amount = 0
-if settle_threshold and settle_to and balance > tonumber(settle_threshold) then
+if settle_threshold and settle_to and balance > tonumber(settle_threshold) and balance > tonumber(settle_to) then
```
sappenin: I don't know Lua terribly well, but from my reading of the [Lua docs](https://www.lua.org/pil/3.3.html), `and` is just simple logical operator.  If that understanding is correct, then wouldn't `if settle_threshold and settle_to and balance > tonumber(settle_threshold)` reduce to just `if true > tonumber(settle_threshold)`, which would always result in `false`?

sappenin: I'm also not 100% sure either what the various balance numbers should be that should trigger settlement. Best I've come to thus far is this:

* `settle_threshold` must be non-nil (if it's nil, then settlement was perhaps disabled on the account).
* `clearing_balance` must be greater than `settle_threshold` (this is the core of the "should I settle logic")
* `settle_threshold` must be greater than `settle_to` (e.g., settleTo=5, settleThreshold=6)

That would yield this:

```
if settle_threshold and (clearing_balance > settle_threshold) and (settle_threshold > settle_to) then
```

crates/interledger-store-redis/src/store.rs
```diff
@@ -86,13 +86,14 @@ local prepaid_amount, settle_threshold, settle_to = unpack(redis.call('HMGET', t
...
     settle_amount = balance - tonumber(settle_to)
 
     -- Update the balance _before_ sending the settlement so that we don't accidentally send
```
sappenin: I suggest we should improve this comment. Something like what I [suggested](https://github.com/emschwartz/interledger-rs/issues/160#issuecomment-515132885) in #160, although what I wrote is somewhat long, so I'll defer to you @gakonst and @emschwartz

crates/interledger-store-redis/src/store.rs
emschwartz: Seems like it would be clearer and more certain to include parentheses

crates/interledger-store-redis/src/store.rs
gakonst: What do you think about: https://github.com/emschwartz/interledger-rs/pull/164/commits/d9791293c1848bd9f88d25d2723ff490c49d02e5?

sappenin: LGTM!

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -782,6 +794,67 @@ fn prefixed_mesage(challenge: Vec<u8>) -> Vec<u8> {
     ret
 }
 
+#[doc(hidden)]
+#[allow(clippy::all)]
+pub fn run_ethereum_engine<R, Si>(
```
emschwartz: What do you think about making this a builder too?

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -2,33 +2,9 @@ use clap::{value_t, App, Arg, SubCommand};
...
-pub fn generate_redis_secret(secret_seed: &[u8; 32]) -> [u8; 32] {
-    let mut redis_secret: [u8; 32] = [0; 32];
-    let sig = hmac::sign(
-        &hmac::SigningKey::new(&digest::SHA256, secret_seed),
-        REDIS_SECRET_GENERATION_STRING.as_bytes(),
-    );
-    redis_secret.copy_from_slice(sig.as_ref());
-    redis_secret
-}
+use interledger_settlement_engines::engines::ethereum_ledger::{run_ethereum_engine, EthAddress};
```
emschwartz: Might be worth splitting this up even more so that we just have each engine exposing its own binary.

In the future, we may also want to move them to separate repos under the same org, but we don't need to do that now.

crates/interledger-store-redis/src/store.rs
```diff
@@ -83,15 +84,19 @@ local to_amount = tonumber(ARGV[2])
 local balance = redis.call('HINCRBY', to_account, 'balance', to_amount)
 local prepaid_amount, settle_threshold, settle_to = unpack(redis.call('HMGET', to_account, 'prepaid_amount', 'settle_threshold', 'settle_to'))
 
--- Check if we should send a settlement for this account
+-- The logic for trigerring settlement is as follows:
+--  1. settle_threshold must be non-nil (if it's nil, then settlement was perhaps disabled on the account).
+--  2. clearing_balance must be greater than settle_threshold (this is the core of the 'should I settle logic')
```
emschwartz: The variable is called `balance` here

```suggestion
--  2. balance must be greater than settle_threshold (this is the core of the 'should I settle logic')
```

crates/interledger-store-redis/src/store.rs
```diff
@@ -83,15 +84,19 @@ local to_amount = tonumber(ARGV[2])
...
+--  3. settle_threshold must be greater than settle_to (e.g., settleTo=5, settleThreshold=6)
 local settle_amount = 0
-if settle_threshold and settle_to and balance > tonumber(settle_threshold) then
+if (settle_threshold and settle_to) and (balance > tonumber(settle_threshold)) and (tonumber(settle_threshold) > tonumber(settle_to)) then
```
emschwartz: Is 0 falsy in Lua? If so, this wouldn't work with `settle_to` set to 0

crates/interledger-store-redis/src/store.rs
```diff
@@ -544,8 +543,8 @@ impl BalanceStore for RedisStore {
                     .query_async(self.connection.as_ref().clone())
                     .map_err(move |err| {
                         error!(
-                            "Error handling fulfill from account: {} to account: {}: {:?}",
-                            from_account_id, to_account_id, err
+                            "Error handling fulfill to account: {}: {:?}",
```
emschwartz: Technically we got the fulfillment _from_ the `to` account

crates/interledger-store-redis/tests/balances_test.rs
```diff
@@ -61,20 +62,15 @@ fn prepare_then_fulfill() {
...
+                            .update_balances_for_fulfill(account1.clone(), 100)
                             .and_then(move |_| {
                                 store_clone_2
                                     .clone()
                                     .get_balance(account0.clone())
                                     .join(store_clone_2.clone().get_balance(account1.clone()))
                                     .and_then(move |(balance0, balance1)| {
                                         assert_eq!(balance0, -100);
-                                        assert_eq!(balance1, 500);
+                                        assert_eq!(balance1, -1000); // the account must be settled down to -1000
```
emschwartz: Why is this -1000? This doesn't really make sense to me

Do these accounts even have settlement details configured?

crates/interledger-store-redis/tests/common/store_helpers.rs
```diff
@@ -26,6 +26,23 @@ pub fn test_store() -> impl Future<Item = (RedisStore, TestContext), Error = ()>
         })
 }
 
+// same as test_store, but uses an account without `settle_to` for netting
+// tests.
+#[allow(dead_code)]
+pub fn test_other_store() -> impl Future<Item = (RedisStore, TestContext), Error = ()> {
```
emschwartz: Instead of making another test store, I think we should just add all of the accounts to the same store and use specific pairs of accounts for certain tests

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -14,14 +14,10 @@ pub trait BalanceStore: AccountStore {
         &self,
         from_account: Self::Account,
         incoming_amount: u64,
-        to_account: Self::Account,
-        outgoing_amount: u64,
```
emschwartz: FWIW, the reason I originally had these other values here was to allow for better ways of accounting for the balance. This store isn't doing proper accounting right now. If the server crashes after `update_balances_for_prepare` but before the fulfillment is received, the money from the `from` account will effectively disappear. A better way to do it would be to put that money (atomically) into another account, from which it would be transferred when the fulfillment comes in. One thing I'm not sure about is whether that pending balance would be linked to the `from` account, to the `from` _and_ `to` accounts (meaning there would be a separate row for pending balance between any pairs of accounts, which doesn't seem great), or neither.

crates/interledger-store-redis/tests/balances_test.rs
```diff
@@ -164,7 +150,7 @@ fn enforces_minimum_balance() {
 
 #[test]
 fn netting_fulfilled_balances() {
```
emschwartz:
```suggestion
// Prepare and Fulfill a packet for 100 units from Account 0 to Account 1
// Then, Prepare and Fulfill a packet for 80 units from Account 1 to Account 0
fn netting_fulfilled_balances() {
```

crates/interledger/tests/eth_ledger_settlement.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+use interledger_settlement_engines::engines::ethereum_ledger::run_ethereum_engine;
+
+#[derive(serde::Deserialize)]
+struct DeliveryData {
+    delivered_amount: u64,
+}
+
+#[test]
+fn eth_ledger_settlement() {
```
emschwartz: It's fine to put the integration tests involving settlement in the `interledger` crate for now but I think we should come up with a better place to house them longer-term. It seems like we might either want to put the integration tests that involve settlement into the `settlement-engines` crate or create another crate just for integration tests

crates/interledger/tests/eth_ledger_settlement.rs
```diff
@@ -0,0 +1,325 @@
...
+                            })
+                    };
+
+                    let create1 = create_account(node1_engine, 1);
+                    let create2 = create_account(node2_engine, 1);
+
+                    let send1 = send_money(node1_http, node2_http, 10);
+                    let send2 = send_money(node1_http, node2_http, 20);
+                    let send3 = send_money(node1_http, node2_http, 40);
+                    let send4 = send_money(node1_http, node2_http, 1);
```
emschwartz: What's the point of sending multiple payments?

crates/interledger/tests/eth_ledger_settlement.rs
```diff
@@ -0,0 +1,325 @@
...
+                                    .send()
+                                    .map_err(|err| {
+                                        eprintln!("Error getting account data: {:?}", err);
+                                        err
+                                    })
+                                    .and_then(|res| res.error_for_status())
+                                    .and_then(|res| res.into_body().concat2())
+                            };
+
+                            get_balance(1, node1_http, "bob").and_then(move |ret| {
```
emschwartz: Could you include a comment about why we'd expect both of those balances to be what they are?

crates/interledger/tests/eth_ledger_settlement.rs
emschwartz: Could you put a comment explaining what this test is doing at a high level?

crates/interledger-settlement-engines/src/main.rs
gakonst: Let's track this separately? I think it's useful for a node operator who runs multiple engines to just have to download one binary. We should just enable conditional compilation for each engine imo.

crates/interledger-store-redis/src/store.rs
gakonst: https://stackoverflow.com/questions/19318901/lua-true-equals-false/19319162#19319162

> All Lua values when used as Booleans evaluate to true, except nil and false. This does not mean that values that evaluate to true are equal to true. If you want to convert a value v to Boolean, use not not v.

Not based on the above, but I will expand the test to capture all cases of the if clause.

crates/interledger-store-redis/tests/balances_test.rs
gakonst: Yes, the settle_to is set to -1000 in the fixtures for that account.

crates/interledger/tests/eth_ledger_settlement.rs
gakonst: None really, just felt like making a more elaborate scenario than making a direct 71 payment.

crates/interledger/tests/eth_ledger_settlement.rs
gakonst: Yeah we could move this to the engines crate, but it would require importing the `Interledger` crate as a dev dependency to use `InterledgerNode` for the tests. I guess that's OK.

crates/interledger-store-redis/src/store.rs
gakonst: Do you have a suggestion for renaming this to a better phrased error, in absence of the `to` account id argument?

crates/interledger-settlement-engines/src/main.rs
gakonst: Tracked in #177

crates/interledger-service-util/src/balance_service.rs
gakonst: Tracked in #176

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
gakonst: Probably better to remove this function and improve the existing engine builder.

crates/interledger-store-redis/src/store.rs
emschwartz:
```suggestion
                            "Error handling Fulfill received from account: {}: {:?}",
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -14,24 +16,18 @@ pub trait BalanceStore: AccountStore {
...
     ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
 
     fn update_balances_for_fulfill(
         &self,
-        from_account: Self::Account,
-        incoming_amount: u64,
         to_account: Self::Account,
         outgoing_amount: u64,
-    ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
+    ) -> Box<dyn Future<Item = (i64, u64), Error = ()> + Send>;
```
emschwartz: Why is the balance being returned if it's not used?

Also, it would be good to include a comment above this function to describe what the returned value represents

crates/interledger-store-redis/tests/balances_test.rs
```diff
@@ -46,7 +46,8 @@ fn prepare_then_fulfill() {
                 let account0 = accounts[0].clone();
                 let account1 = accounts[1].clone();
                 store
-                    .update_balances_for_prepare(accounts[0].clone(), 100, accounts[1].clone(), 500)
+                    // nothing happens with the outgoing amount for prepare
```
emschwartz:
```suggestion
```

there is no outgoing amount here anymore

crates/interledger-store-redis/tests/common/fixtures.rs
```diff
@@ -70,4 +70,13 @@ lazy_static! {
         settlement_engine_url: None,
         settlement_engine_asset_scale: None,
     };
+    pub static ref ACCOUNT_DETAILS_3: AccountDetails = {
```
emschwartz: nit: I think that since we don't actually use Charlie's settlement details, we could just remove them from `ACCOUNT_DETAILS_2` and use that account instead of creating a third one

crates/interledger-service-util/src/balance_service.rs
gakonst: Added a comment in the trait definition

crates/interledger/tests/eth_ledger_settlement.rs
gakonst: Done in https://github.com/emschwartz/interledger-rs/pull/164/commits/cbd40f9e36a7696ced4a9e788b00ae6ae45aa5dc

crates/interledger-store-redis/src/store.rs
gakonst: more tests added here https://github.com/emschwartz/interledger-rs/pull/164/commits/ed3abc5a51ce95103301a91e0420094b464b2d1e, it indeed was not falsy

crates/interledger-store-redis/tests/balances_test.rs
```diff
@@ -61,20 +62,60 @@ fn prepare_then_fulfill() {
...
+            .clone()
+            .get_accounts(vec![2, 3, 4])
+            .map_err(|_err| panic!("Unable to get accounts"))
+            .and_then(move |accounts| {
+                // We test that the following cases get credited properly:
+                // 1. settle_to = None: never settles
+                let no_settle_to_acc = accounts[0].clone();
+                // 2. settle_to > settle_threshold never settles
+                let misconfigured_acc = accounts[1].clone();
+                // 3. settle_to <= settle_threshold settles when balance exceeds settle_threshold
```
emschwartz: It seems like these should be separate tests so that if one of them fails, it's clearer what the issue was

crates/interledger-store-redis/tests/common/fixtures.rs
```diff
@@ -70,4 +70,21 @@ lazy_static! {
         settlement_engine_url: None,
         settlement_engine_asset_scale: None,
     };
+    // account misconfigured with settle_to >= settle_threshold does not get settlements
+    pub static ref ACCOUNT_DETAILS_3: AccountDetails = {
```
emschwartz: another nit: if these accounts are only needed in a specific test, the details should just be in that test and they should be added to the store at the beginning of that test. We don't want to add accounts with all permutations of options to the store just for individual tests that look at those edge cases

crates/interledger-store-redis/tests/balances_test.rs
gakonst: Check https://github.com/emschwartz/interledger-rs/pull/164/commits/211cb7ffb3cc164a3c6d85b38277186f84630f0d

crates/interledger-store-redis/tests/common/fixtures.rs
gakonst: done in 211cb7f

PR #167, opened 2019-07-26
docs(simple/README.md): Add an overview image.

dora-gt:
Added a diagram which shows a simple overview.


dora-gt:
Should we add something else?

emschwartz:
Looks great!

PR #170, opened 2019-07-26
Remove dependence on hashbrown

### LINKED ISSUE #130: Switch to std HashMap
Since Rust 1.36 std::collections::HashMap uses hashbrown. 

> In Rust 1.36.0, the HashMap<K, V> implementation has been replaced with the one in the hashbrown crate which is based on the SwissTable design. While the interface is the same, the HashMap<K, V> implementation is now faster on average and has lower memory overhead. Note that unlike the hashbrown crate, the implementation in std still defaults to the SipHash 1-3 hashing algorithm.

However, when implemented in https://github.com/emschwartz/interledger-rs/commit/44fe296edb19aec4c65e1b4c4af9aefb2ecdb8dd, the tests became flaky: https://circleci.com/gh/emschwartz/interledger-rs/623

Close this issue if you replace hashbrown with std::collections without making the test flaky. Maybe has to do with the different hashing algorithm.

bstrie:
Note that this involves changing a HashMap to a Vec in interledger-ccp, because we were implicitly relying on things being iterated upon in insertion order which is not something that HashMaps usually guarantee, leading to flaky tests upon switching implementations.

I would like someone who is more well-versed in the interledger-ccp logic to take a look at the changes there, since the switch from HashMap to Vec seemed a bit *too* easy. My worry is that there might be some implicit invariants that I'm not entirely sure are still upheld (though if not then I do at least think that the prior implementation may not have been intentionally upholding them either, so it's probably no *more* incorrect than it used to be). See my PR-level comments below for particular cases.

Fixes #130

### DIFF
--- a/crates/interledger-ccp/src/server.rs
+++ b/crates/interledger-ccp/src/server.rs
@@ -1,11 +1,21 @@
-use crate::{packet::*, routing_table::RoutingTable, CcpRoutingAccount, RouteManagerStore};
+#[cfg(test)]
+use crate::packet::PEER_PROTOCOL_CONDITION;
+use crate::{
+    packet::{
+        Mode, Route, RouteControlRequest, RouteUpdateRequest, CCP_CONTROL_DESTINATION,
+        CCP_RESPONSE, CCP_UPDATE_DESTINATION,
+    },
+    routing_table::RoutingTable,
+    CcpRoutingAccount, RouteManagerStore,
+};
 use bytes::Bytes;
 use futures::{
     future::{err, join_all, ok, Either},
     Future, Stream,
 };
-use hashbrown::HashMap;
-use interledger_packet::*;
+#[cfg(test)]
+use interledger_packet::PrepareBuilder;
+use interledger_packet::{Address, ErrorCode, Fulfill, Reject, RejectBuilder};
 use interledger_service::{
     Account, BoxedIlpFuture, IncomingRequest, IncomingService, OutgoingRequest, OutgoingService,
 };
@@ -14,6 +24,7 @@ use lazy_static::lazy_static;
 use log::{debug, error, trace, warn};
 use parking_lot::{Mutex, RwLock};
 use ring::digest::{digest, SHA256};
+use std::collections::HashMap;
 use std::{
     cmp::min,
     convert::TryFrom,
@@ -34,7 +45,7 @@ fn hash(preimage: &[u8; 32]) -> [u8; 32] {
     out
 }
 
-type NewAndWithDrawnRoutes = (Vec<Route>, Vec<Bytes>);
+type NewAndWithdrawnRoutes = (Vec<Route>, Vec<Bytes>);
 
 pub struct CcpRouteManagerBuilder<I, O, S> {
     /// The next request handler that will be used both to pass on requests that are not CCP messages.
@@ -102,7 +113,7 @@ where
             store: self.store.clone(),
             spawn_tasks: self.spawn_tasks,
             forwarding_table: Arc::new(RwLock::new(RoutingTable::default())),
-            forwarding_table_updates: Arc::new(RwLock::new(HashMap::new())),
+            forwarding_table_updates: Arc::new(RwLock::new(Vec::new())),
             last_epoch_updates_sent_for: Arc::new(Mutex::new(0)),
             local_table: Arc::new(RwLock::new(RoutingTable::default())),
             incoming_tables: Arc::new(RwLock::new(HashMap::new())),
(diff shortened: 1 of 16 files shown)


crates/interledger-ccp/src/server.rs
```diff
@@ -137,7 +148,7 @@ pub struct CcpRouteManager<I, O, S, A: Account> {
     forwarding_table: Arc<RwLock<RoutingTable<A>>>,
     last_epoch_updates_sent_for: Arc<Mutex<u32>>,
     /// These updates are stored such that index 0 is the transition from epoch 0 to epoch 1
-    forwarding_table_updates: Arc<RwLock<HashMap<u32, NewAndWithDrawnRoutes>>>,
+    forwarding_table_updates: Arc<RwLock<Vec<NewAndWithDrawnRoutes>>>,
```
bstrie: Concern: is the u32 that was formerly stored here really reducible to a vec index?

crates/interledger-ccp/src/server.rs
```diff
@@ -535,8 +546,8 @@ where
                         forwarding_table.delete_route(prefix);
                     }
 
-                    let epoch = forwarding_table.increment_epoch();
-                    forwarding_table_updates.insert(epoch, (new_routes, withdrawn_routes));
+                    forwarding_table.increment_epoch();
```
bstrie: Concern: the value returned by this call is no longer used anywhere; instead we *assume* that the index of the next vec element created by the following push call is identical to this value. Should I put an assertion here to make sure?

crates/interledger-ccp/src/server.rs
```diff
@@ -619,26 +630,18 @@ where
...
-        let mut new_routes: Vec<Route> = Vec::with_capacity(epochs_to_take);
+        let mut new_routes: Vec<Route> = Vec::with_capacity(end.saturating_sub(start));
         let mut withdrawn_routes: Vec<Bytes> = Vec::new();
         // Iterate through each of the given epochs
-        for (new, withdrawn) in forwarding_table_updates
-            .values()
-            .skip(from_epoch_index as usize)
-            .take(epochs_to_take)
-        {
+        for (new, withdrawn) in &forwarding_table_updates[start..end] {
```
bstrie: Concern: the way this was implemented before, we implicitly assumed that the epochs contained within the map were numbered via a continuous unbroken range beginning at zero (or else otherwise this skip and take couldn't have been correct (or else to_epoch_index and from_epoch_index are poorly named)). This should be equivalent to the slicing operation this is replaced with, but if the underlying assumption is itself wrong then this might be totally nuts.

crates/interledger-ccp/src/server.rs
bstrie: (Put more plainly, and to echo @emschwartz 's sentiment from yesterday: "why did I even use a hashmap here in the first place?")

bstrie:
Build is failing because I forgot to cargo fmt, I'll hold off on fixing that until someone has had a chance to take a look at my comments because I don't want to blow them away just for formatting changes. I'll set up a commit hook in the meantime :P

sentientwaffle:
I think it should be safe to use a `Vec`. The forwarding routing table's epoch can only ever increment. I'd probably stick a `debug_assert` in there just to be explicit, since the code later relies on [iterating over a range of epochs](https://github.com/emschwartz/interledger-rs/blob/b6b7c566092bd5c25816156f2ac308dde6a3ea38/crates/interledger-ccp/src/server.rs#L643-L644).

My best guess as to why it is a map and not a list in the JS implementation is that since some of the other epochs can jump forward or reset (e.g. when the received table id changes) the author assumed that the same was true of the forwarding table.

As an aside, while I was digging through this I noticed that the code at https://github.com/interledgerjs/ilp-connector/blob/master/src/services/route-broadcaster.ts#L503-L505 doesn't seem to have an analogue in the rust impl. Which should be ok -- afaik it's just an optimization to prevent sending superseded routes.

crates/interledger-ccp/src/server.rs
emschwartz: @sentientwaffle pointed out that withdrawn is one word so the camel case should be `Withdrawn` instead of `WithDrawn`

crates/interledger-ccp/src/server.rs
emschwartz: I think this is fine, though it might be worth putting a comment in the code here about this

crates/interledger-ccp/src/server.rs
emschwartz: Seems like it

crates/interledger-ccp/src/server.rs
emschwartz: If the updates are just being pushed onto the `Vec`, then this should be fine. We might want to check that the `from_epoch_index` and `to_epoch_index` are in range or use `forwarding_table_updates.iter().take(to_epoch_index as usize).skip(from_epoch_index as usize)` to avoid panicking if they're out of range for some reason.

bstrie: Not a bad idea, I wasn't sure under what circumstances this function might get called and who might be determining the input parameters, so if we aren't sure about our callers then we can be a bit more defensive.

bstrie:
Comments addressed.

PR #183, opened 2019-07-31
Settlement Engine API interop updates

gakonst:
Modified the Engine APIs to match the remarks made in https://github.com/interledgerjs/settlement-xrp/pull/1

This should make settlement interoperable with Rafiki.


examples/README.md
```diff
@@ -3,7 +3,7 @@
 Here you can find various demos of Interledger.rs' functionality:
 
 1. [Simple Two-Node Payment](./simple/README.md)
-1. [Two-Node Payment with On-Ledger Settlement](./eth-settlement/README.md)
+1. [Two-Node Payment with On-Ledger Ethereum Settlement](./e2e_tests/ethereum_ledger/README.md)
```
emschwartz: What's the point of having an `e2e_tests` folder? All of the examples are end-to-end

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -741,7 +738,9 @@ where
         account_id: String,
         body: Quantity,
     ) -> Box<dyn Future<Item = ApiResponse, Error = ApiResponse> + Send> {
-        let amount = U256::from(body.amount);
+        // should we handle the error here? It comes from the connector client
+        // so this conversion should never fail
+        let amount = U256::from_dec_str(&body.amount).unwrap();
```
emschwartz: Always better to handle the error rather than unwrapping. It should just return a 400 if it's invalid

examples/README.md
gakonst: There was a e2e_tests directory there already

emschwartz: Maybe on your local machine? That directory didn't exist before in git or on my machine

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -480,7 +476,7 @@ where
             client
                 .post(url.clone())
                 .header("Idempotency-Key", tx_hash.to_string())
-                .json(&json!({ "amount": amount }))
+                .json(&json!({ "amount": amount.to_string(), "scale" : 18 }))
```
emschwartz: This still seems like it's going to cause problems

gakonst: Other than the precision problem which would be solved by switching all accounts to use Gwei, what other issue do you see?

gakonst: note that in https://github.com/emschwartz/interledger-rs/pull/187 we make the engine's scale possible to be set via the CLI, which allows us to configure it accordingly to avoid any conversion errors.

kincaidoneil:
@gakonst Does this operate correctly for ERC-20s with non-standard scales/scales that aren't 10^-18? (Most are 18, but as I recall a handful of the top 20 ERC-20s aren't)

gakonst:
@kincaidoneil We can now configure the scale via CLI, so if you're using a non-18 scale asset you can configure it accordingly.

Issue #180, opened 2019-07-30
Make examples more consistent


emschwartz:
The [Simple](https://github.com/emschwartz/interledger-rs/blob/master/examples/simple/README.md) and [Ethereum Settlement](https://github.com/emschwartz/interledger-rs/blob/master/examples/eth-settlement/README.md) examples have slightly different styles and we should try to make them more consistent.

Some of the notable differences:
- The Simple example puts all of the commands in shell scripts while the Ethereum Settlement example puts the code in the README. The advantage of shell scripts is that they can be run directly, while the disadvantage is that it's less nice to read through them than reading the markdown README file. Having both means we need to make sure the code and README stay in sync, which is a bit annoying.
- The Ethereum example gets the account balances directly from Redis while the other uses the HTTP API. We should definitely try to avoid calling the DB directly and instead encourage the use of the API, however it's slightly nicer to print the balance in line with the comment (as is done [here](https://github.com/emschwartz/interledger-rs/tree/master/examples/eth-settlement#6-make-some-payments)) than to print the JSON `{"balance":"100"}` on the line after the comment.
- The Ethereum example suggests opening 7 terminal windows to run the different services, while the scripts in the other example run the services as forked processes and write the logs to different files. Opening 7 terminal windows is pretty excessive so I'd lean towards only requiring one, though that does make it more difficult to watch different processes happen simultaneously.
- The examples made in the settlement README.md's are hard to read due to how the POST parameters are added. We should move these to a file / improve formatting
What do you think? cc @gakonst @bstrie @dora-gt @pgarg-ripple

emschwartz:
Another question is whether the examples should build on one another or stand on their own. For example, we could run them in a series so that in the first you set up two nodes and send payments with no settlement, then add settlement, then add multi-hop. Thoughts?

gakonst:
> Opening 7 terminal windows is pretty excessive so I'd lean towards only requiring one, though that does make it more difficult to watch different processes happen simultaneously.

From a user's visual perspective IMO it's always more useful to first see all terminals together, each doing their thing. Once you're accustomed to the process you can start running processes in the background and redirecting to logs (as I do in the example scripts).

> Another question is whether the examples should build on one another or stand on their own. For example, we could run them in a series so that in the first you set up two nodes and send payments with no settlement, then add settlement, then add multi-hop. Thoughts?

Commented on this here: https://github.com/emschwartz/interledger-rs/pull/184#discussion_r309515966

[...]

PR #190, opened 2019-08-03
Consistency Improvement of Examples

dora-gt:
Fixed examples (simple, eth-settlement) to be consistent.

- eth example uses only 1 redis node (using DB 0 and 1 to be consistent with `simple`)
- eth example outputs its logs into log files
- eth example does NOT stop nodes etc. when run by `run-md.sh` in order to confirm settlements
- both `simple` and `eth-settlement` are runnable with `run-md.sh`
- both `simple` and `eth-settlement` have overview diagrams

Fixes: #180 #201 #202


emschwartz:
> JSONs are not shown in the markdown. Readers might not understand the settings.

What about using something like (example taken from [stack overflow](https://stackoverflow.com/a/7173011)):
```
curl --header "Content-Type: application/json" \
  --request POST \
  --data '{"username":"xyz","password":"xyz"}' \
  http://localhost:3000/api/login
```
where the JSON data is written out as a string instead of being loaded from a file?

examples/README.md
```diff
@@ -9,3 +9,24 @@ Here you can find various demos of Interledger.rs' functionality:
 1. Integrating Interledger Into Your App (Coming Soon!)
 
 Have questions? Feel free to [open an issue](https://github.com/emschwartz/interledger-rs/issues/new) or ask a question [on the forum](https://forum.interledger.org/)!
+
+## Running Scripts Collectively
+Although instruction scripts are shown on the example pages, you can also run the scripts collectively and automatically using [lit](https://github.com/vijithassar/lit). To run scripts, install `lit`:
+
+```bash
+# Under the directory you want to install lit.sh to
+$ curl https://raw.githubusercontent.com/vijithassar/lit/master/lit.sh > lit.sh && chmod +x lit.sh
```
emschwartz: Unless you see a benefit of using `lit`, I would have a slight preference for the solution from the [`mdlp` gist](https://gist.githubusercontent.com/trauber/4955706/raw/0d9e6261bf78da5e448bea383a8ad9d51548843c/mdlp.awk) instead because it's _so_ simple. I think `lit` is a bit overkill because it has functionality for reading in files and writing them to specific places, as opposed to just operating on piped data.

Instead of telling the user to "install" something, it would be more like "just run this simple command, which we've copied into a file called run-md.sh in the repo for your convenience".

examples/simple/README.md
```diff
@@ -1,27 +1,175 @@
...
+        - Ubuntu: run `sudo apt-get install redis-server`
+        - macOS: If you use Homebrew, run `brew install redis`
+
+## Instructions
+
+### 1. Build interledger.rs
+First of all, let's build interledger.rs. This may take a little while.
+
+```bash
+cargo build --bins
```
emschwartz: It might be worth figuring out a way to have the md to sh script run commands that are commented out in the markdown file so that we can include lines like `echo "Building interledger.rs... (this may take up to a few minutes)"` that don't show up in the rendered README but make the runnable script a bit nicer

examples/simple/README.md
```diff
@@ -1,27 +1,175 @@
...
+
+```bash
+# Insert accounts on Node A
+# One account represents Alice and the other represents Node B's account with Node A
+
+# Create Alice's account
+curl \
+    -H "Content-Type: application/json" \
+    -H "Authorization: Bearer admin-a" \
+    -d @./node-a/alice.json \
```
emschwartz: I think we should just include the data here as a string instead of loading it from the file

examples/simple/README.md
dora-gt: It seems nearly impossible because fenced sections are considered as code blocks but I added a function that normal lines start with `<!--!` and end with `-->` are considered as scripts. `!` has no meaning, just for differentiating from normal comments. Multiple lines are not supported so far.

dora-gt: Also markdown files apparently support another comment format like:
```
[xxx]: # (yyy)
```
but I thought HTML comment format was easier to remember.

examples/simple/README.md
dora-gt: It becomes more similar to post parameters but hmmm, a bit nicer maybe?

dora-gt: Then should node config `.yml`s be broken down into separate config argument like:
```
$ cargo run --package interledger -- node --config ./node-a/config.yml &> logs/node-a.log &
```
to
```
cargo run --package interledger -- node \n
--ilp_address=example.node-a \n
--secret_seed=xxxxxx \n
&> logs/node-a.log &
```
?

I think thi is common usage and not so bad. Also we can leave `--config` option as is.

examples/simple/README.md
emschwartz: It seems like a normal HTML comment wrapping a code block section would work too:
```
<!--
\```bash
echo "hello world"
\```
-->
```

(the backslash is in front of the backticks so that it doesn't match the inner ones with the outer ones and render it incorrectly)

examples/simple/README.md
emschwartz: Fair point. We could do either one. The node currently supports having the options passed as environment variables so you could do:
```
ILP_ADDRESS=example.node-a \
ILP_SECRET_SEED=xxxx \
...
```

examples/simple/README.md
dora-gt: I just modified the script to be compatible with multiple lines like:

```
<!--!
echo script goes here
-->
```

This seems a bit simpler so I think this one is better if you don't mind.

The latest commit realizes almost the same as you originally wrote in `run-all.sh` including stopping services and checking balances twice etc.

examples/simple/README.md
dora-gt: fixed.

emschwartz: Actually I put in [this alias](https://github.com/emschwartz/interledger-rs/blob/7e2ff6304e5fbf240e40a4bcb88225ebe10201da/crates/interledger/src/node.rs#L85) so that it would accept `ILP_ADDRESS` instead of `ILP_ILP_ADDRESS` :wink:

examples/eth-settlement/README.md
```diff
@@ -1,12 +1,40 @@
 # Guide to E2E Testing Interledger settlement with the Ethereum Ledger
 
-First, you need an Ethereum network. We'll use
-[ganache-cli](https://github.com/trufflesuite/ganache-cli) which deploys a local
-Ethereum testnet at `localhost:8545`. You're free to use any other
-testnet/mainnet you want. To install `ganache-cli`, run 
-`npm install -g ganache-cli`. You also need to have `redis-server` and
-`redis-cli` available in your PATH. In Ubuntu, you can obtain these by running `sudo apt-get install redis-server`
+## Overview
+This example shows how to configure Interledger.rs nodes and use an Ethereum network including testnet and mainnet as a settlement ledger underneath the nodes. If you are new to Ethereum, learn [here](https://www.ethereum.org/beginners/). To learn what are settlements, refer to [Peering, Clearing and Settling](https://github.com/interledger/rfcs/blob/master/0032-peering-clearing-settlement/0032-peering-clearing-settlement.md).
```
emschwartz:
```suggestion
This example shows how to configure Interledger.rs nodes and use an Ethereum network including testnet and mainnet as a settlement ledger underneath the nodes. If you are new to Ethereum, learn [here](https://www.ethereum.org/beginners/). To learn about settlement in Interledger, refer to [Peering, Clearing and Settling](https://github.com/interledger/rfcs/blob/master/0032-peering-clearing-settlement/0032-peering-clearing-settlement.md).
```

examples/eth-settlement/README.md
```diff
@@ -1,12 +1,40 @@
...
 
+(diagram here)
+
+## Prerequisites
+These are prerequisites and we provide a quick guide each.
```
emschwartz:
```suggestion
```

examples/eth-settlement/README.md
```diff
@@ -1,12 +1,40 @@
...
+
+- An Ethereum network
```
emschwartz:
```suggestion
- [An Ethereum network](#an-ethereum-network) to connect to
```

examples/eth-settlement/README.md
```diff
@@ -1,12 +1,40 @@
...
+- Redis
```
emschwartz:
```suggestion
- [Redis](#redis)
```

examples/eth-settlement/README.md
```diff
@@ -1,12 +1,40 @@
...
+
+### An Ethereum Network
+First, you need an Ethereum network. You could use your own local testnet or also the mainnet.
+
+We'll use [ganache-cli](https://github.com/trufflesuite/ganache-cli) which deploys a local Ethereum testnet at `localhost:8545`. To use `ganache-cli`, we need `node` installed. There are a few ways to install it. If you work on multiple JavaScript or TypeScript projects which require different `node` versions, using `nodebrew` may be suitable.
```
emschwartz: I think `nvm` is more common than `nodebrew`

examples/eth-settlement/README.md
```diff
@@ -1,12 +1,40 @@
...
+
+- [`nodebrew`](https://github.com/hokaccha/nodebrew) (node version manager)
+    - macOS: If you use [Homebrew](https://brew.sh/), run `brew install nodebrew`, `nodebrew install-binary stable` and `nodebrew use stable` then [edit shell config](https://github.com/hokaccha/nodebrew#install) and reload it.
+    - others: Refer to [nodebrew](https://github.com/hokaccha/nodebrew) site.
+- Install independently
+    - macOS: If you use [Homebrew](https://brew.sh/), run `brew install node`
+    - Ubuntu: `sudo apt-get install nodejs npm`
+
+Then you should be able to use `npm`. To install `ganache-cli`, run `npm install -g ganache-cli`.
```
emschwartz: This gets a bit lost at the bottom so it might be worth including before the instructions for installing node. You could lead into the instructions for installing node by saying "if you do not already have Node.js installed on your computer, you can follow these instructions to install it", or something like that

examples/eth-settlement/README.md
```diff
@@ -1,12 +1,40 @@
...
+
+### Redis
+To store data of connectors, we need Redis installed.
```
emschwartz: The Interledger.rs nodes currently use Redis to store their data (SQL database support coming soon!)

examples/eth-settlement/README.md
```diff
@@ -1,12 +1,40 @@
...
+
+- Compile and install from the source code
+    - [Download the source code here](https://redis.io/download)
+- Install using package managers
+    - Ubuntu: run `sudo apt-get install redis-server`
+    - macOS: If you use Homebrew, run `brew install redis`
+
+## Instructions
 We will need **7** terminal windows in total to follow this tutorial in depth. You can run the
 provided `settlement_test.sh` script instead to see how the full process works.
```
emschwartz: This should be updated for the new way of running the whole example

examples/eth-settlement/README.md
```diff
@@ -23,21 +51,29 @@ provided `settlement_test.sh` script instead to see how the full process works.
 
 For ease, you may want to set these environment variables to save you some
 typing:
+
 ```
```
emschwartz:
```suggestion
```bash
```

examples/eth-settlement/README.md
```diff
@@ -23,21 +51,29 @@ provided `settlement_test.sh` script instead to see how the full process works.
...
+mkdir -p logs
 
-## 2. Configure Alice
+ganache-cli -m "abstract vacuum mammal awkward pudding scene penalty purchase dinner depart evoke puzzle" -i 1 &> logs/ganache.log &
+```
+
+To watch what's happening on ganache, or generally when you want to watch logs, use the `tail` command.  
+You can use the command like: `tail -f logs/ganache.log`
+
+### 2. Configure Alice
```
emschwartz: It might be clearer to distinguish "Alice" from "Alice's Node"

examples/eth-settlement/README.md
```diff
@@ -166,9 +202,9 @@ curl http://localhost:8770/accounts -X POST \
 
 Now we have both Alice and Bob up and running.
 
-## 4. Peer each other.
+### 4. Peer each other.
```
emschwartz:
```suggestion
### 4. Peer the nodes
```

examples/eth-settlement/README.md
```diff
@@ -166,9 +202,9 @@ curl http://localhost:8770/accounts -X POST \
...
 
-### Insert Bob's account to Alice's connector
+#### Insert Bob's account to Alice's connector
 ```bash
 curl http://localhost:7770/accounts -X POST \
     -d "ilp_address=example.bob&asset_code=ETH&asset_scale=18&max_packet_amount=10&settlement_engine_url=http://127.0.0.1:3000&settlement_engine_asset_scale=18&http_endpoint=http://127.0.0.1:8770/ilp&http_incoming_token=bob&http_outgoing_token=alice&settle_threshold=70&min_balance=-100&settle_to=10" \
```
emschwartz: I think this would be clearer as JSON instead of as form-encoded data

examples/eth-settlement/README.md
```diff
@@ -182,20 +218,20 @@ Settlement Engine.
...
-## 5. Verify that addresses have been exchanged
+### 5. Verify that addresses have been exchanged
 
 Bob's address in Alice's store: `redis-cli -p 6379 hgetall "settlement:ledger:eth:1"`
 Alice's address in Bob's store: `redis-cli -p 6380 hgetall "settlement:ledger:eth:1"`
 
-## 6. Make some payments!
+### 6. Make some payments!
 
 A `pay_dump.sh` script is provided which you can use to make [SPSP
```
emschwartz: I think we should get rid of this script and put the commands you need to run in this instruction document

examples/eth-settlement/README.md
```diff
@@ -236,3 +272,28 @@ increased as a result of the settlement executions.
...
+gyp ERR! find Python checking if "python" can be used
+gyp ERR! find Python - executable path is "/Users/dora_gt/anaconda3/bin/python"
+gyp ERR! find Python - version is "3.6.2"
+gyp ERR! find Python - version is 3.6.2 - should be >=2.6.0 <3.0.0
+gyp ERR! find Python - THIS VERSION OF PYTHON IS NOT SUPPORTED
+gyp ERR! find Python checking if "python2" can be used
+gyp ERR! find Python - "python2" is not in PATH or produced an error
+```
+
+have to install python 2.7
```
emschwartz: Why is python needed for this at all?

examples/eth-settlement/README.md
```diff
@@ -1,12 +1,40 @@
...
 
+(diagram here)
+
+## Prerequisites
+These are prerequisites and we provide a quick guide each.
+
+- An Ethereum network
+- Redis
+
+### An Ethereum Network
```
emschwartz: We should be consistent about the capitalization of the section headers
```suggestion
### An Ethereum network
```

examples/eth-settlement/README.md
```diff
@@ -1,12 +1,40 @@
 # Guide to E2E Testing Interledger settlement with the Ethereum Ledger
```
emschwartz:
```suggestion
# Interledger with Ethereum On-Ledger Settlement
> A demo that sends payments between 2 Interledger.rs nodes and settles using Ethereum transactions.
```

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
-# Simple Two-Node Payment
+# Guide to Simple Two-Node Payment
```
```suggestion
# Simple Two-Node Interledger Payment
> A demo of sending a payment between 2 Interledger.rs nodes without settlement.
```

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
...
+
+## Overview
 
 This example sets up two local Interledger.rs nodes, peers them together, and sends a payment from one to the other. This example does not involve any remote nodes or networks. 
 
-To run the full example, you can use the [`run-all.sh`](./run-all.sh) script. Otherwise, you can walk through each step below.
+To run the full example, you can use `run-md.sh` as described [here](../README.md). Otherwise, you can walk through each step below.
```
emschwartz: Link to the script

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
...
 
 Each of the services write their logs to files found under the `logs` directory. You can run `tail -f logs/node-a.log`, for example, to watch the logs of Node A.
 
 ![overview](images/overview.svg)
 
 ## Prerequisites
 
-- [Redis](https://redis.io/download)
+- To store data of connectors, we need [Redis](https://redis.io/) installed.
```
```suggestion
- To store the connector data, we need [Redis](https://redis.io/) installed.
```

scripts/run-md.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# mdlp.awk from https://gist.github.com/trauber/4955706
```
emschwartz:
```suggestion
# Based on mdlp.awk from https://gist.github.com/trauber/4955706
```

scripts/run-md.sh
```diff
@@ -0,0 +1,15 @@
...
+# Originally written by @trauber Rich Traube
+
+# cc = code block count
+# hc = html comment count
+# ps = print script
+awk 'BEGIN { one_line_comment="^<!--!(.*)-->$" }
+{
+  if (hc % 2 == 0 && /^```/) { if (/^```(bash)$/) { ps = 1; } else { ps = 0; } cc++; next }
```
emschwartz: Could you include comments to explain what each of these lines do?

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
...
+
+if lsof -tPi :8770 ; then
+    kill `lsof -tPi :8770`
+fi
+
+printf "\n"
+-->
+
+### 1. Build interledger.rs
+First of all, let's build interledger.rs. This may take a little while.
```
```suggestion
First of all, let's build interledger.rs. (This may take a couple of minutes)
```

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
...
+```
+
+### 2. Launch Redis
+
+```bash
+# Create the logs directory if it doesn't already exist
+mkdir -p logs
+
+# Start Redis
+redis-server --version > /dev/null || printf "\e[31mUh oh! You need to install redis-server before running this example\e[m\n"
```
emschwartz: Might be worth putting this in the invisible section. Because this guide tells you to install Redis first, I would expect people to have installed it

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
...
+redis-server &> logs/redis.log &
+```
+
+To watch what's happening on Redis, or generally when you want to watch logs, use the `tail` command.
```
emschwartz: Actually the Redis log doesn't say much of interest

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
...
+To watch what's happening on Redis, or generally when you want to watch logs, use the `tail` command.  
+You can use the command like: `tail -f logs/redis.log`
+
+### 3. Launch 2 Nodes
+
+```bash
+# Turn on debug logging for all of the interledger.rs components and tower-web
+export RUST_LOG=tower-web,interledger/.*
+
+# Start both nodes
```
```suggestion
# Start both nodes
# Note that the configuration options can be passed as environment variables
# or saved to a YAML file and passed to the node with the `--config` or `-c` CLI argument
```

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
...
+
+<!--! printf "The Interledger.rs nodes are up and running!\n\n" -->
+
+Now the Interledger.rs nodes are up and running!  
+You can also watch the logs with: `tail -f logs/node-a.log` or `tail -f logs/node-b.log`.
+
+### 4. Configure the Nodes
+
+Let's create accounts on both nodes. The following script sets up accounts for two users, Alice and Bob. It also creates accounts that represent the connection between Nodes A and B.
+
```
```suggestion

See the [HTTP API docs](../../docs/api.md) for the full list of fields that can be set on an account.

```

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
...
+You need to install Redis to run `redis-server` command. See [Prerequisites](#prerequisites) section.
 
-## Running and Configuring the Nodes
+> curl: (7) Failed to connect to localhost port 7770: Connection refused
 
-The [`start-services.sh`](./start-services.sh) script runs a Redis instance and the two nodes, Node A and Node B.
+Your interledger.rs node is not running. The reason may be:
 
-The [`create-accounts.sh`](./create-accounts.sh) script sets up accounts for two users, Alice and Bob. It also creates accounts that represent the connection between Nodes A and B.
+1. You ran following commands too early. Adjust `sleep 1` command to `sleep 3` or larger.
```
```suggestion
1. You tried to insert the accounts before the nodes had time to spin up. Adjust the `sleep 1` command to `sleep 3` or larger.
```

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
...
+1. You have some other process running on port `7770`. Stop the process and retry.
 
-## Sending a Payment
+To stop the process running on port `7770`, try `` kill `lsof -i:7770 -t` ``. Since this example launches 2 nodes, the port may be other than `7770`. Adjust the port number according to the situation.
 
-The [`send-payment.sh`](./send-payment.sh) script sends a payment from Alice to Bob that is routed from Node A to Node B.
+## Conclusion
 
-You can run the [`check-balances.sh`](./check-balances.sh) script to print each of the accounts' balances (try doing this before and after sending a payment).
+That's it for this example! You've learned how to set up nodes and connect them, even how to send payments between the nodes.
```
```suggestion
That's it for this example! You've learned how to set up Interledger.rs nodes, connect them together, and how to send a payment from one to the other.
```

examples/simple/README.md
```diff
@@ -1,27 +1,297 @@
...
 
-That's it for this example! Check out the [other examples](../README.md) for more complex demos that show more of the features of Interledger, including multi-hop routing and cross-currency payments.
+This is an easy-to-start one so check out the [other examples](../README.md) for more complex demos that show more of the features of Interledger, including multi-hop routing and cross-currency payments.
```
```suggestion
Check out the [other examples](../README.md) for more complex demos that show other features of Interledger, including multi-hop routing and cross-currency payments.
```

examples/README.md
```diff
@@ -9,3 +9,16 @@ Here you can find various demos of Interledger.rs' functionality:
...
+
+## Running Scripts Collectively
+Although instruction scripts are shown on the example pages (.md), you can also run the scripts collectively and automatically using [run-md.sh](../scripts/run-md.sh). To run an example:
+
+```bash
+# Under the example directory, for example, "simple"
+$ cat README.md | ../scripts/run-md.sh
+```
+
+## For those who add examples
```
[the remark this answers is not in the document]
dora-gt: I'll move it later.

examples/eth-settlement/README.md
dora-gt: I'll fix this later, after merging your suggestions.

examples/eth-settlement/README.md
dora-gt: Is this a suggestion? (maybe so... I'll include this)

examples/simple/README.md
dora-gt: 😅 I'll fix.

scripts/run-md.sh
dora-gt: I'm sorry this is so complex. I'll comment later.
(It was hard to use state machine inside awk... so became so complex)

examples/simple/README.md
dora-gt: wow... I didn't notice.

dora-gt:
Thank you sooooooooooo much for editing!
I'll fix the remaining issues soon, I'm going to take a rest😂

examples/eth-settlement/README.md
emschwartz: Oops, yes!

emschwartz:
```suggestion
The Interledger.rs nodes currently use Redis to store their data (SQL database support coming soon!)
```

examples/eth-settlement/README.md
dora-gt: oops.. I'm sorry I don't use Node.js so often. I'll fix

examples/eth-settlement/README.md
dora-gt: Fixed each.

examples/eth-settlement/README.md
dora-gt: Removed.

examples/eth-settlement/README.md
dora-gt: I was also surprised... it seems that ganache or its component uses it. I couldn't install when I first tried to install `ganache-cli`.

examples/eth-settlement/README.md
dora-gt: Fixed though it differs a bit from `simple` example because of a problem of settlement engine spec. The accounts cannot be added in order but should be done at the same time (like in the background).

examples/eth-settlement/README.md
dora-gt: I fixed the structure entirely.

examples/eth-settlement/README.md
dora-gt: I fixed the structure entirely.

examples/simple/README.md
dora-gt: Fixed.

examples/simple/README.md
dora-gt: Fixed.

examples/eth-settlement/README.md
dora-gt: image

examples/eth-settlement/README.md
dora-gt: Fixed.

examples/eth-settlement/README.md
dora-gt: hmmm yes, this seems to require some more. I'll fix.

dora-gt: Fixed, the output will be like:
image

dora-gt:
Addressed issues. If you have any more, please tell me.

docs/CONTRIBUTING.md
```diff
@@ -48,6 +48,8 @@ Please follow the instruction below when making pull requests.
 - Make pull requests against `master` branch of this repository (emschwartz/interledger-rs) from your repository.
 - If reviewers request some changes, please follow the instruction or make discussions if you have any constructive opinions on the PRs you made.
     - Then if you want to make some changes on your PRs, `push -f` is allowed to renew your branch after squashing your new commits. You don't need to open new PRs.
+- For our [examples](../examples/README.md), we adopted kind of `literate programming` approach. If you want to add some examples, it is highly recommended that your markdown files are properly described to be run with [`run-md.sh`](../scripts/run-md.sh) using a pipe.
```
emschwartz:
```suggestion
- For our [examples](../examples/README.md), we adopted a [literate programming](https://en.wikipedia.org/wiki/Literate_programming) approach. The examples are described in Markdown with shell commands included. The [`run-md.sh`](../scripts/run-md.sh) script parses the commands out of the Markdown file and runs them. If you want to add examples, please make sure your instruction file can be parsed and run by that script.
```

emschwartz:
```suggestion
- For our [examples](../examples/README.md), we adopted a [literate programming](https://en.wikipedia.org/wiki/Literate_programming) approach. The examples are described in Markdown with shell commands included. The [`run-md.sh`](../scripts/run-md.sh) script parses the commands out of the Markdown file and runs them. If you want to add examples, please make sure your instruction file can be parsed and run by that script.
```

examples/README.md
```diff
@@ -9,3 +9,13 @@ Here you can find various demos of Interledger.rs' functionality:
...
 Have questions? Feel free to [open an issue](https://github.com/emschwartz/interledger-rs/issues/new) or ask a question [on the forum](https://forum.interledger.org/)!
+
+## Running the Examples
+Each example includes instructions for running it step-by-step in the README.
+
+If you want to run all of the steps automatically, you can use the provided [`run-md.sh`](../scripts/run-md.sh) script to parse and execute the shell commands from the Markdown file:
+
+```bash
+# Under the example directory, for example, "simple"
+$ cat README.md | ../../scripts/run-md.sh
```
emschwartz: Maybe it would be even nicer to make it so you can call `../../scripts/run-md.sh README.md`. We could make it so the script checks for arguments passed to it and `cat`s the file if one was passed. If not, it would read from `stdin` instead

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,433 @@
...
 
-First, you need an Ethereum network. We'll use
-[ganache-cli](https://github.com/trufflesuite/ganache-cli) which deploys a local
-Ethereum testnet at `localhost:8545`. You're free to use any other
-testnet/mainnet you want. To install `ganache-cli`, run 
-`npm install -g ganache-cli`. You also need to have `redis-server` and
-`redis-cli` available in your PATH. In Ubuntu, you can obtain these by running
-`sudo apt-get install redis-server`
+## Overview
+This example shows how to configure Interledger.rs nodes and use an Ethereum network including testnet and mainnet as a settlement ledger underneath the nodes. If you are new to Ethereum, learn [here](https://www.ethereum.org/beginners/). To learn about settlement in Interledger, refer to [Peering, Clearing and Settling](https://github.com/interledger/rfcs/blob/master/0032-peering-clearing-settlement/0032-peering-clearing-settlement.md).
```
emschwartz:
```suggestion
This example shows how to configure Interledger.rs nodes and use an Ethereum network (testnet or mainnet) as a settlement ledger underneath the nodes. If you are new to Ethereum, you can learn about it [here](https://www.ethereum.org/beginners/). To learn about settlement in Interledger, refer to [Peering, Clearing and Settling](https://github.com/interledger/rfcs/blob/master/0032-peering-clearing-settlement/0032-peering-clearing-settlement.md).
```

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,433 @@
...
-Advanced: You can run this against the Rinkeby Testnet by running a node that
-connects to Rinkeby (e.g. `geth --rinkeby --syncmode "light"`) or use a
-third-party node provider such as [Infura](https://infura.io/). You must also [create a
-wallet](https://www.myetherwallet.com/) and then obtain funds via the
-[Rinkeby Faucet](https://faucet.rinkeby.io/).
+![overview](images/overview.svg)
 
-We will need **7** terminal windows in total to follow this tutorial in depth. You can run the
-provided `settlement_test.sh` script instead to see how the full process works.
+As you can see, a node needs its own settlement engine. So we launch two settlement engines.
```
emschwartz: Instead of including this line, I would just mention that "each node needs its own settlement engine" when you're starting up the SEs

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,433 @@
...
 
-For ease, you may want to set these environment variables to save you some
-typing:
-```
-ILP_DIR=<path to interledger-rs>
-ILP=$ILP_DIR/target/debug/interledger
-```
+- [An Ethereum network](#an-ethereum-network) to connect to
+- [Redis](#redis)
+- [geth](#geth)
```
emschwartz: Is that really required? Seems like we shouldn't need to install that if we're just using a local testnet.

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,433 @@
...
-HTTP Incoming Token: `in_bob`
-HTTP Outgoing Token: `out_bob`
-Settle To: `10`
+<!--! printf "Creating accounts:\n" -->
 
 ```bash
-curl http://localhost:8770/accounts -X POST \
-     -d "ilp_address=example.bob&asset_code=ETH&asset_scale=18&max_packet_amount=10&http_endpoint=http://127.0.0.1:8770/ilp&http_incoming_token=in_bob&outgoing_token=out_bob&settle_to=-10" \
-     -H "Authorization: Bearer hi_bob"
+# Adding settlemnt accounts should be done at the same time because it checks each other
```
emschwartz: This seems impossible to do without adding them from a script running on a single machine. I agree we should fix that

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,433 @@
...
+sleep 2
 ```
 
-Now we have both Alice and Bob up and running.
+Now two nodes and its settlement engines are set and accounts for each node are also set up.
+
+Notice how we use Alice's settlement engine endpoint while registering Bob. This means that whenever Alice interacts with Bob's account, she'll use that Settlement Engine.
 
-## 4. Peer each other.
+`settle_threshold`, `min_balance` and `settle_to` are parameters related to how often settlement is made, which is tracked in [issue 121](https://github.com/emschwartz/interledger-rs/issues/121).
```
emschwartz:
```suggestion
The `settle_threshold` and `settle_to` parameters control when settlements are triggered. The node will send a settlement when an account's balance reaches the `settle_threshold`, and it will settle for `balance - settle_to`.
```

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,433 @@
...
+
+### 7. Verify that addresses have been exchanged
```
emschwartz: What's the point of doing this? We should be able to assume it worked and just send a payment

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,433 @@
...
+-H "Authorization: Bearer alice" \
+http://localhost:8770/accounts/1/balance
 ```
 
-Bob pays Alice:
+<!--!
+printf "\n\n"
+printf "Waiting to settle...(This may take 10 to 20 secs depending on the block generation)\n"
+
+while [ `geth --exec "eth.blockNumber" attach http://localhost:8545 2>/dev/null` = $LAST_BLOCK_NUM ]
```
emschwartz: What is this doing?

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,433 @@
...
+### 10. Dump Transaction Logs
+
 ```bash
-curl localhost:8770/pay \
-        -d "{ \"receiver\" : \"http://localhost:7770\", \"source_amount\": 7  }" \
-        -H "Authorization: Bearer in_bob" -H "Content-Type: application/json"
+printf "Last block: "
+geth --exec "eth.getTransaction(eth.getBlock(eth.blockNumber-1).transactions[0])" attach http://localhost:8545 2>/dev/null
+printf "\nCurrent block: "
+geth --exec "eth.getTransaction(eth.getBlock(eth.blockNumber).transactions[0])" attach http://localhost:8545 2>/dev/null
```
emschwartz: I'm tempted to make this optional or suggest taking it out to avoid needing another dependency installed

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,433 @@
...
-The net after this should be that Alice's store shows positive 2 balance on her
-account and negative 2 for Bob. Bob's store should show the
-opposite:
+If you inspect ganache-cli's output, you will notice that the block number has increased as a result of the settlement executions.
 
-```bash
-echo "Bob's balance on Alice's store"
-curl localhost:7770/accounts/1/balance -H "Authorization: Bearer bob"
+### 11. Kill All the Services
+You can finally kill all the services like:
```
emschwartz:
```suggestion
Finally, you can stop all the services as follows:
```

examples/simple/README.md
```diff
@@ -1,27 +1,306 @@
...
+redis-server &> logs/redis.log &
+```
+
+When you want to watch logs, use the `tail` command. You can use the command like: `tail -f logs/redis.log`
+
+### 3. Launch 2 Nodes
+
+```bash
+# Turn on debug logging for all of the interledger.rs components and tower-web
+export RUST_LOG=tower-web,interledger/.*
```
```suggestion
export RUST_LOG=interledger/.*
```

scripts/run-md.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# Based on mdlp.awk from https://gist.github.com/trauber/4955706
+# Originally written by @trauber Rich Traube
+
```
emschwartz:
```suggestion

# This script parses and executes code blocks found in Markdown files.
# In addition to parsing code found between ```bash ... ```,
# it also parses code from special HTML comments that start with <!--!
# This allows us to hide certain commands from view in the rendered
# Markdown file but run them when the whole file is executed.
# This is largely intended for commands that make the output more readable

```

examples/eth-settlement/README.md
dora-gt: Actually if we just want to build a local test net, we don't need it but if we want to educate readers, I think it is better to have `geth` to check whether a settlement block is generated so that readers could understand what will happen if the balance exceeds the settlement threshold. This is originally done in `pay_dump.sh`.

hmmm but if you really want to remove it, I'll remove it.

dora-gt: What about moving it to an optional section.

examples/eth-settlement/README.md
dora-gt: Because ganache takes 10 to 30 seconds to generate a block, we have to wait for it to confirm the settlement is done. This waits for the block generation.

If we move `geth` to an optional section, I'll remove this.

examples/eth-settlement/README.md
dora-gt: OK, let's move this into an optional section.

examples/eth-settlement/README.md
dora-gt: I think this is concerned with the adding settlement accounts problem, so gakonst worried about it and added this section though it may be unnecessary. I'll remove this.

examples/eth-settlement/README.md
dora-gt: Moved it into `Advanced` section.

examples/eth-settlement/README.md
dora-gt: Removed.

examples/eth-settlement/README.md
dora-gt: Removed.

examples/eth-settlement/README.md
dora-gt: Moved it into `Advanced` section.

examples/README.md
dora-gt: Fixed.

examples/README.md
```diff
@@ -9,3 +9,16 @@ Here you can find various demos of Interledger.rs' functionality:
 1. Integrating Interledger Into Your App (Coming Soon!)
 
 Have questions? Feel free to [open an issue](https://github.com/emschwartz/interledger-rs/issues/new) or ask a question [on the forum](https://forum.interledger.org/)!
+
+## Running the Examples
+Each example includes instructions for running it step-by-step in the README.
```
pgarg-ripple: Suggest updating the sentence to - "The README of each example provides step-by-step instructions on how to the example."

emschwartz:
```suggestion
The README of each example provides step-by-step instructions on how to run the example.
```

dora-gt:
I had forgotten to add something important... the Rust environment although readers might have it already, some might not. Personally I think it is worth adding.

I'm going to add it anyway, if you think it is not required, I'll remove it later.

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,420 @@
...
 
-First, you need an Ethereum network. We'll use
-[ganache-cli](https://github.com/trufflesuite/ganache-cli) which deploys a local
-Ethereum testnet at `localhost:8545`. You're free to use any other
-testnet/mainnet you want. To install `ganache-cli`, run 
-`npm install -g ganache-cli`. You also need to have `redis-server` and
-`redis-cli` available in your PATH. In Ubuntu, you can obtain these by running
-`sudo apt-get install redis-server`
+## Overview
+This example shows how to configure Interledger.rs nodes and use an Ethereum network (testnet or mainnet) as a settlement ledger underneath the nodes. If you are new to Ethereum, you can learn about it [here](https://www.ethereum.org/beginners/). To learn about settlement in Interledger, refer to [Peering, Clearing and Settling](https://github.com/interledger/rfcs/blob/master/0032-peering-clearing-settlement/0032-peering-clearing-settlement.md).
```
emschwartz:
```suggestion
This example shows how to configure Interledger.rs nodes and use an Ethereum network (testnet or mainnet) as a settlement ledger for payments sent between the nodes. If you are new to Ethereum, you can learn about it [here](https://www.ethereum.org/beginners/). To learn about settlement in Interledger, refer to [Peering, Clearing and Settling](https://github.com/interledger/rfcs/blob/master/0032-peering-clearing-settlement/0032-peering-clearing-settlement.md).
```

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,420 @@
...
-    3. Redis Store
+## Prerequisites
 
-For ease, you may want to set these environment variables to save you some
-typing:
-```
-ILP_DIR=<path to interledger-rs>
-ILP=$ILP_DIR/target/debug/interledger
-```
+- [The Rust environment](#the-rust-environment)
```
emschwartz:
```suggestion
- [Rust](#rust)
```

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,420 @@
...
+- [An Ethereum network](#an-ethereum-network) to connect to
+- [Redis](#redis)
 
-## 1. Launch Ganache
+### The Rust Environment
```
emschwartz:
```suggestion
### Rust
```

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,420 @@
...
 
-This will launch an Ethereum testnet with 10 prefunded accounts. The mnemonic is
-used because we want to know the keys we'll use for Alice and Bob (otherwise
-they are randomized)
+Because Interledger.rs is written in the Rust language, you need the Rust environment. Refer to the [Getting started](https://www.rust-lang.org/learn/get-started) page or just `curl https://sh.rustup.rs -sSf | sh` and follow the instruction.
```
emschwartz:
```suggestion
Because Interledger.rs is written in the Rust language, you need the Rust environment. Refer to the [Getting started](https://www.rust-lang.org/learn/get-started) page or just `curl https://sh.rustup.rs -sSf | sh` and follow the instructions.
```

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,420 @@
...
 
-ganache-cli -m "abstract vacuum mammal awkward pudding scene penalty purchase
-dinner depart evoke puzzle" -i 1
+### An Ethereum network
+First, you need an Ethereum network. You could use your own local testnet or also the mainnet.
```
emschwartz:
```suggestion
First, you need an Ethereum network. You can either use a local testnet, a remote testnet, or the mainnet.
```

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,420 @@
...
+Fatal: Failed to start the JavaScript console: api modules: Post http://localhost:8545: context deadline exceeded
+```
+
+It seems that you failed to install `ganache-cli`. Try to install it.
 
+## Conclusion
 
-If you inspect ganache-cli's output, you will notice that the block number has
-increased as a result of the settlement executions.
+Done! Full E2E test between 2 users over SPSP utilizing the new settlement architecture in an Ethereum Ledger settlement engine.
```
emschwartz:
```suggestion
This example showed an SPSP payment sent between two Interledger.rs nodes that settled using on-ledger Ethereum transactions.
```

examples/simple/README.md
```diff
@@ -1,27 +1,315 @@
...
 
 ![overview](images/overview.svg)
 
 ## Prerequisites
 
-- [Redis](https://redis.io/download)
+- [The Rust environment](#the-rust-environment)
+- [Redis](#redis)
+
+### The Rust Environment
```
emschwartz: I would also change this to just "Rust" -- I'm not sure what you mean by the Rust environment

examples/simple/README.md
```diff
@@ -1,27 +1,315 @@
...
+To stop the process running on port `7770`, try `` kill `lsof -i:7770 -t` ``. Since this example launches 2 nodes, the port may be other than `7770`. Adjust the port number according to the situation.
 
-The [`send-payment.sh`](./send-payment.sh) script sends a payment from Alice to Bob that is routed from Node A to Node B.
+## Conclusion
 
-You can run the [`check-balances.sh`](./check-balances.sh) script to print each of the accounts' balances (try doing this before and after sending a payment).
+That's it for this example! You've learned how to set up Interledger.rs nodes, connect them together, and how to send a payment from one to the other.
 
-That's it for this example! Check out the [other examples](../README.md) for more complex demos that show more of the features of Interledger, including multi-hop routing and cross-currency payments.
+Check out the [other examples](../README.md) for more complex demos that show other features of Interledger, including multi-hop routing and cross-currency payments.
```
```suggestion
Check out the [other examples](../README.md) for more complex demos that show other features of Interledger, including settlement, multi-hop routing, and cross-currency payments.
```

examples/simple/README.md
dora-gt: Because it has many components, I thought it might suitable but I don't stick to it.

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,416 @@
...
 
-## 3. Configure Bob
+### 3. Launch Ganache
+
+This will launch an Ethereum testnet with 10 prefunded accounts. The mnemonic is used because we want to know the keys we'll use for Alice and Bob (otherwise they are randomized).
 
-1. In a new terminal, execute `redis-server --port 6380` to launch Redis for
-   Bob.
-2. Set Bob's key and address:
 ```bash
```
emschwartz: It would be helpful to add a printf line (that's only run when using `run-md.sh`) that says "Starting local Ethereum testnet"

emschwartz: Right now when I run this, it just prints the following, which doesn't really show me that anything happened with Ethereum. It would be nice to modify the comments to mention ethereum and potentially making sure it prints something related to the tx details.

```
Creating accounts:
Adding Alice's account...
Adding Bob's Account...
Adding Bob's account on Alice's node...
Adding Alice's account on Bob's node...

Checking balances...

Bob's balance on Alice's node: {"balance":"0"}
Alice's balance on Bob's node: {"balance":"0"}

Sending payment of 80 from Alice to Bob
{"delivered_amount":80}

Checking balances...

Bob's balance on Alice's node: {"balance":"10"}
Alice's balance on Bob's node: {"balance":"-80"}

Stopping Interledger nodes
2299
2317
2318
2434
2435
```

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,416 @@
...
 
 ```bash
-curl http://localhost:8770/accounts -X POST \
-     -d "ilp_address=example.bob&asset_code=ETH&asset_scale=18&max_packet_amount=10&http_endpoint=http://127.0.0.1:8770/ilp&http_incoming_token=in_bob&outgoing_token=out_bob&settle_to=-10" \
-     -H "Authorization: Bearer hi_bob"
+# Sending payment of 80 from Alice to Bob
+curl \
+    -H "Authorization: Bearer in_alice" \
+    -H "Content-Type: application/json" \
+    -d '{"receiver":"http://localhost:8770","source_amount":80}' \
```
emschwartz: It's not clear from running the example why the amounts are different from the `simple` example. It would probably be best to make the amounts the same. If we have different amounts showing up on different systems, we should explain why they're different.

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,416 @@
...
+    "asset_code": "ETH",
+    "asset_scale": 18,
+    "max_packet_amount": 10,
+    "settlement_engine_url": "http://localhost:3000",
+    "http_incoming_token": "bob",
+    "http_outgoing_token": "alice",
+    "http_endpoint": "http://localhost:8770/ilp",
+    "settle_threshold": 70,
+    "min_balance": -100,
+    "settle_to" : 10,
```
emschwartz: Looking at the logs, it's kind of confusing why the full amount wasn't settled
```suggestion
    "settle_to" : 0,
```

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,416 @@
...
-## 5. Make some payments!
+## Advanced
 
-A `pay_dump.sh` script is provided which you can use to make [SPSP
-payments](https://interledger.org/rfcs/0009-simple-payment-setup-protocol/)
-between the two. After making an SPSP payment, it subsequently dumps the state
-of the balance/prepaid_amount of each store, so that you can get a sense of what
-is happening under the hood after each payment. You can also monitor your
-connector and settlement engine's logs.
+### Check the Settlement Block Generation
```
emschwartz: Can you add some comments that get printed when using `run-md.sh` that mention the possibility of running these steps?

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,416 @@
-# Guide to E2E Testing Interledger settlement with the Ethereum Ledger
+# Interledger with Ethereum On-Ledger Settlement
```
emschwartz: This example didn't work for me the first time I ran it using `run-md.sh`. I saw the output:

```
Creating accounts:
Adding Alice's account...
Adding Bob's Account...
Adding Bob's account on Alice's node...
Adding Alice's account on Bob's node...

Checking balances...

Bob's balance on Alice's node: curl: (7) Failed to connect to localhost port 7770: Connection refused

Alice's balance on Bob's node: curl: (7) Failed to connect to localhost port 8770: Connection refused


Sending payment of 80 from Alice to Bob
curl: (7) Failed to connect to localhost port 7770: Connection refused


Checking balances...

Bob's balance on Alice's node: curl: (7) Failed to connect to localhost port 7770: Connection refused

Alice's balance on Bob's node: curl: (7) Failed to connect to localhost port 8770: Connection refused


Stopping Interledger nodes
31837
```

We should make sure there's more helpful output if something doesn't work properly

examples/eth-settlement/README.md
dora-gt: Fixed.

Also fixed for nodes to start up (I forgot to insert the script).

examples/eth-settlement/README.md
dora-gt: My intention was to make the amount exceed the settlement threshold. Hmmm, I'll insert an explanation of why.

examples/eth-settlement/README.md
dora-gt: I'm sorry I forgot to insert the script for this example. Added and should be fixed.

examples/eth-settlement/README.md
dora-gt: Wow. Great idea.

examples/eth-settlement/README.md
dora-gt: After applying this, I noticed that it became unclear whether the payment was done or not because it would output like:

```
Bob's balance on Alice's node: {"balance":"0"}
Alice's balance on Bob's node: {"balance":"0"}

Sending payment of 80 from Alice to Bob
{"delivered_amount":80}

Checking balances...

Bob's balance on Alice's node: {"balance":"0"}
Alice's balance on Bob's node: {"balance":"-80"}
```

`Bob's balance on Alice's node` is a bit confusing because it looks like nothing happened. So I think leaving this `10` is worthy.

emschwartz: Actually shouldn't Alice's balance on Bob's node be 0 after it receives the incoming settlement?

The script should probably print Bob's balance (on Bob's node) because that'll be the account that has a +80 balance after the payment has finished.

examples/eth-settlement/README.md
emschwartz: I agree that the amount should exceed the threshold but we should still use the same amounts as the simple example. For example, we could use an amount of 1000 for the payment and just set the `settle_threshold` accordingly so that it is hit. (Note we might actually want to use larger amounts because the normal scale is 6 or 9)

examples/eth-settlement/README.md
dora-gt: I'll revert this for now and add an explanation for this. If you think it still should be `0`, I'll fix later.

examples/eth-settlement/README.md
dora-gt: I see, indeed. I'll fix.

examples/eth-settlement/README.md
dora-gt: Yeah, I felt the same but I think it is because of the implementation.

While an outgoing settlement seems to decreases its balance right after a transaction is issued, an incoming settlement could only when the transaction is confirmed. So there might be a lag.

Hmmm this might be confusing but difficult. It might better to output some explanations for this when running `run-md.sh`. I'll add so look and decide later.

dora-gt: Yeah, I felt the same but I think it is because of the implementation.

While an outgoing settlement seems to decreases its balance right after a transaction is issued (even if the transaction is not contained in a block generated), an incoming settlement could increase only when the transaction is confirmed. So there might be a lag.

Hmmm this might be indeed confusing but difficult. It might better to output some explanations for this when running `run-md.sh`. I'll add and improve, so look and decide later.

examples/eth-settlement/README.md
dora-gt: Fixed to use `500` (the same as `simple`)

examples/eth-settlement/README.md
dora-gt: Fixed and tried to improve but... I don't know why it settles `310` and where the `10` comes from🤔

examples/eth-settlement/README.md
dora-gt: Added.

dora-gt:
Oops, there's a mistake, I'm fixing.

examples/eth-settlement/README.md
dora-gt: Perhaps... the reason is this:
https://github.com/emschwartz/interledger-rs/blob/master/crates/interledger-store-redis/src/store.rs#L90

> (balance > tonumber(settle_threshold))

I changed this to `(balance >= tonumber(settle_threshold))` and it settled `300` though there might be a reason for setting this `>`.

emschwartz: See the discussion in https://github.com/emschwartz/interledger-rs/issues/189

examples/eth-settlement/README.md
```diff
@@ -1,246 +1,493 @@
...
+    -d '{
+    "ilp_address": "example.alice",
+    "asset_code": "ETH",
+    "asset_scale": 18,
+    "max_packet_amount": 10,
+    "http_incoming_token": "in_alice",
+    "http_outgoing_token": "out_alice",
+    "http_endpoint": "http://localhost:7770/ilp",
+    "settle_to" : 0}' \
+    http://localhost:7770/accounts > logs/account-alice-alice.log 2>/dev/null
```
emschwartz: I'm not sure we need to hide the output of this

dora-gt: Regarding the `STDERR`, if we don't redirect the output, it will output like:
image
which looks a bit ugly.

If you are referring to `STDOUT`, indeed it could be `http://localhost:7770/accounts  2>/dev/null` but it breaks the consistency with the following adding account commands because they can**not** be done in the foreground. The following commands have to be done in the background **at the same time** (adding a settlement account problem).

emschwartz: Interesting, for me it only prints the result:
```
{"amount_per_minute_limit":null,"asset_code":"ETH","asset_scale":18,"btp_outgoing_token":null,"btp_uri":null,"http_endpoint":"http://localhost:7770/ilp","http_outgoing_token":"out_alice","id":0,"ilp_address":"example.alice","max_packet_amount":10,"min_balance":null,"packets_per_minute_limit":null,"receive_routes":false,"round_trip_time":500,"routing_relation":"Child","send_routes":false,"settle_threshold":null,"settle_to":0,"settlement_engine_url":null}
```

dora-gt: The image is the output when I do like `http://localhost:7770/accounts > logs/account-alice-alice.log`.

If I do
```
    "http_outgoing_token": "out_alice",
    "http_endpoint": "http://localhost:7770/ilp",
    "settle_to" : 0}' \
    http://localhost:7770/accounts
```
the output will be like:
```
Adding Alice's account...
{"amount_per_minute_limit":null,"asset_code":"ETH","asset_scale":18,"btp_outgoing_token":null,"btp_uri":null,"http_endpoint":"http://localhost:7770/ilp","http_outgoing_token":"out_alice","id":0,"ilp_address":"example.alice","max_packet_amount":10,"min_balance":null,"packets_per_minute_limit":null,"receive_routes":false,"round_trip_time":500,"routing_relation":"Child","send_routes":false,"settle_threshold":null,"settle_to":0,"settlement_engine_url":null}
```

I could do like that but it breaks the consistency as I mentioned above.

emschwartz: Ah, I see. Tracking this issue in https://github.com/emschwartz/interledger-rs/issues/212

Issue #185, opened 2019-07-31
Settlement Engine Asset Scale


gakonst:
1. Should the asset scale conversion be done on the connector or on the engine?
If we do it on the engine, it'd mean that each engine would have to implement that conversion properly, while we could do it on the operator once.

2. How do we handle settlement amounts that are bigger than u64 in the store? Can occur in high granularity EVM transactions (Ether / ERC20 / any other Ethereum fork)

cc @emschwartz @kincaidoneil @sappenin

emschwartz:
A couple of thoughts:
- The connector must be able to use atomic mathematical operations in the database
- All databases support u64 but I'm not sure most or any support larger formats
- Higher precision numbers are mostly an issue because the EVM chose a stupidly large number format, so ideally the implications of that decision would be limited to components that directly touch Ethereum-related things
- If the SE receives a settlement for an amount greater than what the connector supports, the SE might not have a way to "return" that settlement. Therefore, it's important that the connector be able to process whatever the SE sends it -- unless it would end up being up to the SE to submit many "incoming settlements" to represent the one larger amount it got

In practice, most applications using ETH seem to use Gwei instead of Wei, which makes the numbers more reasonable. Is this also true for ERC20 tokens? Or would you ever send 10^-18 Dai?

gakonst:
> In practice, most applications using ETH seem to use Gwei instead of Wei, which makes the numbers more reasonable. Is this also true for ERC20 tokens? Or would you ever send 10^-18 Dai?

If you want to support the micropayments use case, and if the asset you're using is very valuable, then it could be expected that you'd want sub 1e-9 payments

emschwartz:
Actually the question isn't really whether you want to support micropayments, it's whether you want to support micropayments _and_ larger payments. If you are using ETH (or an ERC20) to send micropayments, you could configure your scale to be 12 or 18 or something like that, but that would just limit how large a payment you can send.

kincaidoneil:
I propose we use the design in the RFC:

- **No additional configuration on the connector**. This is a potential footgun.
- **Stateless**. Less configuration to be persisted and looked up from the DB every request.
- **Different scales for different accounts**. If some accounts are denominated in different assets, but use the same SE, those accounts can cleanly use different different scales. This may be important to support multi-asset SE instances, which the current API doesn't preclude.

[...]

PR #192, opened 2019-08-05
Improve Asset Scale conversions

gakonst:
- Implements the logic discussed in #185.
- Removes `settlement_engine_asset_scale` since it is being passed in by the engine per request
- Improves the XRP/ETH test to take into account different decimals and exchange rate

Example:
Connector is making transactions for an account with `account_scale` 9 and `ETH`, so it denominates funds in Gwei. 

When the connector wants to settle, it sends the `amount` and `9` to the engine. 

The engine is configured to use `asset_scale = 18`. The engine will proceed to make an on-chain transaction for `amount * 1e9`. 

The receiver's engine will pick that up, and send `(1e9, 18)` to the receiver's connector. The receiver's connector will then downscale `1e9` from 18 to whatever `asset_scale` it uses for that account.


crates/interledger-settlement-engines/Cargo.toml
```diff
@@ -34,6 +34,7 @@ http = "0.1.17"
 clap = "2.32.0"
 clarity = { git =  "https://github.com/gakonst/clarity" }
 sha3 = "0.8.2"
+bigint = "4.4.1"
```
emschwartz: From the readme:
```
This crate is deprecated and will not be developed further. Users are invited to prefer the uint crate instead.
```

Seems like we should either use `uint` as the suggest or switch to [`num-bigint`](https://crates.io/crates/num-bigint), which seems to be the more popular library

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -757,13 +759,27 @@ where
...
+            result(BigU256::from_dec_str(&body.amount).map_err(move |err| {
                 let error_msg = format!("Error converting to U256 {:?}", err);
                 error!("{:?}", error_msg);
                 (StatusCode::from_u16(400).unwrap(), error_msg)
             }))
             .and_then(move |amount| {
+                // If we receive a Quantity { amount: "1", scale: 9},
+                // we must normalize it to our engine's scale (typically 18
+                // for ETH/ERC20).
+                // Slightly counter-intuitive that `body.scale` is set as `to`.
```
emschwartz: Yeah, what?

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -757,13 +759,27 @@ where
...
+                let amount = amount.normalize_scale(ConvertDetails {
+                    from: engine_scale,
+                    to: body.scale,
+                });
+                // Typecast to use the Convert trait implemented for
+                // bigint::uint::U256 while we're using ethereum_types::U256
+                // from rust-web3 which just re-exports it. Should be safe.
+                let amount: U256 = unsafe { mem::transmute(amount) };
```
emschwartz: Prefer not to use unsafe things here

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
```diff
@@ -234,7 +228,8 @@ fn eth_xrp_interoperable() {
                 client
                     .put(&format!("http://localhost:{}/rates", node2_http))
                     .header("Authorization", "Bearer admin")
-                    .json(&json!({"XRP": 1, "ETH": 1}))
+                    // Let's say 1 ETH = 0.001 XRP for this example
```
emschwartz:
```suggestion
                    // Let's say 0.001 ETH = 1 XRP for this example
```

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
```diff
@@ -343,12 +338,12 @@ fn eth_xrp_interoperable() {
...
-                                    assert_eq!(balances[2], "{\"balance\":\"-71\"}"); // alice has in total paid 71
-                                    assert_eq!(balances[3], "{\"balance\":\"-10\"}"); // bob owes 10 to bob.charlie
-                                    assert_eq!(balances[4], "{\"balance\":\"71\"}"); // bob.charlie has been paid 71
-                                    assert_eq!(balances[5], "{\"balance\":\"10\"}"); // bob.charlie is owed 10 by bob
+                                    assert_eq!(balances[0], -71000); // alice has in total paid 71k Gwei
+                                    assert_eq!(balances[1], -1000); // alice owes bob 1k Gwei
+                                    assert_eq!(balances[2], 1000); // bob is owed 1k Gwei by alice
+                                    assert_eq!(balances[3], -5000); // bob owes 5k drops to bob.charlie
+                                    assert_eq!(balances[4], 71000); // bob.charlie has been paid 71k drops
+                                    assert_eq!(balances[5], 5000); // bob.charlie is owed 5k drops by bob
```
emschwartz: I'm having a hard time figuring out how we ended up at these numbers

crates/interledger-settlement-engines/tests/test_helpers.rs
```diff
@@ -10,7 +10,8 @@ use std::time::Duration;
 
 const CONFIRMATIONS: u8 = 0;
 const CHAIN_ID: u8 = 1;
-pub const ETH_DECIMALS: u8 = 18;
+#[allow(unused)]
+pub const ETH_DECIMALS: u8 = 9;
```
emschwartz: If it's just for tests, I would include the constants directly in the code. For non-test code, I prefer to have the constants declared separately

crates/interledger-settlement/src/lib.rs
```diff
@@ -132,6 +128,19 @@ impl Convert for f64 {
     }
 }
 
+impl Convert for U256 {
+    fn normalize_scale(&self, details: ConvertDetails) -> U256 {
+        let from_scale = details.from;
+        let to_scale = details.to;
+        if from_scale >= to_scale {
+            self.overflowing_mul(U256::from(10u64.pow(u32::from(from_scale - to_scale))))
```
emschwartz: Seems unsafe to use `overflowing_mul`, no?

crates/interledger-settlement-engines/Cargo.toml
gakonst: https://github.com/paritytech/parity-common/tree/master/uint is the one used by Parity, I'd prefer to use that for compatibility with the rust-web3 library.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
gakonst: The alternative is to do: `U256::from_dec_str(amount.to_string()).unwrap()` which should be less performant. 

The value inside `amount` is the same type (array of 4 u64's). 

The issue is that rust-web3's exposed U256, is the [`ethereum-types`](https://github.com/paritytech/parity-common/tree/master/ethereum-types) U256, while the one I implemented in `interledger-settlement` is directly from the [`uint` ](https://github.com/paritytech/parity-common/tree/master/uint) crate.

crates/interledger-settlement-engines/Cargo.toml
gakonst: Now using num-bigint.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
gakonst: Turns out I had implemented it the wrong way, should be fixed in https://github.com/emschwartz/interledger-rs/pull/192/commits/2cdc8c7303c1ca192e6032ae1ceb159290e22986

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
gakonst: I've added a slightly longer explanation, what do you think?

gakonst:
Test fails with:

```
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target/debug/deps/eth_xrp_interoperable-4b4be70b3ccaf056

running 1 test
Listening for incoming XRP payments and polling Redis for accounts that need to be settled
Listening for incoming XRP payments and polling Redis for accounts that need to be settled
[2019-08-06T11:01:01Z ERROR interledger_btp::client] Error connecting to WebSocket server for account: 1 Io(Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" })
[2019-08-06T11:01:02Z ERROR interledger::node] No route found for outgoing account 1
[2019-08-06T11:01:02Z ERROR interledger::node] No route found for outgoing account 1
[2019-08-06T11:01:02Z ERROR interledger_http::client] Error sending HTTP request: Error(Hyper(Error(Connect, Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" })), "http://localhost:3010/ilp")
[2019-08-06T11:01:02Z ERROR interledger_http::client] Error sending HTTP request: Error(Hyper(Error(Connect, Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" })), "http://localhost:3020/ilp")
[2019-08-06T11:01:02Z ERROR interledger_settlement::api] Couldn't load idempotent data
[2019-08-06T11:01:02Z ERROR interledger_ccp::server] Got route broadcast with a routing loop (path includes us): Route { prefix: b"example.bob", path: [b"example.bob", b"example.alice"], auth: [43, 50, 219, 108, 44, 10, 98, 53, 251, 19, 151, 232, 34, 94, 168, 94, 15, 14, 110, 140, 123, 18, 109, 0, 22, 204, 189, 224, 230, 103, 21, 30], props: [] }
[2019-08-06T11:01:02Z ERROR interledger_settlement::api] Couldn't load idempotent data
[2019-08-06T11:01:02Z ERROR interledger_ccp::server] Got route broadcast with a routing loop (path includes us): Route { prefix: b"example.bob", path: [b"example.bob", b"example.alice"], auth: [43, 50, 219, 108, 44, 10, 98, 53, 251, 19, 151, 232, 34, 94, 168, 94, 15, 14, 110, 140, 123, 18, 109, 0, 22, 204, 189, 224, 230, 103, 21, 30], props: [] }
[2019-08-06T11:01:02Z ERROR interledger::node] No route found for outgoing account 1
[2019-08-06T11:01:02Z ERROR interledger_stream::client] Send money stopped because of error: SendMoneyError("Packet was rejected with error: F02 No outgoing route for account: 1 (ILP address of the Prepare packet: Address(\"example.bob.charlie.HBlM1WTNXUm0RN9MGPDgFjofyKeSWcEwWMN2FEeagMA\"))")
[2019-08-06T11:01:02Z ERROR interledger_spsp::client] Error sending payment: SendMoneyError("Packet was rejected with error: F02 No outgoing route for account: 1 (ILP address of the Prepare packet: Address(\"example.bob.charlie.HBlM1WTNXUm0RN9MGPDgFjofyKeSWcEwWMN2FEeagMA\"))")
time,max_amount_in_flight,amount_fulfilled
[2019-08-06T11:01:02Z ERROR interledger_api::routes::spsp] Error sending SPSP payment: SendMoneyError(70000)
Error sending SPSP payment: Error(Status(500), "http://localhost:3010/pay")
test eth_xrp_interoperable ... FAILED
```

I am unable to reproduce this locally.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -757,12 +758,31 @@ where
...
         Box::new(
-            result(U256::from_dec_str(&body.amount).map_err(move |err| {
+            result(BigUint::from_str(&body.amount).map_err(move |err| {
                 let error_msg = format!("Error converting to U256 {:?}", err);
                 error!("{:?}", error_msg);
                 (StatusCode::from_u16(400).unwrap(), error_msg)
             }))
+            .and_then(move |amount_from_connector| {
+                // If we receive a Quantity { amount: "1", scale: 9},
+                // we must normalize it to our engine's scale (typically 18
```
```suggestion
                // we must normalize it to our engine's scale
```

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -757,12 +758,31 @@ where
...
+                // for ETH/ERC20).
```
```suggestion
```

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -757,12 +758,31 @@ where
...
+
+                let amount = amount_from_connector.normalize_scale(ConvertDetails {
```
emschwartz: Is there any way this could fail?

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
```diff
@@ -343,12 +339,33 @@ fn eth_xrp_interoperable() {
...
-                                    assert_eq!(balances[4], "{\"balance\":\"71\"}"); // bob.charlie has been paid 71
-                                    assert_eq!(balances[5], "{\"balance\":\"10\"}"); // bob.charlie is owed 10 by bob
+                                    // Alice has paid Charlie in total 71k Gwei through Bob.
+                                    assert_eq!(balances[0], -71000);
+                                    // Since Alice has configured Bob's
+                                    // `settle_threshold` and `settle_to` to be
+                                    // 70k and -10k respectively, once she
+                                    // exceeded the 70k threshold, she made a 61k
+                                    // Gwei settlement to Bob so that their debt
+                                    // settles down to 10k.
```
emschwartz: Wait, if `settle_to` is -10k and the previous balance was 71k, shouldn't the settlement be sent for 81k?

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
```diff
@@ -343,12 +339,33 @@ fn eth_xrp_interoperable() {
...
+                                    // As Bob forwards money to Charlie, he also
+                                    // eventually exceeds the `settle_threshold`
+                                    // which incidentally is set to 70k. As a
+                                    // result, he must make a XRP ledger
+                                    // settlement of 66k Drops to get his debt
+                                    // back to the `settle_to` value of charlie,
+                                    // which is 5k (71k - 5k = 66k).
+                                    assert_eq!(balances[3], 5000);
+                                    // Charlie's balance indicates that he's
+                                    // received 71k drops from Alice
```
emschwartz:
```suggestion
                                    // received 71k drops (the total amount Alice sent him)
```

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
```diff
@@ -343,12 +339,33 @@ fn eth_xrp_interoperable() {
                                     get_balance(1, node3_http, "admin"),
```
emschwartz: This will be a lot clearer when the accounts are indexed by username

crates/interledger-settlement-engines/tests/test_helpers.rs
```diff
@@ -8,17 +8,16 @@ use std::str;
 use std::thread::sleep;
 use std::time::Duration;
 
-const CONFIRMATIONS: u8 = 0;
-const CHAIN_ID: u8 = 1;
-pub const ETH_DECIMALS: u8 = 18;
-#[allow(unused)]
-pub const XRP_DECIMALS: u8 = 6;
-
```
emschwartz: I think it's fine to use constants like this in the same file -- I just thought it was harder to follow the tests when they're used across files

crates/interledger-settlement/src/api.rs
```diff
@@ -157,32 +169,45 @@ impl_web! {
...
+                    let account_id = account.id();
+                    let amount = amount_from_engine.normalize_scale(ConvertDetails {
+                        from: engine_scale,
+                        to: account.asset_scale(),
+                    });
+                    // If we'd overflow, settle for the maximum u64 value
+                    let safe_amount = if let Some(amount) = amount.to_u64() {
+                        amount
+                    } else {
+                        std::u64::MAX
```
emschwartz: Interesting choice. It might be worth logging something here if this happens

crates/interledger-settlement/src/api.rs
```diff
@@ -157,32 +169,45 @@ impl_web! {
...
-                let ret = Bytes::from("Success");
-                Ok((StatusCode::OK, ret))
+                .and_then(move |amount_from_engine| {
+                    let account_id = account.id();
+                    let amount = amount_from_engine.normalize_scale(ConvertDetails {
+                        from: engine_scale,
+                        to: account.asset_scale(),
+                    });
+                    // If we'd overflow, settle for the maximum u64 value
+                    let safe_amount = if let Some(amount) = amount.to_u64() {
```
emschwartz: I think it's slightly less of a foot-gun to do this instead of using a different variable name:

```suggestion
                    let amount = if let Some(amount) = amount.to_u64() {
```

crates/interledger-settlement/src/api.rs
```diff
@@ -157,32 +169,45 @@ impl_web! {
...
+                        std::u64::MAX
+                    };
+                    store.update_balance_for_incoming_settlement(account_id, safe_amount, idempotency_key)
+                    .map_err(move |_| {
+                        let error_msg = format!("Error updating balance of account: {} for incoming settlement of amount: {}", account_id, amount);
+                        error!("{}", error_msg);
+                        (StatusCode::from_u16(500).unwrap(), error_msg)
+                    })
+                    .and_then(move |_| {
+                        // TODO: Return a Quantity of safe_amount and account.asset_scale()
```
emschwartz: This should be added to this PR

crates/interledger-settlement/src/lib.rs
```diff
@@ -46,12 +48,6 @@ impl Quantity {
 pub struct SettlementEngineDetails {
```
emschwartz: Is this struct needed if it only has one field in it now?

crates/interledger-settlement/src/lib.rs
```diff
@@ -110,24 +106,36 @@ pub trait Convert {
...
 impl Convert for u64 {
     fn normalize_scale(&self, details: ConvertDetails) -> Self {
-        let from_scale = details.from;
-        let to_scale = details.to;
-        if from_scale >= to_scale {
-            self * 10u64.pow(u32::from(from_scale - to_scale))
+        let scale_diff = (details.from as i8 - details.to as i8).abs() as u8;
+        let scale = 10u64.pow(scale_diff.into());
+        if details.to >= details.from {
+            self * scale
```
emschwartz: Should this math be done with bignums so it doesn't have the possibility of overflowing?

crates/interledger-settlement-engines/tests/test_helpers.rs
gakonst: Do you mind leaving it as is?

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
gakonst: Forgot to update the comment, `settle_to` is 10k in the account configuration. You are right, if it were -10k, then it'd settle for 81k.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
gakonst: How so? It receives a `BigUInt` and performs a multiplication/division that's safe:

```

impl Convert for BigUint {
    fn normalize_scale(&self, details: ConvertDetails) -> Self {
        let scale_diff = (details.from as i8 - details.to as i8).abs() as u8;
        let scale = 10u64.pow(scale_diff.into());
        if details.to >= details.from {
            self.mul(scale)
        } else {
            self.div(scale)
        }
    }
}
```

crates/interledger-settlement/src/api.rs
gakonst: I'd rather be explicit about the variable name here and leave it as `safe_amount`. Why do you think it's a footgun?

crates/interledger-settlement/src/lib.rs
gakonst: I opted to leave it there since we still haven't finalized all the settlement details, and might end up deciding to add back some more values, e.g. some settlement engine uid or similar. 

Happy to remove it and replace `SettlementEngineDetails` with `Url` instances once/if we agree that there is no more info required to refer to an engine.

crates/interledger-settlement/src/lib.rs
gakonst: Done, I have added a warning for `f64` overflows. I doubt they're ever gonna happen due to the current numbers used. Otherwise, we could add a [BigDecimals](https://github.com/akubera/bigdecimal-rs) crate. Note that the readme states that it was not added in the `num` crate for design reasons

crates/interledger-settlement/src/api.rs
gakonst: It has not been added to the [XRP engine](https://github.com/interledgerjs/settlement-xrp) which is part of our E2E tests, so I figured I'd add this feature in a separate PR - since it's not related to "improving asset scale conversions"

crates/interledger-settlement/src/api.rs
emschwartz: I like using shadowing when doing these kind of transformations on numbers, just to make sure that I don't accidentally use the wrong one in something after applying the transformation (and to make sure that anyone working on that code later will avoid the same mistake).

crates/interledger-settlement/src/lib.rs
emschwartz: Sounds good. Can you put a TODO or a note somewhere to track that?

crates/interledger-service-util/src/exchange_rates_service.rs
```diff
@@ -90,9 +90,13 @@ where
                 .build()));
             };
 
+            // Note that this is not an overflow safe operation. Given realistic
```
emschwartz: If it's not, I think it would be best to make sure that case properly creates an error and is handled correctly (in this case I think it should reject the packet)

crates/interledger-settlement/src/lib.rs
```diff
@@ -110,24 +109,48 @@ pub trait Convert {
 
 impl Convert for u64 {
     fn normalize_scale(&self, details: ConvertDetails) -> Self {
-        let from_scale = details.from;
-        let to_scale = details.to;
-        if from_scale >= to_scale {
-            self * 10u64.pow(u32::from(from_scale - to_scale))
+        let scale_diff = (details.from as i8 - details.to as i8).abs() as u8;
+        let scale = 10u64.pow(scale_diff.into());
+        let num = BigUint::from(*self);
```
emschwartz: It kind of seems like we should avoid BigUint math here but return a `Result` to indicate if it overflowed. Why is it better to round an overflowing amount down to u64::MAX?

crates/interledger-settlement/src/api.rs
gakonst: OK I agree with you.

crates/interledger-settlement/src/lib.rs
gakonst: If an engine sends a settlement request to the operator, I'd want the operator to update their store for as much as it can handle, hence round down to u64::MAX (and in the follow up PR #199 this will be taken into account by the engine).

What we can do is return indeed a result, and `unwrap_or(u64::MAX)` on the connector side, while for exchange rate conversions rate reject.

crates/interledger-service-util/src/exchange_rates_service.rs
gakonst: As of https://github.com/emschwartz/interledger-rs/pull/192/commits/449e4a6da33a22e286985f98dff99e65d2ebb909 a Reject is returned if the conversion fails

crates/interledger-settlement/src/lib.rs
gakonst: As of 449e4a6 a result is returned on overflow

crates/interledger-service-util/src/exchange_rates_service.rs
```diff
@@ -90,13 +90,52 @@ where
...
+                from: request.from.asset_scale(),
+                to: request.to.asset_scale(),
             });
-            let outgoing_amount = (request.prepare.amount() as f64) * scaled_rate;
-            request.prepare.set_amount(outgoing_amount as u64);
-            trace!("Converted incoming amount of: {} {} (scale {}) from account {} to outgoing amount of: {} {} (scale {}) for account {}", request.original_amount, request.from.asset_code(), request.from.asset_scale(), request.from.id(), outgoing_amount, request.to.asset_code(), request.to.asset_scale(), request.to.id());
+
+            match outgoing_amount {
+                Ok(outgoing_amount) => {
+                    // f64 which cannot fit in u64 gets cast as 0
```
emschwartz:
```suggestion
                    // If an f64 is larger than the maximum value for a u64 but gets
                    // cast to a u64, it will end up being 0
```

crates/interledger-service-util/src/exchange_rates_service.rs
```diff
@@ -90,13 +90,52 @@ where
...
+                    if outgoing_amount != 0.0 && outgoing_amount as u64 == 0 {
```
emschwartz: Will this check really work? :thinking:

crates/interledger-service-util/src/exchange_rates_service.rs
```diff
@@ -90,13 +90,52 @@ where
...
+                            data: &[],
+                        }
+                        .build()));
+                    }
+                    request.prepare.set_amount(outgoing_amount as u64);
+                    trace!("Converted incoming amount of: {} {} (scale {}) from account {} to outgoing amount of: {} {} (scale {}) for account {}",
+                        request.original_amount, request.from.asset_code(), request.from.asset_scale(), request.from.id(),
+                        outgoing_amount, request.to.asset_code(), request.to.asset_scale(), request.to.id());
+                }
+                Err(_) => {
```
emschwartz: When would this return an error versus turning into 0? This code is kind of confusing so it would definitely be helpful to put some comments describing what's going on

crates/interledger-service-util/src/exchange_rates_service.rs
gakonst: Added a test which shows that it does

PR #199, opened 2019-08-08
Connector & Engine adjustments for asset scale precision loss leftovers

### LINKED ISSUE #185: Settlement Engine Asset Scale

gakonst:
Closes #185


crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -504,11 +510,64 @@ where
...
+                                        // Scale the amount settled by the
+                                        // connector back up to our scale
+                                        let scaled_connector_amount = connector_amount
+                                            .normalize_scale(ConvertDetails {
+                                                from: quantity.scale,
+                                                to: engine_scale,
+                                            });
+
+                                        let diff: BigUint = engine_amount - scaled_connector_amount;
+                                        if diff > BigUint::from(0u32) {
```
emschwartz: Import `Zero` from `num_traits`
```suggestion
                                        if diff > Zero::zero() {
```

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -504,11 +510,64 @@ where
...
+                                            // connector settled less than we
+                                            // instructed it to, so we must save
+                                            // the difference for the leftovers
+                                            Either::A(store.save_leftovers(account_id_clone2, diff))
```
emschwartz: Couldn't this take the number format the ETH engine uses normally rather than a BigInt?

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -105,6 +107,21 @@ impl EthereumLedgerRedisStore {
     }
 }
 
+impl LeftoversStore for EthereumLedgerRedisStore {
+    fn save_leftovers(&self, account_id: String, leftovers: BigUint) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        let mut pipe = redis::pipe();
+        pipe.set(format!("leftovers:{}", account_id), leftovers.to_string())
```
emschwartz: What good is saving this amount if it won't be picked up and sent again? Shouldn't this go somewhere that the SE will look at later and try to tell the connector about?

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
gakonst: I expect that every engine built in Rust will use the `LeftoversStore` trait, so I figured we'd make this as generic as possible

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -487,39 +489,116 @@ where
...
-                        Ok(())
-                    } else {
-                        Err(())
-                    }
+            let account_id_clone2 = account_id.clone();
+            let amount = amount.clone();
+            let url = url.clone();
+
+            // settle for amount + leftovers
+            store.load_leftovers(account_id.clone())
```
gakonst: This call loads the value, but does not clear it, since it will be overwritten with the value sent by the connector after it's notified for settlement. Maybe it should? Can we have a race condition where there are multiple settlement calls for the same account, where the leftovers get read multiple times?

gakonst:
Adding a type_length_limit check at the main.rs file does not seem to allow tests to build:

```
cargo build --all-features --all-targets
   Compiling interledger-settlement-engines v0.1.0 (/Users/gakonst/projects/xpring-contract/interledger-rs/crates/interledger-settlement-engines)
error: reached the type-length limit while instantiating `<std::vec::IntoIter<ethereum_typ...]]], std::result::Result<(), !>>`
     |
     = note: consider adding a `#![type_length_limit="1689024"]` attribute to your crate

error: aborting due to previous error

error: Could not compile `interledger-settlement-engines`.
warning: build failed, waiting for other jobs to finish...
error: reached the type-length limit while instantiating `<std::vec::IntoIter<ethereum_typ...]]], std::result::Result<(), !>>`
     |
     = note: consider adding a `#![type_length_limit="1689024"]` attribute to your crate

error: aborting due to previous error

error: Could not compile `interledger-settlement-engines`.
warning: build failed, waiting for other jobs to finish...
error: reached the type-length limit while instantiating `<std::vec::IntoIter<ethereum_typ...]]], std::result::Result<(), !>>`
     |
     = note: consider adding a `#![type_length_limit="1689024"]` attribute to your crate

error: aborting due to previous error

error: Could not compile `interledger-settlement-engines`.
```

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
emschwartz: Yeah that's not safe. It should atomically read and change the "leftover" amount

crates/interledger-settlement-engines/src/stores/mod.rs
```diff
@@ -28,3 +29,18 @@ pub trait IdempotentEngineStore {
         data: Bytes,
     ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
 }
+
+pub trait LeftoversStore {
+    /// Saves the leftover data
+    fn save_leftovers(
+        &self,
+        account_id: String,
+        leftovers: BigUint,
```
emschwartz: I'm still not totally convinced this should be a `BigUint`. What if it were an associated type instead so that it would be up to the specific implementation what it returns?

For example, if the SE and associated store used `u64s`, it wouldn't make sense for the leftovers to be stored as a `BigUint` because we know it couldn't be more than a `u64`.

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -105,6 +108,43 @@ impl EthereumLedgerRedisStore {
...
+                .and_then(move |(_conn, _ret): (_, Value)| Ok(())),
+        )
+    }
+
+    fn load_leftovers(
+        &self,
+        account_id: String,
+    ) -> Box<dyn Future<Item = BigUint, Error = ()> + Send> {
+        let mut pipe = redis::pipe();
+        pipe.get(format!("leftovers:{}", account_id))
```
emschwartz: This should also atomically set that amount to zero

crates/interledger-settlement-engines/src/stores/mod.rs
```diff
@@ -28,3 +29,18 @@ pub trait IdempotentEngineStore {
...
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
+
+    /// Saves the leftover data
```
emschwartz: Wrong description

crates/interledger-settlement-engines/src/stores/mod.rs
```diff
@@ -28,3 +29,18 @@ pub trait IdempotentEngineStore {
...
+    fn load_leftovers(
```
emschwartz: If the behavior is changed to modify the leftovers value, it might be better to use a different name for this function

crates/interledger-settlement-engines/src/stores/mod.rs
gakonst: Good point, we should make it take an associated type and just use that.

crates/interledger-settlement-engines/src/stores/mod.rs
gakonst: What do you think about `pop_leftovers`? It loads the value and sets it back to 0.

emschwartz: Accurate but it sounds kind of strange to me. Maybe just `get_leftovers`?

Alternatively, "leftovers" is a kind of strange way to describe some part of the incoming settlement, so we could do something like `get_uncredited_settlement_amount`.

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
gakonst: Uses Redis' https://redis.io/commands/getset no

crates/interledger-settlement-engines/src/stores/mod.rs
gakonst: `load_and_clear_uncredited_settlement_amount` is a bit verbose but I like it.
`save_uncredited_settlement_amount` also

gakonst:
Maybe related: https://github.com/rust-lang/rust/issues/54540#issuecomment-494749912

gakonst:
@emschwartz this should be good for review again. Curious what you think on Box'ing the futures. Seems like there's no good alternative.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -288,64 +298,66 @@ where
...
-                    // combine all the futures for that range of blocks
-                    Either::B(join_all(notify_eth_txs_fut))
-                };
-
-                notify_all_txs_fut.and_then(move |ret| {
-                    trace!("Transactions settled {:?}", ret);
-                    // now that all transactions have been processed successfully, we
-                    // can save `fetch_until` as the latest observed block
-                    store_clone.save_recently_observed_block(fetch_until)
+        Box::new(
```
emschwartz: I'm fine with boxing the future here but it might be worth putting a comment to explain why it was necessary

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -288,64 +298,66 @@ where
...
+            web3.eth()
+                .block_number()
+                .map_err(move |err| error!("Could not fetch current block number {:?}", err))
+                .and_then(move |current_block| {
+                    trace!("Current block {}", current_block);
+                    // get the safe number of blocks to avoid reorgs
+                    let fetch_until = current_block - confirmations;
+                    // U256 does not implement IntoFuture so we must wrap it
```
emschwartz: I don't think any number types do so this comment seems unnecessary
```suggestion
```

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -105,6 +109,53 @@ impl EthereumLedgerRedisStore {
...
+    type AssetType = BigUint;
+
+    fn save_leftovers(
+        &self,
+        account_id: String,
+        leftovers: Self::AssetType,
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        trace!("Saving leftovers {:?} {:?}", account_id, leftovers);
+        let mut pipe = redis::pipe();
+        pipe.set(format!("leftovers:{}", account_id), leftovers.to_string())
```
emschwartz: This should be adding to that amount rather than overwriting what's there

gakonst: Yeah that's correct since there may be a race condition. However, the amount being written is a BigNumber, we cannot do arithmetic with these, unless we store them as numbers in redis. 

What should we do if the connector replies with leftovers greater than u64?

emschwartz: Kind of a janky solution but we could use a list of strings in Redis. When you want to save an amount, you would add another number to the list. When you want to get the leftovers, you would load the whole list (and empty it out -- although this creates a possibility that information would get lost if the SE crashes right then) and sum up the numbers parsed from the strings using the big integer library.

kincaidoneil: I like the list of strings idea. When we fetch the leftovers, I also think we can mitigate any risk of the SE crashing:
1. Generate a new idempotency key locally
2. Run a pipeline or Lua script that:
  a. Gets the uncredited list (and returns right away if it's empty)
  b. Sets new key(s) with the idempotency key, its expiration, and the list of amounts it corresponds to
  b. Empties the uncredited list
  d. Returns the list of amounts corresponding to the idempotency key
3. Then, the BigNumber logic locally can sum up the list of amounts for the incoming settlement request. If it crashes and comes back online, it can safely lookup all unexpired idempotency keys, and add up their list of amounts locally to compute the total amount that should be sent in each of the incoming settlement requests.
4. If that idempotency key expires/the request fails after a few retries, we can atomically append its list of amounts to the list of uncredited amounts and delete the idempotency key.

gakonst: The list of strings logic has been added, thanks for the suggestion @emschwartz.

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -351,27 +363,31 @@ mod tests {
...
-                    store
-                        .load_uncredited_settlement_amount(acc)
-                        .map_err(|err| eprintln!("Redis error: {:?}", err))
-                        .and_then(move |ret| {
-                            assert_eq!(amount, ret);
-                            let _ = context;
-                            Ok(())
-                        })
-                })
+            let amount = BigUint::from(100u64);
```
emschwartz: This test should use numbers that are larger than u64, just to ensure it does exactly what we expect

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -105,6 +109,83 @@ impl EthereumLedgerRedisStore {
...
+        account_id: String,
+        uncredited_settlement_amount: Self::AssetType,
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        trace!(
+            "Saving uncredited_settlement_amount {:?} {:?}",
+            account_id,
+            uncredited_settlement_amount
+        );
+        let mut pipe = redis::pipe();
+        pipe.lpush(
```
emschwartz: Might be worth including a comment recording why these amounts are stored as a list

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -105,6 +109,83 @@ impl EthereumLedgerRedisStore {
...
+
+    fn load_uncredited_settlement_amount(
+        &self,
+        account_id: String,
+    ) -> Box<dyn Future<Item = Self::AssetType, Error = ()> + Send> {
+        trace!("Loading uncredited_settlement_amount {:?}", account_id);
+        let mut pipe = redis::pipe();
+        // Loads the value and resets it to 0
+        pipe.lrange(
+            format!("uncredited_settlement_amount:{}", account_id),
```
emschwartz: Should this key be prefixed with the SE type like the other keys?

Issue #186, opened 2019-07-31
Replace numeric account IDs and configured ILP addresses with usernames and dynamically generated addresses


emschwartz:
- [x] Replace numeric IDs in `interledger-store-redis` / the node with unique ids
- [x] Add usernames to accounts as uid aliases
- [x] Make SPSP payments work with usernames
- [x] Add an auth method to the HTTP API that includes the username (HTTP Basic Auth, separate header, or JWT)
- [x] Remove need for auth tokens to be globally unique
- [x] Remove `ilp_address` from the `Account` (or make it optional, in which case if it is present it would override the automatically generated address)
- [x] Handle ILDCP queries by appending a dot (`.`) and the username to the Node's ILP address
- [x] Merge `IldcpAccount` trait into base `Account` trait and rename `client_address` method to `ilp_address`
- [ ] Don't broadcast node's child accounts if their addresses start with the node's ILP address (if node `example.a` has an account for `example.a.b`, it should broadcast a route to `example.a` but not to `example.a.b`)

Based on https://github.com/emschwartz/interledger-rs/issues/126, https://github.com/emschwartz/interledger-rs/issues/147, and https://github.com/emschwartz/interledger-rs/issues/161

kincaidoneil:
Will clients be able to change their usernames? If so, it might be nice to also have an immutable unique ID for each account to interface with settlement engines (e.g. a UUID).

emschwartz:
Right now I don't see allowing users to change their usernames as a high priority feature, and it's potentially a fairly bad idea to allow. What if you change your username and a friend or customer sends money to your old payment pointer?

gakonst:
Notes from call with Evan:

- unique username, no uuid 
- auth token key in database becomes (username,http_token) so that http_tokens are no longer globally unique 
- multiple authorization methods, 1 with Basic, and 1 with Bearer (`username:token` non-standard format)
- is there anytime a use for an account to have an ILP address that's not part of the node's address space?
- check if we can merge Ildcp account into Account

emschwartz:
> auth token key in database becomes (username,http_token) so that http_tokens are no longer globally unique

Alternatively, it might be better to just index all of the accounts by username and when the auth comes in, look up the account and check the token against what's listed in the account record. Currently we need a separate mapping of token -> accountId but I think we can get rid of that with this change.

kincaidoneil:
> unique username, no uuid

[...]

PR #205, opened 2019-08-12
feat: deletion of accounts from nodes

bstrie:
Adds a new HTTP endpoint that allows the admin of a node to delete accounts on that node.

Part of https://github.com/emschwartz/interledger-rs/issues/196

### DIFF
--- a/crates/interledger-store-redis/src/store.rs
+++ b/crates/interledger-store-redis/src/store.rs
@@ -1,3 +1,22 @@
+// The informal schema of our data in redis:
+//   send_routes_to         set         used for CCP routing
+//   receive_routes_from    set         used for CCP routing
+//   next_account_id        string      unique ID for each new account
+//   rates:current          hash        exchange rates
+//   routes:current         hash        dynamic routing table
+//   routes:static          hash        static routing table
+//   accounts:<id>          hash        details for each account
+//   http_auth              hash        maps hmac of cryptographic credentials to an account
+//   btp_auth               hash        maps hmac of cryptographic credentials to an account
+//   btp_outgoing
+// For interactive exploration of the store,
+// use the redis-cli tool included with your redis install.
+// Within redis-cli:
+//    keys *                list all keys of any type in the store
+//    smembers <key>        list the members of a set
+//    get <key>             get the value of a key
+//    hgetall <key>         the flattened list of every key/value entry within a hash
+
 use super::account::*;
 use super::crypto::generate_keys;
 use bytes::Bytes;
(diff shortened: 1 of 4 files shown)


crates/interledger-store-redis/src/store.rs
```diff
@@ -376,15 +395,54 @@ impl RedisStore {
...
+            // TODO: a retrieve_account API to avoid making Vecs which we only need one element of
+            self.retrieve_accounts(vec![id])
+                .and_then(|accounts| accounts.get(0).cloned().ok_or(()))
+                .and_then(|account| {
+                    let mut pipe = redis::pipe();
+                    pipe.atomic();
+
+                    pipe.del(account_details_key(account.id));
+
+                    // TODO: delete the keys from http_auth or btp_auth that correspond to this id
```
emschwartz: This should be done in this PR

emschwartz: Actually, the separate `http_auth` and `btp_auth` indexes are going to be removed when https://github.com/emschwartz/interledger-rs/issues/186 is implemented so never mind about fixing this TODO (once you authenticate with your username, we won't need a global index of auth tokens. Instead, the API can look up your username and get the associated token from the account record)

bstrie:
@emschwartz test added, let me know if it looks good to you and I'll squash and merge.

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -26,6 +26,23 @@ fn insert_accounts() {
...
 }
 
+#[test]
+fn delete_accounts() {
+    block_on(test_store().and_then(|(store, context)| {
+        store.get_all_accounts().and_then(move |accounts| {
+            let old_len = accounts.len();
+            store.remove_account(accounts[0].id()).and_then(move |_| {
+                store.get_all_accounts().and_then(move |accounts| {
+                    assert_eq!(old_len, accounts.len() + 1);
```
emschwartz: I would check to make sure the specific account you tried to remove is no longer there

bstrie: I wanted to ask, are all of these tests racing with each other on the same instance of redis? Because if so, then it's going to be difficult to test for this at all until #186 allows us to create accounts with specific IDs.

emschwartz: Each one should start a new Redis instance. See:
- https://github.com/emschwartz/interledger-rs/blob/es-fix-ccp-btp/crates/interledger-store-redis/tests/common/redis_helpers.rs#L126-L127
- https://github.com/emschwartz/interledger-rs/blob/es-fix-ccp-btp/crates/interledger-store-redis/tests/common/redis_helpers.rs#L83-L86

bstrie:
Changes made, commits squashed and rebased.

PR #207, opened 2019-08-13
CCP Fixes

emschwartz:
Fixes https://github.com/emschwartz/interledger-rs/issues/155

### DIFF
--- a/crates/interledger-btp/src/server.rs
+++ b/crates/interledger-btp/src/server.rs
@@ -266,7 +266,7 @@ fn get_auth(
             match message {
                 // Skip non-binary messages like Pings and Pongs
                 // Note that the BTP protocol spec technically specifies that
-                // the auth message MUST be the first packet send over the 
+                // the auth message MUST be the first packet sent over the 
                 // WebSocket connection. However, the JavaScript implementation
                 // of BTP sends a Ping packet first, so we should ignore it.
                 // (Be liberal in what you accept but strict in what you send)
--- a/crates/interledger-btp/src/server.rs
+++ b/crates/interledger-btp/src/server.rs
@@ -9,9 +9,16 @@ use log::{debug, error, warn};
 use ring::digest::{digest, SHA256};
 use std::{net::SocketAddr, str};
 use tokio_executor::spawn;
-use tokio_tcp::TcpListener;
-use tokio_tungstenite::{accept_async_with_config, stream::Stream as MaybeTlsStream};
-use tungstenite::protocol::{Message, WebSocketConfig};
+use tokio_tcp::{TcpListener, TcpStream};
+use tokio_tungstenite::{
+    accept_async_with_config, stream::Stream as WsStreamType, MaybeTlsStream, WebSocketStream,
+};
+use tungstenite::{
+    protocol::{Message, WebSocketConfig},
+    Error as WebSocketError,
+};
+
+type WsStream = WebSocketStream<MaybeTlsStream<TcpStream>>;
 
 const MAX_MESSAGE_SIZE: usize = 40000;
 
@@ -47,7 +54,7 @@ where
                 let service_clone = service_clone.clone();
                 let store = store.clone();
                 accept_async_with_config(
-                    MaybeTlsStream::Plain(stream),
+                    WsStreamType::Plain(stream),
                     Some(WebSocketConfig {
                         max_send_queue: None,
                         max_message_size: Some(MAX_MESSAGE_SIZE),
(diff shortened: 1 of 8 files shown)


crates/interledger-ccp/src/routing_table.rs
```diff
@@ -153,16 +153,23 @@ where
             return Ok(Vec::new());
         }
 
+        self.epoch = request.to_epoch_index;
+        trace!(
+            "Updated routing table {} to epoch: {}",
```
dora-gt: Though this is a super nitpick but it seems that the table is not updated at this time. It is actually updated after L184.

crates/interledger-ccp/src/server.rs
```diff
@@ -190,6 +190,7 @@ where
         let clone = self.clone();
         self.update_best_routes(None)
             .and_then(move |_| clone.send_route_updates())
+            .then(|_| Ok(()))
```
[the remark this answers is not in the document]
emschwartz: Returning an error here would stop the broadcast loop so we want to return `Ok` even if there was some error in the loop

crates/interledger-ccp/src/server.rs
```diff
@@ -730,17 +750,28 @@ fn get_best_route_for_prefix<A: CcpRoutingAccount>(
...
-                props: Vec::new(),
-            },
-        ));
+    // Check if we have a configured route for that specific prefix
+    // or any shorter prefix
+    // TODO: should we match results that don't line up with the dot-separated segments?
+    // (for example, should "example.a" match for the prefix "example.abc"?)
+    // note: we could also implement this by looking through a sorted list to find the longest
+    // matching prefix instead of using a HashMap lookup
+    let segments: Vec<&[u8]> = prefix.split(|c| c == &b'.').collect();
```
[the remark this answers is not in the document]
emschwartz: I'm going to hold off on making the switch for now. Most of this file uses `Bytes` instead of `Address` already (because it deals with prefixes that may not be valid ILP Addresses) and it's not that much logic to duplicate. Thanks for pointing it out though

crates/interledger-btp/src/server.rs
```diff
@@ -213,46 +250,79 @@ where
...
+            + Sink<SinkItem = Message, SinkError = WebSocketError>,
+    ),
+    Error = (),
+> {
     connection
+        .skip_while(|message| {
+            match message {
+                // Skip non-binary messages like Pings and Pongs
+                // Note that the BTP protocol spec technically specifies that
+                // the auth message MUST be the first packet send over the
```
gakonst:
```suggestion
                // the auth message MUST be the first packet sent over the 
```

crates/interledger-ccp/src/server.rs
```diff
@@ -730,17 +755,28 @@ fn get_best_route_for_prefix<A: CcpRoutingAccount>(
...
-            Route {
-                prefix: account.client_address().to_bytes(),
-                auth: [0; 32],
-                path: Vec::new(),
-                props: Vec::new(),
-            },
-        ));
+    // Check if we have a configured route for that specific prefix
+    // or any shorter prefix
+    // TODO: should we match results that don't line up with the dot-separated segments?
```
gakonst: I do not think so, routing should match by the longest prefix on a per separator basis IMO. `example.a` matches `example.a.b`, but not `example.abc`

crates/interledger-ccp/src/server.rs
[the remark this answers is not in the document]
gakonst: Using the address type here should make all these conversions easier by using the `segments` method and iterating over it

crates/interledger-ccp/src/server.rs
```diff
@@ -730,17 +755,28 @@ fn get_best_route_for_prefix<A: CcpRoutingAccount>(
...
+    // (for example, should "example.a" match for the prefix "example.abc"?)
+    // note: we could also implement this by looking through a sorted list to find the longest
+    // matching prefix instead of using a HashMap lookup
+    let segments: Vec<&[u8]> = prefix.split(|c| c == &b'.').collect();
+    for i in 0..segments.len() {
+        let prefix = &segments[0..segments.len() - i].join(&b'.');
+        if let Some(account) = configured_routes.get(prefix.as_ref() as &[u8]) {
```
gakonst: shouldn't we also take the prefix length into account?

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -84,6 +84,11 @@ where
         &mut self,
         request: OutgoingRequest<A>,
     ) -> Box<dyn Future<Item = Fulfill, Error = Reject> + Send> {
+        // Don't bother touching the store for zero-amount packets
+        if request.prepare.amount() == 0 && request.original_amount == 0 {
```
gakonst: When does one of `amount` and `original_amount` equal 0 but the other one does not?

crates/interledger-ccp/src/server.rs
emschwartz: This logic will find the longest matching configured prefix. 

For example, if your routing table looks like:

| Route Type | Prefix | Account to Forward To |
|---|---|---|
| Configured | `example.a` | 1 |
| Configured | `example.a.b` | 2 |
| Received from Peer | `example.a.b.c` | 3 |

And you are looking to route a packet for `example.a.b.c`, it will go to Account 2.

crates/interledger-service-util/src/balance_service.rs
emschwartz: If `original_amount` is 0, I think `amount` will always be 0. However, the `original_amount` could be positive but then rounded down to 0 when the exchange rate and scale change are applied.

emschwartz: Added a comment to explain this

PR #211, opened 2019-08-14
Fix #210: Add EchoService to Node

### LINKED ISSUE #210: Using `ilp-ping` to ping node that `moneyd` is connected to fails
I'm able to connect to a rust node via `moneyd` + `btp` successfully. However, when trying to use `ilp-ping`, I get the following error:

```vagrant@ubuntu-bionic:~$ ilp-ping test.xpring-dev.rs1
ILP-PING test.xpring-dev.rs1
Error sending ping. code=T01 message= triggeredBy=
Error sending ping. code=T01 message= triggeredBy=
Error sending ping. code=T01 message= triggeredBy=
Error sending ping. code=T01 message= triggeredBy=
```
In the above example , `moneyd` is connected to that same node `test.xpring-dev.rs1`.

dora-gt:
- Fixes #210 
- Add some `trace!` and `debug!` to improve bug investigations
- Add `EchoService` to the chain of `node`.
- Materials used in the investigation is [here](https://github.com/dora-gt/interledger-rs-backyard/tree/master/quality-assurance/issue-%23210).

`ilp-ping` should be able to `ping` like:
image


crates/interledger-btp/src/service.rs
```diff
@@ -233,6 +233,11 @@ where
                             Ok(fulfill) => Packet::Fulfill(fulfill),
                             Err(reject) => Packet::Reject(reject),
                         };
+                        match &packet {
```
emschwartz: I'm not sure we want to log the packet in the BTP handler

If we do though, it would be cleaner to move the log statements into the match above so the branch would look like:
```
Ok(fulfill) => {
    trace!("...");
    Packet::Fulfill(fulfill)
}
```

crates/interledger-packet/src/packet.rs
```diff
@@ -252,6 +255,10 @@ impl fmt::Debug for Prepare {
                 &hex::encode(self.execution_condition()),
             )
             .field("data_length", &self.data().len())
+            .field(
+                &format!("data[..{}]", data_limit),
+                &String::from_utf8_lossy(&self.data()[..data_limit]),
```
emschwartz: In most cases the data won't be UTF8 so I don't think it's worth printing (that was why I replaced printing the actual data with just printing the length)

crates/interledger-packet/src/packet.rs
```diff
@@ -492,6 +507,10 @@ impl fmt::Debug for Reject {
             )
             .field("triggered_by", &self.triggered_by())
             .field("data_length", &self.data().len())
+            .field(
+                &format!("data[..{}]", data_limit),
+                &String::from_utf8_lossy(&self.data()[..data_limit]),
```
emschwartz: Similarly, I don't think it's worth printing the fulfill or reject data

crates/interledger-service-util/src/echo_service.rs
```diff
@@ -59,8 +60,10 @@ where
         let should_echo = request.prepare.destination() == self.ilp_address
             && request.prepare.data().starts_with(ECHO_PREFIX.as_bytes());
         if !should_echo {
+            debug!("Not responding to this request: {:?}", request);
```
emschwartz: This is going to print on every other packet going through so we should remove this

crates/interledger-service-util/src/echo_service.rs
```diff
@@ -59,8 +60,10 @@ where
...
             return Box::new(self.next.handle_request(request));
         }
+        debug!("Responding to this request: {:?}", request);
```
emschwartz:
```suggestion
        debug!("Responding to Echo protocol request: {:?}", request);
```

crates/interledger-packet/src/packet.rs
dora-gt: yeah... it helped `ECHO` but it might be the only (or extremely rare) case. I'll remove.

PR #214, opened 2019-08-15
[WIP] Add usernames and account UUIDs

gakonst:
Still very WIP.

Will eventually fix https://github.com/emschwartz/interledger-rs/issues/186.

notes:
SPSP test will fail because the default account is not set.

### DIFF
--- a/crates/interledger-store-redis/src/account.rs
+++ b/crates/interledger-store-redis/src/account.rs
@@ -28,7 +28,7 @@ const ACCOUNT_DETAILS_FIELDS: usize = 21;
 
 #[derive(Clone, Debug, Serialize)]
 pub struct Account {
-    pub(crate) id: u64,
+    pub(crate) id: AccountId,
     #[serde(serialize_with = "address_to_string")]
     pub(crate) ilp_address: Address,
     // TODO add additional routes
@@ -99,7 +99,7 @@ where
 }
 
 impl Account {
-    pub fn try_from(id: u64, details: AccountDetails) -> Result<Account, ()> {
+    pub fn try_from(id: AccountId, details: AccountDetails) -> Result<Account, ()> {
         let http_endpoint = if let Some(ref url) = details.http_endpoint {
             Some(Url::parse(url).map_err(|err| error!("Invalid URL: {:?}", err))?)
         } else {
@@ -177,6 +177,55 @@ impl AccountWithEncryptedTokens {
         }
 
         self.account
+    }
+}
+
+use std::fmt::Display;
+use uuid::Uuid;
+
+#[derive(Eq, Hash, Debug, Default, Serialize, Copy, Clone)]
+pub struct AccountId(Uuid);
+
+impl AccountId {
+    pub fn new() -> Self {
+        let uid = Uuid::new_v4();
+        AccountId(uid.clone())
+    }
+}
+
+impl PartialEq for AccountId {
+    fn eq(&self, other: &AccountId) -> bool {
+        self.0.eq(&other.0)
+    }
+}
+
+impl FromStr for AccountId {
+    type Err = ();
+
+    fn from_str(src: &str) -> Result<Self, Self::Err> {
+        let uid = Uuid::from_str(&src).map_err(|_| ())?;
+        Ok(AccountId(uid.clone()))
+    }
+}
+
+impl Display for AccountId {
+    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> Result<(), ::std::fmt::Error> {
+        f.write_str(&self.0.to_hyphenated().to_string())
+    }
+}
+
+impl ToRedisArgs for AccountId {
+    fn write_redis_args(&self, out: &mut Vec<Vec<u8>>) {
+        self.0.to_hyphenated().to_string().write_redis_args(out);
+    }
+}
+
+impl FromRedisValue for AccountId {
+    fn from_redis_value(v: &Value) -> Result<Self, RedisError> {
+        let account_id = String::from_redis_value(v)?;
+        let uid = Uuid::from_str(&account_id)
+            .map_err(|_| RedisError::from((ErrorKind::TypeError, "Invalid account id string")))?;
+        Ok(AccountId(uid))
     }
 }
 
@@ -366,7 +415,7 @@ fn get_bool(key: &str, map: &HashMap<String, Value>) -> bool {
 }
 
 impl AccountTrait for Account {
-    type AccountId = u64;
+    type AccountId = AccountId;
 
     fn id(&self) -> Self::AccountId {
         self.id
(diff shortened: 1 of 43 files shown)


crates/interledger-store-redis/src/account.rs
```diff
@@ -180,6 +180,55 @@ impl AccountWithEncryptedTokens {
...
+    fn eq(&self, other: &AccountId) -> bool {
+        self.0.eq(&other.0)
+    }
+}
+
+impl FromStr for AccountId {
+    type Err = ();
+
+    fn from_str(src: &str) -> Result<Self, Self::Err> {
+        let uid = Uuid::from_str(&src).map_err(|_| ())?;
```
emschwartz: Might be worth just passing the error through

crates/interledger-store-redis/src/account.rs
```diff
@@ -180,6 +180,55 @@ impl AccountWithEncryptedTokens {
     }
 }
 
+use std::fmt::Display;
+use uuid::Uuid;
+
+#[derive(Eq, Hash, Debug, Default, Serialize, Copy, Clone)]
+pub struct AccountId(Uuid);
```
emschwartz: Could you put a comment explaining why we have an `AccountId` struct instead of just using a `Uuid` directly?

crates/interledger-store-redis/src/store.rs
```diff
@@ -271,19 +271,28 @@ impl RedisStoreBuilder {
...
     encryption_key: Arc<aead::SealingKey>,
     decryption_key: Arc<aead::OpeningKey>,
 }
 
 impl RedisStore {
-    fn get_next_account_id(&self) -> impl Future<Item = u64, Error = ()> {
-        cmd("INCR")
-            .arg(NEXT_ACCOUNT_ID_KEY)
-            .query_async(self.connection.as_ref().clone())
+    fn generate_account_id(&self) -> impl Future<Item = AccountId, Error = ()> {
```
emschwartz: This doesn't really make sense to me. Why not have a `new` method on `AccountId` and then just add the ID to `accounts` when on the insert account call

crates/interledger-store-redis/src/store.rs
```diff
@@ -759,46 +772,35 @@ impl NodeStore for RedisStore {
...
     // TODO limit the number of results and page through them
     fn get_all_accounts(&self) -> Box<dyn Future<Item = Vec<Self::Account>, Error = ()> + Send> {
         let decryption_key = self.decryption_key.clone();
-        Box::new(
-            cmd("GET")
-                .arg(NEXT_ACCOUNT_ID_KEY)
-                .query_async(self.connection.as_ref().clone())
+        let mut pipe = redis::pipe();
+        let connection = self.connection.clone();
+        pipe.smembers("accounts");
```
emschwartz: Seems like this pipe is discarded

crates/interledger-store-redis/src/store.rs
```diff
@@ -1229,7 +1247,7 @@ impl SettlementStore for RedisStore {
 
     fn update_balance_for_incoming_settlement(
         &self,
-        account_id: u64,
+        account_id: <Self::Account as AccountTrait>::AccountId,
```
emschwartz: All of the `<Self::Account as AccountTrait>::AccountId`s can be replaced with just `AccountId`

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -98,63 +102,71 @@ fn get_all_accounts() {
...
-            assert_eq!(
-                accounts[0].client_address(),
-                &Address::from_str("example.bob").unwrap()
-            );
-            let _ = context;
-            Ok(())
-        })
+    block_on(test_store().and_then(|(store, context, accs)| {
+        let store_clone = store.clone();
+        let acc = accs[0].clone();
```
emschwartz: I would make this a specific account and hard-code the ID its getting in the test

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -98,63 +102,71 @@ fn get_all_accounts() {
...
-            Ok(())
-        })
+    block_on(test_store().and_then(|(store, context, accs)| {
+        let store_clone = store.clone();
+        let accs = vec![accs[1].clone(), accs[0].clone()];
+        let account_ids: Vec<AccountId> = accs.iter().map(|a| a.id()).collect::<_>();
+        store_clone
+            .get_accounts(account_ids)
+            .and_then(move |accounts| {
+                // note reverse order is intentional
```
emschwartz: Huh? That doesn't make sense. If I pass in a vec of IDs, I'd expect the position in the array of each account to be the same for the original vec and the results

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -98,63 +102,71 @@ fn get_all_accounts() {
...
+fn decrypts_outgoing_tokens_acc() {
+    block_on(test_store().and_then(|(store, context, accs)| {
+        let acc = accs[0].clone();
+        store
+            .get_accounts(vec![acc.id()])
+            .and_then(move |accounts| {
+                let account = accounts[0].clone();
+                assert_eq!(
+                    account.get_http_auth_token().unwrap(),
+                    acc.get_http_auth_token().unwrap(),
```
emschwartz: Again, better to be specific in the tests

crates/interledger-store-redis/tests/common/store_helpers.rs
```diff
@@ -12,17 +12,25 @@ lazy_static! {
     static ref TEST_MUTEX: Mutex<()> = Mutex::new(());
 }
 
-pub fn test_store() -> impl Future<Item = (RedisStore, TestContext), Error = ()> {
+pub fn test_store() -> impl Future<Item = (RedisStore, TestContext, Vec<Account>), Error = ()> {
```
emschwartz: What's the point of returning the vec of accounts?

crates/interledger/tests/three_nodes.rs
```diff
@@ -44,7 +44,7 @@ fn three_nodes() {
 
     let node1 = InterledgerNode {
         ilp_address: Address::from_str("example.one").unwrap(),
-        default_spsp_account: Some(0),
+        default_spsp_account: None,
```
emschwartz: You're going to switch this back to being `Some`, right?

emschwartz:
Also, shouldn't the title of this PR be something like "Add usernames and account UUIDs" or something like that? (It's switching to _non_-numeric account IDs)

crates/interledger-store-redis/src/store.rs
gakonst: If `Account` stopped using the `AccountId` datatype for the `id` field, and suddenly used another variable name, then it'd not compile (I think?). This lets it be more general

crates/interledger-store-redis/tests/accounts_test.rs
gakonst: This was already there from your tests, illustrating that the returned elements come in the same order as the argument. It has the same functionality as before.

gakonst:
Closing this. Will be split in several PRs:1
1. Generalize Store data type
2. Add usernames + auto-generate ilp address from it in ildcp requests (unless overriden)
3. per account auth tokens
4. CCP broadcasting fixes

Issue #109, opened 2019-06-18
Validator service thinks packets are expired


emschwartz:
Occasionally while running the tests, the `validator_service` seems to think that all packets being sent are expired:

```
[2019-06-18T22:02:53Z ERROR interledger_service_util::validator_service] Incoming packet expired 85ms ago at SystemTime { tv_sec: 1560895373, tv_nsec: 76000000 } (time now: SystemTime { tv_sec: 1560895373, tv_nsec: 161851819 })
[2019-06-18T22:02:53Z ERROR interledger_service_util::validator_service] Incoming packet expired 84ms ago at SystemTime { tv_sec: 1560895373, tv_nsec: 180000000 } (time now: SystemTime { tv_sec: 1560895373, tv_nsec: 264038171 })
[2019-06-18T22:02:53Z ERROR interledger_service_util::validator_service] Incoming packet expired 85ms ago at SystemTime { tv_sec: 1560895373, tv_nsec: 283000000 } (time now: SystemTime { tv_sec: 1560895373, tv_nsec: 368878297 })
[2019-06-18T22:02:53Z ERROR interledger_service_util::validator_service] Incoming packet expired 91ms ago at SystemTime { tv_sec: 1560895373, tv_nsec: 388000000 } (time now: SystemTime { tv_sec: 1560895373, tv_nsec: 479655783 })
[2019-06-18T22:02:53Z ERROR interledger_service_util::validator_service] Incoming packet expired 93ms ago at SystemTime { tv_sec: 1560895373, tv_nsec: 497000000 } (time now: SystemTime { tv_sec: 1560895373, tv_nsec: 590578699 })
[2019-06-18T22:02:53Z ERROR interledger_service_util::validator_service] Incoming packet expired 87ms ago at SystemTime { tv_sec: 1560895373, tv_nsec: 609000000 } (time now: SystemTime { tv_sec: 1560895373, tv_nsec: 696319490 })
[2019-06-18T22:02:53Z ERROR interledger_service_util::validator_service] Incoming packet expired 86ms ago at SystemTime { tv_sec: 1560895373, tv_nsec: 715000000 } (time now: SystemTime { tv_sec: 1560895373, tv_nsec: 801855038 })
```

I'm not sure what causes this but wonder if it might be some weirdness with using [`SystemTime`](https://doc.rust-lang.org/std/time/struct.SystemTime.html):
[...]

PR #218, opened 2019-08-15
Use chrono for date parsing and comparisons in the validator

emschwartz:
Related to https://github.com/emschwartz/interledger-rs/issues/109

I'm not sure whether switching to `chrono` will solve the problem, but a) it might and b) the logs will be a bit more readable if it does.

Unfortunately the issue identified in #109 is hard to reproduce and it's unclear why it happens so I can't think of a better fix for it now. Anyone else have other ideas? If not, we could merge this and wait to see if the problem still arises in future builds.

### DIFF
--- a/crates/interledger-service-util/Cargo.toml
+++ b/crates/interledger-service-util/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/emschwartz/interledger-rs"
 [dependencies]
 bytes = "0.4.12"
 byteorder = "1.3.1"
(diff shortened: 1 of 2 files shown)


dora-gt:
I'm not sure [this](https://github.com/emschwartz/interledger-rs/blob/0bc4c37b7ad0103e97c678dddc8fa4d9d8f3ce4e/crates/interledger-service-util/src/validator_service.rs#L148) is accurate. Is that OK? (`now` is defined above).

PR #220, opened 2019-08-16
Generalize Account Ids

gakonst:
Part of fixing #186.

Adds a `AccountId` struct which uses `Uuid` internally for connector accounts. Each new generated account creates a new random uuid. 

In order to send an SPSP payment to the non-default account, users are required to specify the receiver's uuid on the receiver's store. This is slightly bad UX, and will be solved by adding username aliases in a follow up PR.

The engine now no longer uses the `Account` trait (in preparation of the upcomign `IldcpAccount` into `Account` trait merge), and instead defines a `id` method in `EthereumAccount`.


crates/interledger-ccp/src/lib.rs
```diff
@@ -26,8 +26,10 @@ mod test_helpers;
 
 pub use server::{CcpRouteManager, CcpRouteManagerBuilder};
 
+use serde::{Deserialize, Serialize};
+
 #[repr(u8)]
-#[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
+#[derive(Clone, Copy, Debug, PartialEq, PartialOrd, Serialize, Deserialize)]
```
emschwartz: What is this needed for?

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -768,6 +768,7 @@ where
         let self_clone = self.clone();
         let store: S = self.store.clone();
         let account_id = account_id.id;
+        debug!("Got account id to parse: {:?}", account_id);
```
emschwartz: Seems like this should be removed. It'll already log an error if it gets one it cannot parse

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
```diff
@@ -16,6 +15,7 @@ use log::{error, trace};
 
 use crate::stores::redis_store_common::{EngineRedisStore, EngineRedisStoreBuilder};
 use crate::stores::LeftoversStore;
+use interledger_store_redis::AccountId;
```
emschwartz: I don't think this crate should depend on `interledger_store_redis` at all

crates/interledger-store-redis/src/store.rs
```diff
@@ -1229,7 +1238,7 @@ impl SettlementStore for RedisStore {
 
     fn update_balance_for_incoming_settlement(
         &self,
-        account_id: u64,
+        account_id: <Self::Account as AccountTrait>::AccountId,
```
```suggestion
        account_id: AccountId,
```

crates/interledger-store-redis/src/store.rs
```diff
@@ -1250,7 +1259,7 @@ impl SettlementStore for RedisStore {
 
     fn refund_settlement(
         &self,
-        account_id: u64,
+        account_id: <Self::Account as AccountTrait>::AccountId,
```
```suggestion
        account_id: AccountId,
```

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -98,63 +102,71 @@ fn get_all_accounts() {
...
-            Ok(())
-        })
+    block_on(test_store().and_then(|(store, context, accs)| {
+        let store_clone = store.clone();
+        let accs = vec![accs[1].clone(), accs[0].clone()];
+        let account_ids: Vec<AccountId> = accs.iter().map(|a| a.id()).collect::<_>();
+        store_clone
+            .get_accounts(account_ids)
+            .and_then(move |accounts| {
+                // note reverse order is intentional
```
emschwartz: This still seems like a bug, no?

crates/interledger/src/main.rs
```diff
@@ -337,7 +338,10 @@ pub fn main() {
                             .ok(),
                         settlement_engine_url: None,
                     };
-                    tokio::run(insert_account_redis(redis_uri, &server_secret, account));
+                    tokio::run(
+                        insert_account_redis(redis_uri, &server_secret, account)
+                            .and_then(move |_| Ok(())),
```
emschwartz: Shouldn't it throw an error if this call fails?

crates/interledger/tests/test_helpers.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub delivered_amount: u64,
+}
+
+#[derive(serde::Deserialize)]
+pub struct BalanceData {
+    pub balance: String,
+}
+
+#[allow(unused)]
+pub fn create_account<T: Serialize>(
```
emschwartz: Seems like this should make it clear that it's creating an account on the SE rather than the connector

crates/interledger-ccp/src/lib.rs
gakonst: `RoutingRelation` exists in `Account` in store_redis, and it's needed to serialize/deserialize it. Previously you were serializing it manually.

crates/interledger-store-redis/tests/accounts_test.rs
gakonst: We request account ids in reverse order, thus we get accounts in the reverse order, where do you think there's a bug? I've made the code slightly clearer on that part.

crates/interledger/src/main.rs
gakonst: I think it does? I had to add the `and_then` because the return type for `tokio::run` has to be empty.

crates/interledger-store-redis/tests/accounts_test.rs
emschwartz: Oh, right. Now I don't remember why I made it like that

crates/interledger/src/main.rs
emschwartz: Oh my bad, I misread that as a `.then(move |_| Ok(()))`

crates/interledger-settlement-engines/src/stores/redis_ethereum_ledger/store.rs
gakonst: Running an engine API requires an [idempotent store](https://github.com/emschwartz/interledger-rs/blob/master/crates/interledger-settlement-engines/src/api.rs#L29). 

I had copied it in the engines crate so that we did not depend on it, but then the idempotent store was refactored to be only in `interledger_store_redis` (which is what resulted in this PR https://github.com/emschwartz/interledger-rs/pull/221) to avoid code duplication.

emschwartz: I don't understand. Isn't the `IdempotentStore` implemented in [`redis_store_common`](https://github.com/emschwartz/interledger-rs/blob/gakonst-generic-account-ids/crates/interledger-settlement-engines/src/stores/redis_store_common.rs)?

gakonst: I hadn't merged master at the time. Anyway, https://github.com/emschwartz/interledger-rs/pull/221 is reverted in this PR, the redis dependency is only a dev dependency now for tests. We now use String types only in engine keys, in order to avoid making ANY assumption about what type of type the `account_id` is.

gakonst:
This keeps failing at random e2e tests with the error:

```
[2019-08-17T18:41:35Z ERROR interledger_http::client] Error sending HTTP request: Error(Hyper(Error(Connect, Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" })), "http://localhost:3010/ilp")
```

Unclear what causes it. It was happening every now and then before, but now it seems more consistent.

PR #222, opened 2019-08-16
Enforce maximum expiry duration

### LINKED ISSUE #163: Ensure that we enforce a maximum timeout for transactions
This came up in a meeting today regarding attacks against distributed systems and @emschwartz wanted to make sure that we enforce such timeouts.

emschwartz:
Resolves #163

### DIFF
--- a/crates/interledger-service-util/src/expiry_shortener_service.rs
+++ b/crates/interledger-service-util/src/expiry_shortener_service.rs
@@ -1,10 +1,12 @@
+use chrono::{DateTime, Duration, Utc};
 use interledger_service::{Account, OutgoingRequest, OutgoingService};
-use std::time::Duration;
+use log::debug;
 
-pub const DEFAULT_ROUND_TRIP_TIME: u64 = 500;
+pub const DEFAULT_ROUND_TRIP_TIME: u32 = 500;
+pub const DEFAULT_MAX_EXPIRY_DURATION: u32 = 30000;
 
 pub trait RoundTripTimeAccount: Account {
-    fn round_trip_time(&self) -> u64 {
+    fn round_trip_time(&self) -> u32 {
         DEFAULT_ROUND_TRIP_TIME
     }
 }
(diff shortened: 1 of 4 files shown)


crates/interledger-service-util/src/expiry_shortener_service.rs
```diff
@@ -38,11 +49,131 @@ where
...
     /// 2. Reduce the packet's expiry by that amount
-    /// 3. Forward the request
+    /// 3. Ensure that the packet expiry does not exceed the maximum expiry duration
+    /// 4. Forward the request
     fn send_request(&mut self, mut request: OutgoingRequest<A>) -> Self::Future {
-        let time_to_subtract = request.from.round_trip_time() + request.to.round_trip_time();
-        let new_expiry = request.prepare.expires_at() - Duration::from_millis(time_to_subtract);
-        request.prepare.set_expires_at(new_expiry);
+        let time_to_subtract =
+            i64::from(request.from.round_trip_time() + request.to.round_trip_time());
```
dora-gt: Just from my curiosity, what is the strategy of setting this `time_to_subtract`?

Let's say there are 4 nodes: `A`, `B`, `C`, `D`. Each RTT **of next node**  is `2` sec. If a packet just goes from `A` to `B` and back to `A`, it takes `2` secs (`1` sec each way). Now, I think the minimum (not realistic but just for example) required duration is `6` secs like the following.

```
# not considering in-node process time
A -(1sec)-> B -(1sec)-> C -(1sec)-> D (fulfill)
A <-(1sec)- B <-(1sec)- C <-(1sec)- D
```

Then if the logic applied to this situation, the required duration would be `8` secs because node `B` and `C` reduces `4` secs each. It seems not efficient because they have duplicated areas.

I suppose these principles:

1. The original sender supposes the required duration as `(estimated hops) * (estimated one hop RTT)`.
    - In this case, it would be `6` secs because `estimated hops = 3, estimated one hop RTT = 2 secs`.
    - I know this might be difficult because we don't have the means to estimate hops.
1. Each node subtracts only `RTT of sender node / 2 (= 1 sec)`.
    - This is based on a principle of `not to lose MY fund (=I need to reduce the time to send back the fulfill to the sender. If I subtract 1 sec, that should be sufficient)`.

There might be some more buffer but I think this is simple and easy to understand.

emschwartz: I'm going to open a new issue to discuss this, since the `round_trip_time` behavior is only tangentially related to enforcing the maximum expiry duration

PR #223, opened 2019-08-16
feat: HTTP endpoint to update account info (admin only)

bstrie:


crates/interledger-api/src/routes/accounts.rs
```diff
@@ -199,10 +199,29 @@ impl_web! {
                 )
         }
 
+        #[put("/accounts/:id")]
+        #[content_type("application/json")]
+        fn http_put_account(&self, id: String, body: AccountDetails, authorization: String) -> impl Future<Item = Value, Error = Response<()>> {
+            let parsed_id: Result<A::AccountId, ()> = A::AccountId::from_str(&id).map_err(|_| error!("Invalid id"));
+            self.validate_admin(authorization)
+                .and_then(move |store| match parsed_id {
```
emschwartz: I would replace this match with:

```
if let Ok(parsed_id) = A::AccountId::from_str(&id) {
...
} else {
error!("Invalid id: {}", id);
...
}
```

crates/interledger-store-redis/src/store.rs
```diff
@@ -286,8 +286,9 @@ impl RedisStore {
             .and_then(|(_conn, next_account_id): (_, u64)| Ok(next_account_id - 1))
     }
 
-    fn create_new_account(
+    fn redis_insert_account(
```
```suggestion
    fn redis_insert_or_update_account(
```

crates/interledger-store-redis/src/store.rs
```diff
@@ -309,106 +310,128 @@ impl RedisStore {
...
-
-                    if let Some(auth) = btp_incoming_token_hmac {
-                        keys.push("BTP auth".to_string());
-                        pipe.hexists("btp_auth", auth.as_ref());
-                    }
-                    if let Some(auth) = http_incoming_token_hmac {
-                        keys.push("HTTP auth".to_string());
-                        pipe.hexists("http_auth", auth.as_ref());
-                    }
+                    pipe.exists(accounts_key(id));
```
emschwartz: Since there's only one command, you can leave out the pipe and just use a single command, which will also mean you don't need to handle a `Vec<bool>` as the response type

crates/interledger-store-redis/src/store.rs
```diff
@@ -309,106 +310,128 @@ impl RedisStore {
...
-                                "Error checking whether account details already exist: {:?}",
-                                err
-                            )
+                        .map_err(|err| error!("Error checking whether ID exists: {:?}", err))
+                        .and_then(move |(connection, result): (SharedConnection, Vec<bool>)| {
+                            if let Some(_) = result.iter().position(|val| *val) {
+                                Account::try_from(id, account)
+                                    .and_then(move |account| Ok((connection, account)))
+                            } else {
+                                warn!("No account exists with ID {}, cannot insert account {:?}", id, account);
```
```suggestion
                                warn!("No account exists with ID {}, cannot update account {:?}", id, account);
```

crates/interledger-store-redis/src/store.rs
```diff
@@ -309,106 +310,128 @@ impl RedisStore {
...
+                        pipe.query_async(connection.as_ref().clone())
+                            .map_err(|err| {
+                                error!(
+                                    "Error checking whether account details already exist: {:?}",
+                                    err
+                                )
+                            })
+                            .and_then(
+                                move |(connection, results): (SharedConnection, Vec<bool>)| {
+                                    if let Some(index) = results.iter().position(|val| *val) {
```
emschwartz: This would be easier to read using the [`any`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) method

crates/interledger-store-redis/src/store.rs
```diff
@@ -309,106 +310,128 @@ impl RedisStore {
...
-            self.get_next_account_id()
-                .and_then(|id| {
-                    debug!("Next account id is: {}", id);
-                    Account::try_from(id, account)
-                })
-                .and_then(move |account| {
-                    // Check that there isn't already an account with values that must be unique
-                    let mut keys: Vec<String> = vec!["ID".to_string()];
-
+            match specified_id {
```
emschwartz: I find this function kind of confusing to read. Can you put comments explaining what each branch means and generally what it's doing?

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -45,6 +45,20 @@ fn delete_accounts() {
     .unwrap();
 }
 
+#[test]
```
emschwartz: Can you also add a test to make sure it doesn't overwrite the `balance` and `prepaid_amount`?

crates/interledger-store-redis/src/store.rs
bstrie: Ah, I was wondering how to control that behavior, I couldn't find that in the redis lib docs.

bstrie:
Comments addressed, need to rebase now.

bstrie:
Rebased atop #220, will be able to merge once that's in.

crates/interledger-store-redis/src/store.rs
```diff
@@ -311,107 +312,135 @@ impl RedisStore {
...
 
                     // Set account details
-                    pipe.cmd("HMSET").arg(account_details_key(account.id)).arg(account.clone().encrypt_tokens(&encryption_key))
+                    pipe.cmd("HMSET").arg(accounts_key(account.id)).arg(account.clone().encrypt_tokens(&encryption_key))
                         .ignore();
 
-                    // Set balance-related details
-                    pipe.hset_multiple(account_details_key(account.id), &[("balance", 0), ("prepaid_amount", 0)]).ignore();
+                // Set balance-related details iff the account doesn't already exist
+                if specified_id.is_none() {
```
emschwartz: I'm not a big fan of using this to determine whether or not to overwrite the balance. It's such a big footgun that I think I would prefer to separate the functionality for updating and adding accounts and just copy over the logic to do so (we can refactor it into a helper function that takes the `pipe`, though it's not that much logic so I'm not that worried about duplicating it.

gakonst:
#220 is merged, so could you please rebase to master? Thanks!

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -199,10 +199,30 @@ impl_web! {
                 )
         }
 
+        #[put("/accounts/:id")]
+        #[content_type("application/json")]
+        fn http_put_account(&self, id: String, body: AccountDetails, authorization: String) -> impl Future<Item = Value, Error = Response<()>> {
```
emschwartz: We may want to change this method in the future to allow creating accounts with a `PUT` instead of a `POST`, but it's fine like this for now

bstrie:
@emschwartz PR rebased to master and updated to separate out the insertion and updating logic into separate functions. Now that #220 has landed I wonder if I ought to fix those race TODOs I discovered before merging this?

emschwartz:
I think it might be better to merge this now and submit another PR after https://github.com/interledger-rs/interledger-rs/pull/233, since that will remove some of the checks that were needed before account insertion before.

bstrie:
Good point, and I won't mind not having to worry about rebasing this again. :P

PR #232, opened 2019-08-20
Fix example bug

dora-gt:
- Examples got broken because account specs have been changed.
- Settlement examples output adding account logs into files so that account id could be obtained later.
- This kind of problem could be avoided if once we add example tests.


examples/eth_xrp_three_nodes/README.md
```diff
@@ -187,7 +187,6 @@ export RUST_LOG=interledger=debug
 # Start Alice's settlement engine (ETH)
 cargo run --package interledger-settlement-engines -- ethereum-ledger \
 --key 380eb0f3d505f087e438eca80bc4df9a7faa24f868e69fc0440261a0fc0567dc \
---server_secret aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa \
```
dora-gt: It seems that this parameter was removed.

dora-gt:
I've checked both:

- scripts could be run by `run-md.sh`
- scripts could be run manually

So I'll merge this.

PR #233, opened 2019-08-20
Add Usernames and make authorization account-specific

gakonst:
Part of https://github.com/interledger-rs/interledger-rs/issues/186

- Adds usernames per account
- Adds a map from account id to btp/http token type
- Adds a map from username to account id
- Uses the username->account id map to make all user facing endpoints (SPSP/Balance) work with usernames, resulting in huge UX improvement
- Bearer auth methods are using an encoded token which includes both the username and the password, separated by a ':' delimiter. This was done to avoid changing any interfaces & minimize code changes.

Sometimes there's URL encoded params for ':', which we manually replace. We could add a url decoding dependency to do that more robustly perhaps.

BTP URIs now are passed as `username:token` instead of just `:token`.

The incoming token is set as `password_of_peer`.
The outgoing token is set as `username_of_our_account_on_peer:password_of_our_account_on_peer`.


crates/interledger-service/src/lib.rs
```diff
@@ -183,3 +183,13 @@ where
         Box::new((self.handler)(request).into_future())
     }
 }
+
+/// by default convert string usernames to u64 usernames
+/// (this is useful for tests because we cannot implement
+/// FromUsername for u64 since it's a foreign trait)
+/// For some reason #[cfg(test)] does not work here.
+impl FromUsername for u64 {
```
gakonst: Not sure how I feel about this, `#[cfg(test)]` should've worked here, but it seems like it's not made available to other crates.

crates/interledger-store-redis/tests/common/fixtures.rs
```diff
@@ -4,16 +4,18 @@ use lazy_static::lazy_static;
...
+        username: "alice".to_string(),
         asset_scale: 6,
         asset_code: "XYZ".to_string(),
         max_packet_amount: 1000,
         min_balance: Some(-1000),
         http_endpoint: Some("http://example.com/ilp".to_string()),
         http_incoming_token: Some("incoming_auth_token".to_string()),
-        http_outgoing_token: Some("outgoing_auth_token".to_string()),
-        btp_uri: Some("btp+ws://:btp_token@example.com/btp".to_string()),
+        http_outgoing_token: Some("dylan:outgoing_auth_token".to_string()),
```
emschwartz: Why does the account with the username "alice" use the username "dylan" on other nodes? That seems confusing

crates/interledger-service/src/lib.rs
emschwartz: Also, I've run into the `[cfg(test)]` thing before -- pretty sure that isn't exported outside of the crate

crates/interledger-store-redis/tests/common/fixtures.rs
gakonst: The outgoing token basically specifies the credentials your node will use on the other node. I agree that it can be confusing.

Dylan runs a node. Alice inserts Dylan in her store with username `dylan` and incoming http token `dylan_password`.

When Dylan inserts Alice into his connector, he must specify the account he will use when connecting with Alice. Previously, he'd only use `dylan_password`, but since the tokens are no longer globally unique, he must also specify his username. So it's going to be `dylan:dylan_password`.

crates/interledger-service/src/lib.rs
```diff
@@ -106,6 +107,14 @@ pub trait AccountStore {
         &self,
         account_ids: Vec<<<Self as AccountStore>::Account as Account>::AccountId>,
     ) -> Box<dyn Future<Item = Vec<Self::Account>, Error = ()> + Send>;
+
+    fn get_account_id_from_username(
```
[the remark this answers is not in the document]
gakonst: This method is no longer present

crates/interledger-service/src/lib.rs
gakonst: I do not, but tests use `u64` everywhere, so it `FromUsername` must be  implemented for u64. It cannot be implemented in tests sicne it's a foreign type-trait, so it's either:
1. Wrap all u64 accountIds in tests in a newtype and implement FromUsername there (lot of boilerplate copy paste code)
2. Do what I did here

crates/interledger-store-redis/src/store.rs
```diff
@@ -59,9 +59,13 @@ const DEFAULT_POLL_INTERVAL: u64 = 30000; // 30 seconds
 // trips for messages to be sent to and from Redis, as well as locks to ensure no other
 // process is accessing Redis at the same time.
 // For more information on scripting in Redis, see https://redis.io/commands/eval
-static ACCOUNT_FROM_INDEX: &str = "
-local id = redis.call('HGET', KEYS[1], ARGV[1])
-if not id then
+static ACCOUNT_FROM_TOKEN: &str = "
+local token_type = ARGV[1]
```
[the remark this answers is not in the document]
gakonst: What optionality are you referring to? There's no username being involved in the script, you pass it the id/token, it fetches the saved token for the id you provided, and if it matches the token you provided it returns the account. What's wrong with this approach?

crates/interledger-service/src/lib.rs
emschwartz: Why not just change the tests to use strings instead of `u64`s?

crates/interledger-settlement-engines/tests/test_helpers.rs
```diff
@@ -85,15 +84,15 @@ pub fn start_eth_engine(
...
     engine_port: u16,
-    account_id: T,
+    account_id: AccountId,
 ) -> impl Future<Item = String, Error = ()> {
     let client = reqwest::r#async::Client::new();
     client
         .post(&format!("http://localhost:{}/accounts", engine_port))
         .header("Content-Type", "application/json")
-        .json(&json!({ "id": account_id }))
+        .json(&json!({ "id": account_id.to_hex() }))
```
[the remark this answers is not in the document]
gakonst: Ids are `AccountId([u8; 32])`, we convert it to a hex-string from an array for the engine.

crates/interledger-store-redis/tests/common/fixtures.rs
emschwartz: Okay but it still doesn't make sense why the `username` specified in this account is `alice`

gakonst: `alice` is the account being inserted in Dylan's node in this example. So Dylan has to add the username / password of the account that he owns that he wants to use for transacting with Alice. The `outgoing_auth_token` is not a token that's related to `alice`'s informatino, it is a token that's used whenever Dylan wants to send something to Alice, so that he can authenticate to her node.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -141,30 +142,51 @@ impl_web! {
...
-                            .map_err(move |_| {
-                                debug!("Account not found: {}", id);
-                                Response::error(404)
-                            })
-                            .and_then(|mut accounts| Ok(json!(accounts.pop().unwrap()))))
-                    } else {
-                        Either::B(store.get_account_from_http_token(&authorization[BEARER_TOKEN_START..])
+            let username_clone = username.clone();
+            let auth_clone = authorization.clone();
+            store.get_account_id_from_username(username.clone())
```
emschwartz: It seems somewhat inconsistent that `get_account_from_http_token` would return the Account while this method only gives the ID (and then the next call is made to get the account)

One option would be to have two methods `get_account_from_username` and `get_account_from_id` (or just `get_account`) that would provide two ways to get the full `Account` details. If the next call only takes the account ID, that would be fine because you could just get that ID from the the `Account`

crates/interledger-api/src/routes/spsp.rs
```diff
@@ -65,33 +63,47 @@ impl_web! {
...
-            let id: Result<A::AccountId, ()> = A::AccountId::from_str(&id).map_err(|_| error!("Invalid id: {}", id));
-            result(id)
-                .map_err(|_| Response::builder().status(400).body(()).unwrap())
-                .and_then(move |id| store.get_accounts(vec![id])
+            store.get_account_id_from_username(username.clone())
+            .map_err(move |_| {
+                error!("Error getting account id from username: {}", username);
+                Response::builder().status(500).body(()).unwrap()
+            })
+            .and_then(move |id| store.get_accounts(vec![id])
```
emschwartz: As with above, it seems like we should just go from username -> `Account` and skip this `get_accounts` call

crates/interledger-btp/Cargo.toml
```diff
@@ -16,6 +16,7 @@ futures = "0.1.25"
 interledger-ildcp = { path = "../interledger-ildcp", version = "0.2.1" }
 interledger-packet = { path = "../interledger-packet", version = "0.2.1" }
 interledger-service = { path = "../interledger-service", version = "0.2.1" }
+interledger-http = { path = "../interledger-http", version = "0.2.1" }
```
emschwartz: This dependency seems wrong

crates/interledger-btp/src/server.rs
```diff
@@ -146,10 +146,12 @@ where
 
 struct Auth {
     request_id: u32,
-    username: Option<String>,
+    username: Option<String>, // This username is different from the one saved per account (?)
```
emschwartz: Huh?

crates/interledger-http/src/lib.rs
```diff
@@ -30,6 +31,99 @@ pub trait HttpStore: Clone + Send + Sync + 'static {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parses_correctly() {
+        assert_eq!(
+            Auth::parse("Basic ZXZhbjpzY2h3YXJ6").unwrap(),
+            Auth::new("evan", "schwarz")
```
emschwartz:
```suggestion
            Auth::new("evan", "schwartz")
```
Not a fan of using my name in the tests cause it makes it look like I'm super conceited unless you look at `git blame` to see who included it

crates/interledger-http/src/lib.rs
```diff
@@ -30,6 +31,99 @@ pub trait HttpStore: Clone + Send + Sync + 'static {
...
         token: &str,
     ) -> Box<dyn Future<Item = Self::Account, Error = ()> + Send>;
 }
+
+// Helpers for parsing authorization methods
+pub const BEARER_TOKEN_START: usize = 7;
+pub const BASIC_TOKEN_START: usize = 6;
+
+#[derive(Eq, PartialEq, Debug, Clone)]
+pub struct Auth {
```
emschwartz: Can you make sure the HTTP client is sending its auth details in a way the server will understand?

crates/interledger-store-memory/src/store.rs
```diff
@@ -124,13 +124,21 @@ impl AccountStore for InMemoryStore {
...
+        Box::new(ok(1))
+    }
 }
 
 impl HttpStore for InMemoryStore {
     type Account = Account;
 
     fn get_account_from_http_token(
         &self,
+        _username: &str, // What's going on here? InMemoryStore is very confusing.
```
emschwartz: Lol, wut?

crates/interledger-store-redis/src/store.rs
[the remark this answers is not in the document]
emschwartz: Got it. Can you add some comments either in the script or above it to make sure that's clear?

crates/interledger-store-redis/src/store.rs
```diff
@@ -296,37 +300,23 @@ impl RedisStore {
...
                 result(Account::try_from(id, account))
                 .and_then(move |account| {
-                    // Check that there isn't already an account with values that must be unique
-                    let mut keys: Vec<String> = vec!["ID".to_string()];
-
+                    // Check that there isn't already an account with values that MUST be unique
                     let mut pipe = redis::pipe();
                     pipe.exists(account_details_key(account.id));
+                    pipe.exists(format!("username:{}", account.username));
+                    pipe.hexists("btp_auth", account.id);
```
emschwartz: Shouldn't need this anymore

crates/interledger-store-redis/src/store.rs
```diff
@@ -296,37 +300,23 @@ impl RedisStore {
...
+                    pipe.hexists("http_auth", account.id);
```
emschwartz: Nor this

crates/interledger-store-redis/src/store.rs
```diff
@@ -360,12 +353,12 @@ impl RedisStore {
                     pipe.hset_multiple(account_details_key(account.id), &[("balance", 0), ("prepaid_amount", 0)]).ignore();
 
                     // Set incoming auth details
-                    if let Some(auth) = btp_incoming_token_hmac_clone {
-                        pipe.hset("btp_auth", auth.as_ref(), account.id).ignore();
+                    if let Some(auth) = btp_incoming_token {
+                        pipe.hset("btp_auth", account.id, auth).ignore();
```
emschwartz: Shouldn't need this

crates/interledger-store-redis/src/store.rs
```diff
@@ -360,12 +353,12 @@ impl RedisStore {
...
                     }
 
-                    if let Some(auth) = http_incoming_token_hmac_clone {
-                        pipe.hset("http_auth", auth.as_ref(), account.id).ignore();
+                    if let Some(auth) = http_incoming_token {
+                        pipe.hset("http_auth", account.id, auth).ignore();
```
emschwartz: Or this

emschwartz: Or this

crates/interledger-store-redis/src/store.rs
```diff
@@ -640,17 +654,20 @@ impl BtpStore for RedisStore {
 
     fn get_account_from_btp_token(
         &self,
+        username: &str,
         token: &str,
     ) -> Box<dyn Future<Item = Self::Account, Error = ()> + Send> {
         // TODO make sure it can't do script injection!
         // TODO cache the result so we don't hit redis for every packet (is that necessary if redis is often used as a cache?)
         let decryption_key = self.decryption_key.clone();
+        // reconstruct the token from the username to use the account id
```
emschwartz: What does that mean?

crates/interledger-store-redis/src/store.rs
```diff
@@ -718,19 +735,22 @@ impl BtpStore for RedisStore {
...
     type Account = Account;
 
+    /// Checks if the stored token for the provided account id matches the
+    /// provided token, and if so, returns the account associated with that token
     fn get_account_from_http_token(
         &self,
+        username: &str,
         token: &str,
     ) -> Box<dyn Future<Item = Self::Account, Error = ()> + Send> {
         // TODO make sure it can't do script injection!
```
emschwartz: Pretty important TODO :)

crates/interledger-store-redis/src/store.rs
gakonst: What do you think about how it's implemented now?

crates/interledger-http/src/lib.rs
gakonst: what do you think about https://github.com/interledger-rs/interledger-rs/pull/233/commits/b86864f64d34a9354fb785a11d3a67ae0865b468?

crates/interledger-api/src/routes/accounts.rs
gakonst: So in our call we said that we should keep using `id`s with any store related interfaces, which seems consistent with your comment here: https://github.com/interledger-rs/interledger-rs/issues/186#issuecomment-523597502. I'd rather not have to add an extra method in the interface, just to save this lookup.

crates/interledger-api/src/routes/spsp.rs
gakonst: https://github.com/interledger-rs/interledger-rs/pull/233#discussion_r316552808

crates/interledger-btp/Cargo.toml
gakonst: In #237 the Auth struct is refactored in interledger-service, so the http-import is removed from btp.

crates/interledger-store-redis/src/store.rs
gakonst: Fixed in #237

Issue #76, opened 2019-05-31
Use redis::Script to cache lua scripts in Redis


emschwartz:
For the Redis store, using `EVAL` with the full lua script causes that text to be sent to Redis each time. Instead of doing this, we can use the [`Script`](https://docs.rs/redis/0.11.0-beta.1/redis/struct.Script.html) struct to handle automatically caching the script (and calling it by its hash instead of the full text).

The main blocker to doing this before was the lack of support for asynchronously invoking `Script`s, but that was added in https://github.com/mitsuhiko/redis-rs/pull/206 and included in versions >=  0.11.0-beta.1.

PR #235, opened 2019-08-22
Only load script into Redis once

emschwartz:
Fixes https://github.com/interledger-rs/interledger-rs/issues/76

Depends on https://github.com/mitsuhiko/redis-rs/pull/233


gakonst:
Should we ask for a minor release of redis-rs, `0.11.1`, so that we update our deps with the new feature and merge this pr?

Issue #215, opened 2019-08-15
Add documentation for interledger node configuration


emschwartz:

dora-gt:
I think this is very important to newcomers and we should prioritize this.

emschwartz:
Closed by https://github.com/interledger-rs/interledger-rs/pull/290

Issue #206, opened 2019-08-12
Allow env vars to override config file values


inthecloud247:
Right now, I can set config values either by using the config file, or via env vars. For example:

```
ilp_address: "test.mynode.node1"
```

is equivalent to:

```
export ilp_ilp_address="test.mynode.node1"
```

However, when I set both values I get an error:

```
thread 'main' panicked at 'Must provide config file name or config environment variables: duplicate field `ilp_address`', src/libcore/result.rs:999:5
```

The expected behavior (imo) would be that env var values should override config file values. This would allow us to set sane default values in a config file, while also allowing new values to be set via env vars at runtime.

emschwartz:
There may be a very simple fix to this, but if not it may be worth fixing this when we switch config parsing libraries (see https://github.com/emschwartz/interledger-rs/issues/171)

emschwartz:
Closed by https://github.com/interledger-rs/interledger-rs/pull/290

Issue #194, opened 2019-08-07
Make settlement-engine to take environmental values


dora-gt:
While `node` takes environmental values, `ethereum-ledger` does not.
It might be better to take environmental values to have consistency.

dora-gt:
If @gakonst doesn't mind, I'd like to fix this just for my Rust study.

gakonst:
Sure.

dora-gt:
It seems that this issue is related to #169, #171, #206. These should probably be solved at the same time.

emschwartz:
Closed by https://github.com/interledger-rs/interledger-rs/pull/290

Issue #113, opened 2019-06-20
Fix --version flag in CLI


ekrenzke:
Currently, running the CLI with `-V` and `--version` returns `interledger` instead of the correct version number (0.4.0).

emschwartz:
:+1: We should also tag a new version some time soon

emschwartz:
Closed by https://github.com/interledger-rs/interledger-rs/pull/290

PR #236, opened 2019-08-22
CLI Improvement

### LINKED ISSUE #171: Use gumdrop or structopt for CLI option parsing
Instead of `clap`

Based on suggestion from @tarcieri in https://github.com/emschwartz/interledger-rs/issues/169

dora-gt:
- Fixes
    - #171 Use gumdrop or structopt for CLI option parsing
    - #113 Fix --version flag in CLI
    - #206 Allow env vars to override config file values
    - #194 Make settlement-engine to take environmental values
    - #215 Add documentation for interledger node configuration
- Version is obtained from crate's version
- ~~Using `structopt` to reduce the complexity of the source.~~
- ~~Using `#[allow(clippy::large_enum_variant)]` because options of account add command are too many. It may be difficult to fix this unless we make a patch to `structopt`~~
    - ~~There is someone [doing the same](https://github.com/sile/ekvsb/blob/master/src/main.rs#L35)~~

**BREAKING CHANGES**
- Ethereum settlement engine now doesn't accept `key` argument. Use `private_key` instead.
- ~~Ethereum settlement engine now doesn't accept `watch_incoming` flag because it was a flag and default_value was true, which doesn't make sense and `structopt` considers it as an error. Instead I implemented `without_incoming_watcher` flag which disables incoming watcher.~~
    - ~~`structopt`: `help: message: default_value is meaningless for bool`~~
- ~~`node` now doesn't accept `ilp_address` argument. Instead `address` argument or `ILP_ADDRESS` env var will be accepted.~~


crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,209 @@
...
+        };
+    }
+}
+
+impl EthereumLedgerOpt {
+    fn execute(&self) {
+        let mut config = config::Config::new();
+        config.merge(self.clone()).unwrap();
+        config
+            .merge(config::Environment::with_prefix("ILP"))
```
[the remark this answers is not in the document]
dora-gt: I thought `ILP` was required at least and adding `ETH_SE` or something would be too much (it becomes like `ILP_ETH_SE_ETHEREUM_ENDPOINT`), so... I think just `ILP` is enough🤔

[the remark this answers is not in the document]
dora-gt: I just left it as `ILP` but if there are more suitable options, it would be easy to fix.

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,209 @@
...
+        match &self.sub_command {
+            TopSubCommands::EthereumLedger(opt) => opt.execute(),
+        };
+    }
+}
+
+impl EthereumLedgerOpt {
+    fn execute(&self) {
+        let mut config = config::Config::new();
+        config.merge(self.clone()).unwrap();
```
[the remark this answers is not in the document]
dora-gt: Totally refactored.

crates/interledger-settlement-engines/tests/eth_ledger_settlement.rs
```diff
@@ -59,7 +59,7 @@ fn eth_ledger_settlement() {
 
     let node1_secret = cli::random_secret();
     let node1 = InterledgerNode {
-        ilp_address: Address::from_str("example.alice").unwrap(),
+        address: Address::from_str("example.alice").unwrap(),
```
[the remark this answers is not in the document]
dora-gt: The reason was: https://github.com/interledger-rs/interledger-rs/pull/236/files#diff-61630da1fc6519df646c48daf81debebR333-R337
but this was actually a bad idea. Fixed using another method.

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,209 @@
...
-                confirmations,
-                asset_scale,
-                poll_frequency,
-                connector_url,
-                token_address,
-                watch_incoming,
-            ));
-        }
-        _ => app.print_help().unwrap(),
+    fn collect(&self) -> Result<HashMap<String, Value, RandomState>, ConfigError> {
```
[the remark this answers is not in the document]
dora-gt: Totally refactored.

crates/interledger-settlement-engines/src/main.rs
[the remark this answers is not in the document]
dora-gt: Totally refactored. Now every command accepts config file as a positional argument.

crates/interledger/src/node.rs
```diff
@@ -83,10 +83,8 @@ where
 #[derive(Deserialize, Clone)]
 pub struct InterledgerNode {
     /// ILP address of the node
-    // Rename this one because the env vars are prefixed with "ILP_"
-    #[serde(alias = "address")]
```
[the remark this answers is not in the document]
dora-gt: https://github.com/interledger-rs/interledger-rs/pull/236#discussion_r318513974

examples/eth_xrp_three_nodes/README.md
```diff
@@ -193,7 +193,6 @@ cargo run --package interledger-settlement-engines -- ethereum-ledger \
 --connector_url http://127.0.0.1:7771 \
 --redis_uri redis://127.0.0.1:6379/0 \
 --asset_scale 6 \
---watch_incoming true \
```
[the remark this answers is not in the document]
dora-gt: The reason was written in the PR explanation but we no longer use `structopt` so this should be reverted now (actually, the default value is `true`).

image

dora-gt:
I fixed some of the issues but I still have to fix documents. I'm working on this.

docs/operating-manuals/eth-se-configurations.md
```diff
@@ -0,0 +1,81 @@
...
+There are some parameters which determine how your Ethereum settlement engine works. You could specify these parameters by the following methods.
+
+1. By command line arguments
+1. By environment variables
+
+```bash #
+# 1.
+# passing by command line arguments
+# --{parameter name} {value}
+cargo run --package interledger -- node --port 3000
```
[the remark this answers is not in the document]
dora-gt: 😨I'm sorry. I'm refactoring.

docs/operating-manuals/node-configurations.md
```diff
@@ -0,0 +1,75 @@
...
+There are some parameters which determine how your node works. You could specify these parameters by the following methods.
+
+1. By a configuration file
+    - [YAML](https://yaml.org/), [JSON](https://www.json.org/), [HJSON](https://hjson.org/), [TOML](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.5.0.md) and [INI](http://en.wikipedia.org/wiki/INI_file) format files will be accepted.
+1. By environment variables
+
+```bash #
+# 1.
+# passing by a configuration file in JSON, HJSON, TOML, YAML, or INI format
+cargo run --package interledger -- node --config config.yml
```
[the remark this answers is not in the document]
dora-gt: I tried but it seems to be an unstable version.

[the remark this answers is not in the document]
dora-gt: Oops, my environment was old.

docs/operating-manuals/README.md
```diff
@@ -0,0 +1,6 @@
+# Interledger.rs Operating Manuals
+
+## Initial Set up
+- [Node Configurations](node-configurations.md)
```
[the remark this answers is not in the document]
dora-gt: Totally refactored.

docs/operating-manuals/eth-se-configurations.md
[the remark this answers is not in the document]
dora-gt: Totally refactored.

docs/operating-manuals/eth-se-configurations.md
```diff
@@ -0,0 +1,81 @@
...
+### Parameters
+
+The parameters are shown in the following format.
+
+- `parameter name`
+    - format
+        - example
+    - explanation
+
+#### Required
```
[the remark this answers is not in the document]
dora-gt: Totally refactored. Now we don't have explanations of each command. Instead, `--help` explanation is given.

docs/operating-manuals/node-configurations.md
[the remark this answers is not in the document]
dora-gt: Fixed.

dora-gt:
I think I addressed issues, if you have anything else, please tell me.

dora-gt:
> configuration files should be passed in with node config.yml instead of node --config config.yml

Currently it accepts like this.

> we should try to use the required flag on Clap so that the --help prints correctly

I've tried several hours to hack this... but I couldn't come up with how to cleanly and entirely realize this... Anyway, let me try some more.

dora-gt:
The problem is that if we set `required` flag, it is parsed BEFORE parameters from configuration files are given. So even if we set a configuration file, it warns that some required parameters are not set. To prevent this,

- We might need to hack `help`
    - Negative, there are so many points that generate `help` strings, we cannot hack this entirely.
- We might need to hack `env` or `get_matches_from`
    - Negative, we might need to parse twice or need to parse on our own, which looks a bit ugly.
    - That said, if we have any solution, I think this approach is closer to it.

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,197 @@
...
-                            .default_value("5000"),
-                        Arg::with_name("watch_incoming")
-                            .long("watch_incoming")
-                            .help("Launch a blockchain watcher that listens for incoming transactions and notifies the connector upon sufficient confirmations")
-                            .default_value("true"),
-                    ])
-        ]
-    );
+                .args(&[
+                    Arg::with_name("config")
```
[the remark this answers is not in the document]
dora-gt: It actually accepts positional arg of config file though there were some confusing settings like `long` or `short`. I removed these.

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,197 @@
...
+            let config = get_config(ethereum_ledger_matches, "ILP");
+            get_or_error(config.try_into::<EthereumLedgerOpt>()).execute();
+        }
+        ("", None) => app.print_help().unwrap(),
+        _ => unreachable!(),
+    }
+}
+
+impl EthereumLedgerOpt {
+    fn execute(&self) {
```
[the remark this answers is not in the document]
dora-gt: Totally refactored. Hope it seems cleaner.

crates/interledger/src/main.rs
```diff
@@ -20,15 +20,25 @@ pub fn main() {
...
+        .setting(AppSettings::SubcommandsNegateReqs)
+        .setting(AppSettings::ArgsNegateSubcommands)
         .subcommands(vec![
             SubCommand::with_name("spsp")
                 .about("Client and Server for the Simple Payment Setup Protocol (SPSP)")
                 .subcommands(vec![
                     SubCommand::with_name("server")
                         .about("Run an SPSP Server that automatically accepts incoming money")
                         .args(&[
+                            Arg::with_name("config")
```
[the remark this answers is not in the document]
dora-gt: Now configs are all positional.

crates/interledger/src/main.rs
```diff
@@ -70,73 +86,121 @@ pub fn main() {
                                 .long("receiver")
                                 .short("r")
                                 .takes_value(true)
-                                .required(true)
-                                .help("Payment Pointer of the receiver"),
+                                .help("(REQUIRED) Payment Pointer of the receiver"),
```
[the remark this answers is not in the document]
dora-gt: I did some hack to solve 3 ways problem (arg, config, env). Now it works.

- Even if no args are given, config file or envs are used to parse arguments.
    - Required fields still work with it.
- Prioritizes in order of `env vars` -> `config file settings` -> `args`

crates/interledger/src/main.rs
```diff
@@ -187,181 +251,284 @@ pub fn main() {
...
+    btp_server: String,
+    #[serde(default = "default_as_false")]
+    ilp_over_http: bool,
+    ilp_address: String,
+    incoming_auth_token: String,
+    #[serde(default = "default_as_false")]
+    quiet: bool,
+}
+
+impl SpspServerOpt {
```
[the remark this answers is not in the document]
dora-gt: Totally refactored. I think now it is not confusing as it is actually options.

dora-gt:
- `Config` is a positional arg.
- Required fields are shown in the help.
- Warns if the required fields are not given.

image

image

image

dora-gt:
oops... I found a bug😨 I'm fixing.

crates/interledger/src/main.rs
```diff
@@ -21,15 +22,24 @@ pub fn main() {
         random_token(),
         random_token()
     );
+
     let mut app = App::new("interledger")
         .about("Blazing fast Interledger CLI written in Rust")
+        .version(crate_version!())
+        .setting(AppSettings::SubcommandsNegateReqs)
+        .after_help("")
```
emschwartz: I don't think this does anything

crates/interledger/src/main.rs
```diff
@@ -84,84 +100,143 @@ pub fn main() {
...
+                    Arg::with_name("ilp_address")
+                        .long("ilp_address")
+                        .takes_value(true)
+                        .required(true)
+                        .help("ILP Address of this account"),
+                    Arg::with_name("secret_seed")
+                        .long("secret_seed")
+                        .takes_value(true)
+                        .required(true)
+                        .help("Root secret used to derive encryption keys. This MUST NOT be changed after once you started up the node. You could obtain randomly generated one using `openssl rand -hex 32`"),
```
emschwartz:
```suggestion
                        .help("Root secret used to derive encryption keys. This MUST NOT be changed after once you started up the node. You can generate a random secret by running `openssl rand -hex 32`"),
```

crates/interledger/src/main.rs
```diff
@@ -84,84 +100,143 @@ pub fn main() {
...
+                    Arg::with_name("admin_auth_token")
+                        .long("admin_auth_token")
+                        .takes_value(true)
+                        .required(true)
+                        .help("HTTP Authorization token for the node admin (sent as a Bearer token). Refer to: https://bit.ly/2Lk4xgF https://bit.ly/2XaUiBb"),
```
emschwartz:
```suggestion
                        .help("HTTP Authorization token for the node admin (sent as a Bearer token)")
```

crates/interledger/src/main.rs
```diff
@@ -170,203 +245,372 @@ pub fn main() {
...
-                    };
-                    tokio::run(
-                        insert_account_redis(redis_uri, &server_secret, account)
-                            .and_then(move |_| Ok(())),
-                    );
+        ("node", Some(node_matches)) => match node_matches.subcommand() {
+            ("accounts", Some(node_accounts_matches)) => match node_accounts_matches.subcommand() {
+                ("add", Some(node_accounts_add_matches)) => {
+                    merge_args(&mut config, &node_accounts_add_matches);
+                    runner.run(get_or_error(config.try_into::<NodeAccountsAddOpt>()));
```
emschwartz: I think this code would be clearer without the `Runner` thing. You could just copy that code into here, since it's not actually abstracting anything away

crates/interledger/tests/three_nodes.rs
```diff
@@ -47,7 +47,7 @@ fn three_nodes() {
 
     let node1 = InterledgerNode {
         ilp_address: Address::from_str("example.one").unwrap(),
-        default_spsp_account: Some(0),
+        default_spsp_account: None,
```
emschwartz: Shouldn't this be set to one of the accounts?

docs/operating-manuals/README.md
```diff
@@ -0,0 +1,70 @@
...
+ILP_ADDRESS=example.alice \
+ILP_OTHER_PARAMETER=other_value \
+cargo run -- node
+```
+
+The commands prioritize parameters in the following order (high to low).
+
+- Environment variables
+- Configuration files
+- Command line arguments
```
emschwartz: It's slightly confusing that this list is in the opposite order as the one above. I would just write this list once and clarify there what the prioritization is

crates/interledger/tests/three_nodes.rs
gakonst: No need, all tests explicitly declare the accounts they're sending to.

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,266 @@
...
+        _ => unreachable!(),
+    }
+}
+
+fn merge_config_file(mut app: App, config: &mut Config) -> Result<Vec<String>, ()> {
+    // not to cause `required fields error`.
+    reset_required(&mut app);
+    let matches = app.get_matches_safe();
+    if matches.is_err() {
+        // if app could not get any appropriate match, just return not to show help etc.
```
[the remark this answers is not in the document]
dora-gt: In this section, I don't want to show the help text, because this is just for obtain the `config` argument.

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,266 @@
...
+    }
+    let matches = &matches.unwrap();
+    let mut path = Vec::<String>::new();
+    let subcommand = get_deepest_command(matches, &mut path);
+    if let Some(config_path) = subcommand.value_of("config") {
+        let file_config = config::File::with_name(config_path);
+        let file_config = file_config.collect().unwrap();
+
+        // if the key is not defined in the given config already, set it to the config
+        // because the original values override the ones from the config file
```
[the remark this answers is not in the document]
dora-gt: I don't see the clear difference because, in either case, we can override it by environment variables. hmmm do you have a preference?

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,266 @@
...
+            config.set("ilp_address", value).unwrap();
+        }
+    }
+
+    config
+}
+
+// sets env value into each optional value
+// only applied to the specified last command
+fn set_app_env(env_config: &Config, app: &mut App, path: &[String], depth: usize) {
```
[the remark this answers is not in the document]
dora-gt: In short, this is to avoid required field error.

This sets env vars and config file settings into `item.v.env`, which is used when `Parser` parses the arguments and if this value is set, the `Parser` reads the value from it and doesn't warn if the argument is not given.

crates/interledger/src/main.rs
dora-gt: It actually does... I also feel this is ugly but if I don't do this, it sometimes outputs like:

image

I don't know why...😔

crates/interledger/src/main.rs
dora-gt: I wanted to make it readable. In the original code, there were many lines between subcommand matches, so it looked hard to read for me. I think this is clearer than inserting the code here because we can understand it intuitively as:

- Here is to parse subcommands
- There are many subcommands and some of these also have its own subcommands
- Subcommands execute some logic using options

I think too long code is hard to understand. Isn't breaking the responsibility into some parts a good method? At least, I've believed so.

dora-gt: ummm but it actually doesn't abstract. Don't you also want to make it just a `fn`?

crates/interledger/tests/three_nodes.rs
dora-gt: Thank you.

crates/interledger/src/main.rs
emschwartz: Oh you mean it doesn't print a newline after?

gakonst:
Feature request from: https://github.com/interledger-rs/interledger-rs/issues/108#issuecomment-526541799

We want to add an option to load config files from stdin, allowing a user to do `cat alice.yaml | interledger node`. The use case for this is that we're going to have encrypted config files, and the user will run `sops -d alice.yaml | interledger node`.

crates/interledger/src/main.rs
dora-gt: Yes... I tried some, such as moving `help` fn after the other options but it didn't address the problem. I couldn't come up with the cause.

dora-gt:
Which format would you like to pass to STDIN?

I'm asking because unfortunately `Config` file module was private and I can't reuse the code. I have to write some glue code to parse it though it may be possible to copy as it is licensed under  `MIT/Apache 2.0`.

gakonst:
> Which format would you like to pass to STDIN?

So the config module understands which deserializer to use based on the config file extension (`.yaml`, `.json`, etc.). Us passing it from STDIN means that the config module cannot parse it. We could manually import `serde_yaml` and `serde_json`, and try all until one works.

dora-gt:
Since our CLI supports YAML, TOML, JSON, etc... thanks to `Config` module, it seems natural that we support these format for STDIN.

Ok, I'll try.

dora-gt:
@gakonst it might not look so clean but I tried to reuse the code from `Config` as much as possible. It uses many crates but these are actually used already, so it adds nothing.

How does it look? 

https://github.com/interledger-rs/interledger-rs/commit/6bcf2bd6f86e2956f9a6b67172d9462e8c13dbbc

dora-gt:
Oops... I forgot to explain this in the document. I'm sorry I'll fix tomorrow.

crates/interledger-settlement-engines/Cargo.toml
```diff
@@ -37,6 +37,12 @@ sha3 = "0.8.2"
 num-bigint = "0.2.2"
 num-traits = "0.2.8"
 lazy_static = "1.3.0"
+config = "0.9.3"
+rust-ini = "0.13"
```
emschwartz: Instead of separately including the dependencies, we can just enable the features on `config` (see https://github.com/mehcode/config-rs#usage)

Also, I'm not sure we really need to support INI or HJSON files. Yaml, JSON, and maybe TOML seem useful

crates/interledger-settlement-engines/Cargo.toml
```diff
@@ -37,6 +37,12 @@ sha3 = "0.8.2"
...
+serde-hjson = "0.8.2"
+toml = "0.4.1"
+yaml-rust = "0.4"
+libc = "0.2.62"
```
emschwartz: What is this needed for?

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,313 @@
...
+                        .help("Name of config file (in JSON, HJSON, TOML, YAML, or INI format)"),
+                    Arg::with_name("stdin")
+                        .long("stdin")
+                        .help("If given, read config from stdin (in JSON, HJSON, TOML, YAML, or INI format)"),
+                    Arg::with_name("http_address")
+                        .long("http_address")
+                        .takes_value(true)
+                        .default_value("127.0.0.1:3000")
+                        .help("Port to listen for settlement requests on"),
+                    Arg::with_name("key")
```
emschwartz: While we're breaking the configuration, I think we might want to change the name of this property to something like `secret_key` (what is it called by the ethereum components?)

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,313 @@
...
+        return Err(());
+    }
+    let matches = &matches.unwrap();
+    let mut path = Vec::<String>::new();
+    let subcommand = get_deepest_command(matches, &mut path);
+    let mut config_path: Option<String> = None;
+    if let Some(config_path_arg) = subcommand.value_of("config") {
+        config_path = Some(config_path_arg.to_string());
+    };
+    let stdin_flag = subcommand.is_present("stdin");
```
emschwartz: Is there a benefit to requiring a flag to make this work versus always accepting whatever is piped to `stdin`? (Honest question, I'm not sure and there totally could be a benefit to being more explicit)

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,313 @@
...
+}
+
+fn merge_std_in(config: &mut Config) {
+    let stdin = std::io::stdin();
+    let mut stdin_lock = stdin.lock();
+    let mut buf = Vec::new();
+    if let Ok(_read) = stdin_lock.read_to_end(&mut buf) {
+        if let Ok(buf_str) = String::from_utf8(buf) {
+            let format: FileFormat;
+            if yaml_rust::YamlLoader::load_from_str(&buf_str).is_ok() {
```
emschwartz: Instead of parsing the file twice, you can make the `if else` block return `Option<ConfigHash>` or something like that

crates/interledger-settlement-engines/src/main.rs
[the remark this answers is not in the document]
emschwartz: It would be good to describe this fully in the comments

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,313 @@
...
+            Some(EthAddress::from_str(&token_address).unwrap())
+        } else {
+            None
+        };
+        let redis_uri = Url::parse(&opt.redis_uri).expect("redis_uri is not a valid URI");
+
+        // TODO make key compatible with
+        // https://github.com/tendermint/signatory to have HSM sigs
+
+        tokio::run(run_ethereum_engine(
```
emschwartz: Why doesn't `run_ethereum_engine` just take this option struct?

crates/interledger-settlement-engines/src/main.rs
gakonst: The ethereum engine has a `Signer` trait which is currently implemented for `Secret<String>`. In this case, `secret_key` or `private_key` should be fine.

crates/interledger-settlement-engines/src/main.rs
gakonst: https://github.com/prettier/prettier/issues/4559 similar discussion, didn't seem necessary

crates/interledger-settlement-engines/Cargo.toml
dora-gt: It is required for these lines: https://github.com/interledger-rs/interledger-rs/blob/3002b221f09799e6b2b9f6fd07b922618ae3db57/crates/interledger/src/main.rs#L375-L384

It seems not so good but I had to check what is the format of STDIN. Should we specify like `--stdin json` instead? 🤔

Anyway, indeed we might not need to support `HJSON` and `INI`. According to the above, I'll fix.

crates/interledger-settlement-engines/Cargo.toml
dora-gt: Ooops... I'm sorry I forgot to remove this. I was looking into how to detect STDIN automatically. https://github.com/dora-gt/exploring_libc/tree/master/crates/fstat

I'll remove.

crates/interledger-settlement-engines/src/main.rs
dora-gt: umm just checked in `go-ethereum`,

```
$ ag private[-_]key | wc -l
19

$ ag secret[-_]key | wc -l
0
```

Change to `private_key` maybe?

crates/interledger-settlement-engines/src/main.rs
dora-gt: Actually I was checking the method to read STDIN **only if it is given** (without blocking).

I found 2 options.

1. Check whether STDIN is not pointed to TTY
1. Check whether STDIN is pointed to FIFO (anonymous or named)
    - https://github.com/dora-gt/exploring_libc/tree/master/crates/fstat

It almost works but I was not confident this would work perfectly. I think option 2 is more proper because `STDIN != TTY` contains many options and `STDIN == FIFO` looks a bit clearer.

Also, there is a benefit of specifying `--stdin` if we set it to take values like `--stdin json` which is more efficient internally.

What do you think?

crates/interledger-settlement-engines/src/main.rs
dora-gt: The reason why I wrote like this is that I had to write duplicated code like:
https://github.com/mehcode/config-rs/blob/master/src/file/format/json.rs#L22-L55
https://github.com/mehcode/config-rs/blob/master/src/file/format/yaml.rs#L33-L66

Hmmm but actually it is ugly and not efficient. I'll fix.

crates/interledger-settlement-engines/src/main.rs
dora-gt: Indeed, this is weird😨 I'll fix.

crates/interledger-settlement-engines/src/main.rs
dora-gt: While `yaml_rust::YamlLoader::load_from_str` returns it's own format like `Vec<Yaml>`, `config::file::format::yaml::from_yaml_value` is private. So I had to write duplicated code like `from_yaml_value`.

I thought it might be better to reuse code but it actually inefficient. I'll fix.

dora-gt: I'm sorry I could understand. I was completely wrong.

dora-gt: Fixed.

crates/interledger-settlement-engines/src/main.rs
dora-gt: Fixed.

crates/interledger-settlement-engines/src/main.rs
dora-gt: I thought about this, and maybe... using `STDIN != TTY` is better. Let me explain though it would be long.

I tried or thought about these options.

1. Always read from STDIN
    - Not good because it could block when it is run just in a terminal (=TTY).
    - Non-blocking IO is also not good, the reason is below.
1. Read when STDIN is readable (using `poll`, checking the IO is ready)
    - Not good because the reason why the IO is not ready might be that the writer is still processing and busy.
    - Imagine this case: `cmd_a | cmd_b`. `cmd_a` may require a few seconds to write, and then `cmd_b` execute `poll`, it recognizes the STDIN is not ready.
1. Read when STDIN is FIFO (using `fstat`)
    - Maybe not the best. Though it works in most cases, when STDIN is redirected from a file like: `cmd < config`, it doesn't work.
    - Then `STDIN is FIFO || STDIN is FILE` looks good but it almost looks like `STDIN != TTY`. -> Why don't we use `STDIN != TTY`?
1. Read when STDIN is NOT TTY (using `isatty`)
    - Though it requires parse tries to detect the format, it seems to be able to handle most cases.
    - Actually `--stdin {format}` has a benefit of efficiency, parsing a few times may not require so much time as follows.

```
time cat conf/ini.ini | cargo run --all-features -- node --stdin
real	0m0.274s
user	0m0.192s
sys	0m0.077s
```

It's only 0.x sec. Maybe no problem.

So, I'll change this to use `isatty` not using `--stdin`.

crates/interledger-settlement-engines/Cargo.toml
dora-gt: Now `INI` and `HJSON` are not supported by default.
Features of `ini` and `hjson` are available.

crates/interledger-settlement-engines/src/main.rs
dora-gt: Refactored: https://github.com/interledger-rs/interledger-rs/pull/236/commits/b19099a0fa5574ef4d5244d6d0e74a77234db504

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,303 @@
...
+                    Arg::with_name("config")
+                        .takes_value(true)
+                        .index(1)
+                        .help(&CONFIG_HELP),
+                    Arg::with_name("http_address")
+                        .long("http_address")
+                        .takes_value(true)
+                        .default_value("127.0.0.1:3000")
+                        .help("Port to listen for settlement requests on"),
+                    Arg::with_name("key")
```
gakonst:
```suggestion
                    Arg::with_name("private_key")
```

gakonst:
FYI: I've tested encrypting a YAML configuration file with `sops` and then running a node with `sops -d alice.yaml | cargo run -- node` and it's working well.

dora-gt:
Thank you @gakonst , unfortunately I found a bug. If it is run on Docker, it tries to read from STDIN which causes an IO-block. It is because that `isatty` is `false` on Docker even when it is not piped.

I'm trying to fix this.

dora-gt:
^ was not a problem, just using `-t` option for Docker.

crates/interledger/src/main.rs
```diff
@@ -170,203 +253,467 @@ pub fn main() {
...
+    let (name, subcommand_matches) = matches.subcommand();
+    path.push(name.to_string());
+    if let Some(matches) = subcommand_matches {
+        return get_deepest_command(matches, path);
+    }
+    matches
+}
+
+fn reset_required(app: &mut App) {
+    app.p.required.clear();
```
emschwartz: If we reset which are required, what will happen if you fail to pass a required parameter?

docs/operating-manuals/README.md
```diff
@@ -0,0 +1,71 @@
...
+cargo run -- node --help
+```
+
+### Specifying Parameters
+
+Interledger.rs commands such as `node` and `ethereum-ledger` accept configuration options in the following ways:
+
+1. Command line arguments
+1. Configuration files
+1. STDIN
```
emschwartz:
```suggestion
1. Standard In (stdin)
```

dora-gt:
Thank you Evan for taking time, I'll check and fix.

crates/interledger-settlement-engines/Cargo.toml
```diff
@@ -17,6 +23,7 @@ interledger-service = { path = "../interledger-service", version = "0.2.1" }
 interledger-settlement = { path = "../interledger-settlement", version = "0.1.0" }
 interledger-service-util = { path = "../interledger-service-util", version = "0.2.1" }
 interledger-ildcp = { path = "../interledger-ildcp", version = "0.2.1" }
+interledger-store-redis = { path = "../interledger-store-redis", version = "0.2.1" }
```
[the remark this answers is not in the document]
dora-gt: Ah I should've noticed that `ConnectionInfo` is finally pointed to `redis`'s `ConnectionInfo`. 🙇

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -983,50 +986,80 @@ fn prefixed_mesage(challenge: Vec<u8>) -> Vec<u8> {
...
             Ok(())
         })
 }
 
+#[derive(Deserialize, Clone)]
+pub struct EthereumLedgerOpt {
+    pub private_key: String,
+    pub http_address: SocketAddr,
+    pub ethereum_endpoint: String,
+    pub token_address: Option<String>,
```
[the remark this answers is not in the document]
dora-gt: Cool, I'll check.

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,303 @@
...
+                        .long("http_address")
+                        .takes_value(true)
+                        .default_value("127.0.0.1:3000")
+                        .help("Port to listen for settlement requests on"),
+                    Arg::with_name("private_key")
+                        .long("private_key")
+                        .takes_value(true)
+                        .required(true)
+                        .help("private key for settlement account"),
+                    Arg::with_name("ethereum_endpoint")
```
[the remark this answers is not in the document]
dora-gt: Agree, it becomes more consistent.

Honestly, I think there are some more arguments that might need to be changed, I would list...

- `redis_uri` -> `redis_url` I think both are technically correct but `redis` module uses `url` internally like `parse_redis_url` so I think `url` is better
- `redis_connection` -> `redis_url` there are 2 expressions, which looks not good
- `port` -> `xxx_address` because specifying only a port is not sufficient
- `btp_server` -> `btp_server_url`
- `http_server` -> `http_server_url`
- `btp_uri` -> `btp_server_url`

What about these?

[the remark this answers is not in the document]
dora-gt: or... `port` and `xxx_address` could be exclusive arguments. If only a port is given, it is considered `127.0.0.1:port`, which looks a bit useful.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -983,50 +986,80 @@ fn prefixed_mesage(challenge: Vec<u8>) -> Vec<u8> {
...
-    R: IntoConnectionInfo,
-    Si: EthereumLedgerTxSigner + Clone + Send + Sync + 'static,
-{
-    let redis_uri = redis_uri.into_connection_info().unwrap();
+pub fn run_ethereum_engine(opt: EthereumLedgerOpt) -> impl Future<Item = (), Error = ()> {
+    let token_address = if let Some(token_address) = &opt.token_address {
+        if token_address.len() == 20 {
+            Some(EthAddress::from_str(token_address).unwrap())
+        } else {
+            None
```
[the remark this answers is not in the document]
dora-gt: Indeed. I'll fix.

crates/interledger-settlement-engines/Cargo.toml
```diff
@@ -37,6 +44,12 @@ sha3 = "0.8.2"
 num-bigint = "0.2.2"
 num-traits = "0.2.8"
 lazy_static = "1.3.0"
+config = "0.9.3"
+libc = "0.2.62"
+rust-ini = { version = "0.13", optional = true }
```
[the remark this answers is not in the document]
dora-gt: My bad... When I refactored, I should've refactored these also.

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,303 @@
...
+    }
+}
+
+fn merge_std_in(config: &mut Config) {
+    let stdin = std::io::stdin();
+    let mut stdin_lock = stdin.lock();
+    let mut buf = Vec::new();
+    if let Ok(_read) = stdin_lock.read_to_end(&mut buf) {
+        if let Ok(buf_str) = String::from_utf8(buf) {
+            let mut config_hash = None;
```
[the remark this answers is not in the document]
dora-gt: Clean!

crates/interledger/src/main.rs
dora-gt: It's still required when we parse the arguments actually because we are resetting an instance which was cloned (though this is not so good). So, it will warn if we lack some required parameters as follows:

image

image

crates/interledger/src/main.rs
```diff
@@ -84,84 +108,143 @@ pub fn main() {
...
-                    .subcommand(SubCommand::with_name("accounts")
-                        .subcommand(SubCommand::with_name("add")
+                        .index(1)
+                        .help(&CONFIG_HELP),
+                    Arg::with_name("ilp_address")
+                        .long("ilp_address")
+                        .takes_value(true)
+                        .required(true)
+                        .help("ILP Address of this account"),
+                    Arg::with_name("secret_seed")
```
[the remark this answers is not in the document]
dora-gt: Isn't it already `secret_seed`? Do you mean `secret_key` or something? Sorry if I'm missing something.

crates/interledger-settlement-engines/Cargo.toml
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger-settlement-engines/Cargo.toml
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -1,24 +1,27 @@
...
 use super::utils::{filter_transfer_logs, make_tx, sent_to_us, ERC20Transfer};
+use super::EthAddress;
 use clarity::Signature;
 use log::{debug, error, trace};
 use sha3::{Digest, Keccak256 as Sha3};
 use std::collections::HashMap;
 use std::iter::FromIterator;
 
 use hyper::StatusCode;
+use interledger_store_redis::ConnectionInfo;
```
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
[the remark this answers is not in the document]
dora-gt: Merged into: https://github.com/interledger-rs/interledger-rs/pull/236#discussion_r321404072

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,303 @@
...
+                    Arg::with_name("redis_uri")
+                        .long("redis_uri")
+                        .takes_value(true)
+                        .default_value("redis://127.0.0.1:6379")
+                        .help("Redis database to add the account to"),
+                    Arg::with_name("chain_id")
+                        .long("chain_id")
+                        .takes_value(true)
+                        .default_value("1")
+                        .help("The chain id so that the signer calculates the v value of the sig appropriately"),
```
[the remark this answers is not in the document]
dora-gt: Added comment.

crates/interledger-settlement-engines/Cargo.toml
```diff
@@ -4,6 +4,12 @@ version = "0.1.0"
 authors = ["Georgios Konstantopoulos <me@gakonst.com>"]
 edition = "2018"
 
+[features]
```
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger-settlement-engines/src/main.rs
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,303 @@
...
+    let mut formats = Vec::new();
+    // JSON is always supported because the crate is used already
+    formats.push("JSON");
+    if cfg!(feature = "yaml") {
+        formats.push("YAML");
+    }
+    if cfg!(feature = "toml") {
+        formats.push("TOML");
+    }
+    if cfg!(feature = "hjson") {
```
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger/Cargo.toml
```diff
@@ -6,13 +6,14 @@ description = "Interledger client library"
...
 repository = "https://github.com/interledger-rs/interledger-rs"
+default-run = "interledger"
 
 [lib]
 name = "interledger"
 path = "src/lib.rs"
 
 [features]
-default = ["cli"]
+default = ["cli", "yaml", "json", "toml"]
```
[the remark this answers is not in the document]
dora-gt: Fixed.

dora-gt:
Fixed most points, I'm wondering just: https://github.com/interledger-rs/interledger-rs/pull/236#discussion_r321547737

crates/interledger-settlement-engines/src/main.rs
```diff
@@ -1,106 +1,303 @@
...
-                token_address,
-                watch_incoming,
-            ));
+            std::process::exit(1);
         }
-        _ => app.print_help().unwrap(),
     }
 }
+
+fn is_fd_tty(file_descriptor: c_int) -> bool {
```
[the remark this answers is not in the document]
dora-gt: Added comment.

crates/interledger/src/main.rs
emschwartz: Yeah looks good

crates/interledger/src/main.rs
```diff
@@ -84,84 +105,143 @@ pub fn main() {
...
+                    Arg::with_name("redis_connection")
+                        .long("redis_connection")
+                        .takes_value(true)
+                        .default_value("redis://127.0.0.1:6379")
+                        .help("Redis URI (for example, \"redis://127.0.0.1:6379\" or \"unix:/tmp/redis.sock\")"),
+                    Arg::with_name("http_address")
+                        .long("http_address")
+                        .takes_value(true)
+                        .help("IP address and port to listen for HTTP connections. This is used for both the API and ILP over HTTP packets. ILP over HTTP is a means to transfer ILP packets instead of BTP connections"),
+                    Arg::with_name("settlement_address")
```
emschwartz:
```suggestion
                    Arg::with_name("settlement_api_bind_address")
```

crates/interledger/src/main.rs
```diff
@@ -84,84 +105,143 @@ pub fn main() {
...
+                        .long("settlement_address")
                         .takes_value(true)
-                        .help("Name of config file (in JSON, TOML, YAML, or INI format)"))
-                    .subcommand(SubCommand::with_name("accounts")
-                        .subcommand(SubCommand::with_name("add")
+                        .help("IP address and port to listen for the Settlement Engine API"),
+                    Arg::with_name("btp_address")
```
emschwartz:
```suggestion
                    Arg::with_name("btp_bind_address")
```

crates/interledger/src/main.rs
```diff
@@ -170,203 +250,400 @@ pub fn main() {
...
+}
+
+fn reset_required(app: &mut App) {
+    app.p.required.clear();
+    for subcommand in &mut app.p.subcommands {
+        reset_required(subcommand);
+    }
+}
+
+fn get_or_error<T>(item: Result<T, ConfigError>) -> T {
```
emschwartz: nit: couldn't we just use `unwrap` or `expect` on the `Result` above (instead of having this separate function)

crates/interledger/src/main.rs
```diff
@@ -170,203 +250,400 @@ pub fn main() {
...
+    #[serde(default = "default_as_false")]
+    receive_routes: bool,
+    routing_relation: String,
+    min_balance: i64,
+    round_trip_time: u32,
+    packets_per_minute_limit: Option<u32>,
+    amount_per_minute_limit: Option<u64>,
+}
+
+fn default_as_false() -> bool {
```
emschwartz: Not needed because booleans default to false

docs/operating-manuals/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+```bash #
+# 1.
+# Passing by command line arguments.
+# --{parameter name} {value}
+cargo run -- node --ilp_address example.alice
+
+# 2.
+# Passing by a configuration file in JSON, TOML, YAML format.
+# To support HJSON and INI format, you need to turn on `hjson` or `ini` feature when you compile.
```
emschwartz:
```suggestion
```

crates/interledger/src/main.rs
dora-gt: If I don't specify this, it causes an error of ``missing field `ilp_over_http` `` but it seems that specifying `.default_value("false")` to the argument solves it though it looks a bit weird because it is a flag and it doesn't take a value.

crates/interledger/src/main.rs
dora-gt: What I wanted to avoid here was the outputs like

```
thread 'main' panicked at 'configuration error: missing field `ilp_over_http`', src/libcore/result.rs:999:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

I thought it looked our fault though actually it is just a configuration error. I'll fix the comment and consider later.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -983,50 +986,70 @@ fn prefixed_mesage(challenge: Vec<u8>) -> Vec<u8> {
...
+                "Ethereum Settlement Engine listening on: {}",
+                &opt.http_address
+            );
             Ok(())
         })
 }
 
+#[derive(Deserialize, Clone)]
+pub struct EthereumLedgerOpt {
+    pub private_key: String,
```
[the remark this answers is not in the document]
dora-gt: Indeed. I'll try.

crates/interledger/src/main.rs
dora-gt: I think this is better.

```
Configuration error: "invalid IP address syntax".
Try specifying some missing fields or check its format.
```

If you prefer `expect`, I'll change.

crates/interledger/src/main.rs
```diff
@@ -84,84 +105,143 @@ pub fn main() {
...
+                                .takes_value(true)
+                                .index(1)
+                                .help("Name of config file (in JSON, YAML, or TOML format)"),
                             Arg::with_name("redis_uri")
                                 .long("redis_uri")
-                                .help("Redis database to add the account to")
-                                .default_value("redis://127.0.0.1:6379"),
+                                .default_value("redis://127.0.0.1:6379")
+                                .help("Redis database to add the account to"),
                             Arg::with_name("server_secret")
```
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -983,50 +986,70 @@ fn prefixed_mesage(challenge: Vec<u8>) -> Vec<u8> {
...
+    pub http_address: SocketAddr,
+    pub ethereum_endpoint: String,
```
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger/src/main.rs
dora-gt: I tinkered but finally realized mine was not so good because Serde deserialization process doesn't tell which field was not deserialized when it failed to deserialize. So setting `RUST_BACKTRACE=1` might be helpful. Fixed not to use `get_or_error `.

crates/interledger-settlement-engines/src/main.rs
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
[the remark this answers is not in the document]
dora-gt: It seems that the structure was changed on the master branch, and I'm not sure what is the very correct approach, so just left it as `Secret<String>`.

@gakonst are there more appropriate approaches?

crates/interledger/src/main.rs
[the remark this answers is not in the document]
dora-gt: Although there are some `server_secret` in the source, I couldn't distinguish those were intended. So I just left as were.

crates/interledger/src/main.rs
dora-gt: Should've used `#[serde(default)]` maybe.
I'll check later.

examples/eth-settlement/README.md
```diff
@@ -371,7 +367,7 @@ printf "\nAlice's balance on Bob's node: "
 AB_BALANCE=`curl \
 -H "Authorization: Bearer alice:alice_password" \
 http://localhost:8770/accounts/alice/balance 2>/dev/null`
-EXPECTED_BALANCE='{"balance":"-500"}'
+EXPECTED_BALANCE='{"balance":"0"}'
```
emschwartz: Why is this changed?

crates/interledger/src/main.rs
```diff
@@ -21,50 +26,89 @@ pub fn main() {
...
+        .after_help("")
         .subcommands(vec![
             SubCommand::with_name("spsp")
                 .about("Client and Server for the Simple Payment Setup Protocol (SPSP)")
+                .setting(AppSettings::SubcommandsNegateReqs)
                 .subcommands(vec![
                     SubCommand::with_name("server")
                         .about("Run an SPSP Server that automatically accepts incoming money")
                         .args(&[
+                            Arg::with_name("config")
```
emschwartz:
```suggestion
                            // Positional arguments
                            Arg::with_name("config")
```

crates/interledger/src/main.rs
```diff
@@ -21,50 +26,89 @@ pub fn main() {
...
+                                .takes_value(true)
+                                .index(1)
+                                .help("Name of config file (in JSON, YAML, or TOML format)"),
                             Arg::with_name("port")
```
emschwartz:
```suggestion
                            // Non-positional arguments
                            Arg::with_name("port")
```

docs/operating-manuals/README.md
```diff
@@ -0,0 +1,69 @@
+# Interledger.rs Operating Manuals
```
emschwartz: Since this file is a lot shorter now, what do you think about including these details directly in the top-level README? If you don't think it's worth it, I would get rid of the `operating-manuals` directory and put this file just in `docs`, and then link to it from the README

examples/eth-settlement/README.md
dora-gt: Actually it is not correct. It doesn't work.
Sometime before it has been changed, but since then, it has been not working, I think.

docs/operating-manuals/README.md
dora-gt: I thought it would grow but actually it is very short and I'm OK with moving. I'll fix.

dora-gt: Fixed.

emschwartz:
Merged in as https://github.com/interledger-rs/interledger-rs/pull/290

PR #237, opened 2019-08-22
Username Sanitization

gakonst:
Opening as a separate PR to be merged on top of #233 to avoid cluttering that PR. It does not disallow characters from different scripts or ones which can be considered "confusable". Username length is 2-32 (arbitrary), characters are normalized in unicode and also casefolded.


crates/interledger-service/src/auth/token.rs
```diff
@@ -0,0 +1,104 @@
...
+            username: Username::from_str(username)?,
+            password: password.to_owned(),
+        })
+    }
+
+    pub fn to_bearer(&self) -> String {
+        format!("Bearer {}:{}", self.username, self.password)
+    }
+
+    pub fn username(&self) -> Username {
```
emschwartz: Might be better to return `&Username` so the caller can determine if they need to clone it

crates/interledger-service/src/auth/token.rs
```diff
@@ -0,0 +1,104 @@
...
+        self.username.clone()
+    }
+
+    pub fn password(&self) -> &str {
+        &self.password
+    }
+
+    pub fn parse(s: &str) -> Result<Self, ()> {
```
emschwartz: Why not make this `from_str`?

crates/interledger-service/src/auth/token.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    fn parse_basic(s: &str) -> Result<Self, ()> {
+        let decoded = base64::decode(&s).map_err(|_| ())?;
+        let text = std::str::from_utf8(&decoded).map_err(|_| ())?;
+        Auth::parse_text(text)
+    }
+
+    // Currently, we use bearer tokens in a non-standard way, where they each have a
+    // username and a password in them. In the future, we will deprecate Bearer auth
```
emschwartz:
```suggestion
    // username and a password in them. In the future, we will deprecate the non-standard Bearer auth
```

crates/interledger-service/src/auth/username.rs
```diff
@@ -0,0 +1,105 @@
...
+use serde::{Deserialize, Serialize};
+use unicase::UniCase;
+use unicode_normalization::UnicodeNormalization;
+
+lazy_static! {
+    // No need to add unicode flags, already supported by the crate.
+    static ref USERNAME_PATTERN: Regex = Regex::new(r"^\w{2,32}$").unwrap();
+}
+
+/// Usernames can be unicode and must be between 2 and 32 characters
```
emschwartz:
```suggestion
/// Usernames can be unicode and must be between 2 and 32 characters.
///
/// This type applies the following checks/transformations to usernames to ensure uniqueness:
/// 1. [NFKC Normalization](https://en.wikipedia.org/wiki/Unicode_equivalence#Normalization)
/// 2. Checks the string is 2-32 word characters only (no special characters except `_`)
/// 3. Uses [case folding](https://www.w3.org/International/wiki/Case_folding) to convert to lowercase in a language-aware manner
```

crates/interledger-service/src/auth/username.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+impl AsRef<str> for Username {
+    #[inline]
+    fn as_ref(&self) -> &str {
+        self.0.as_ref()
+    }
+}
+
+impl std::ops::Deref for Username {
```
emschwartz: nit: import `Deref` at the top of the file

crates/interledger-service/src/auth/username.rs
```diff
@@ -0,0 +1,105 @@
...
+    type Target = str;
+
+    fn deref(&self) -> &str {
+        self.0.as_ref()
+    }
+}
+
+impl FromStr for Username {
+    type Err = ();
```
emschwartz: Seems like it would be useful to define actual parsing errors for the different cases (especially useful for the tests to make sure that the cases are being caught because of the intended problem)

crates/interledger-service/src/auth/username.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[test]
+    fn numbers_symbols_work() {
+        assert_eq!(
+            Username::from_str("a_lic123").unwrap().to_string(),
+            "a_lic123"
+        );
+    }
+
+    #[test]
+    fn works_with_unicode() {
```
emschwartz: It would be good to add some tests for case folding and unicode normalization

crates/interledger-store-redis/src/store.rs
```diff
@@ -339,7 +339,7 @@ impl RedisStore {
                     pipe.sadd("accounts", id).ignore();
 
                     // Save map for Username -> Account ID
-                    pipe.hset("usernames", account.username(), id).ignore();
+                    pipe.hset("usernames", account.username().as_bytes().to_vec(), id).ignore();
```
emschwartz: Couldn't the username be passed as a `&str`?

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -209,37 +214,51 @@ impl_web! {
...
+                                Ok(json!(account))
+                            })
+                    )
+                })
             })
         }
 
         #[put("/accounts/:username")]
         #[content_type("application/json")]
         fn http_put_account(&self, username: String, body: AccountDetails, authorization: String) -> impl Future<Item = Value, Error = Response<()>> {
```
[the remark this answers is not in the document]
gakonst: I've tried this but I hit this unimplemented code block in Tower_Web: https://github.com/carllerche/tower-web/blob/master/src/extract/serde.rs#L66-L69. https://github.com/interledger-rs/interledger-rs/commit/e2df406356f46428fec4da0acf73d6fe7353457e

crates/interledger-service/src/auth/username.rs
gakonst: https://github.com/interledger-rs/interledger-rs/pull/237/commits/47d57db990049f07f84455ff829a395258063f46#diff-3aeef3b973a269116a734f4e5e207747R63-R79 ?

crates/interledger-api/src/routes/accounts.rs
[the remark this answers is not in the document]
gakonst: I could not implement Extract directly because [`captures`](https://github.com/carllerche/tower-web/blob/master/src/extract/context.rs#L43-L45) is not exposed publicly (and it is what allows you to get the `:username` parameter). 

We could upstream an implementation, but we probably need to implement it manually, as we want it to perform `Username::from_str(username)` and it seems to me that the derive would just wrap the provided username as `Username(username)`, bypassing any checks.

crates/interledger-btp/src/lib.rs
```diff
@@ -80,6 +80,12 @@ mod client_server {
     };
     use tokio::runtime::Runtime;
 
+    use lazy_static::lazy_static;
+
+    lazy_static! {
```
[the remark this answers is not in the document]
gakonst: It could, but I preferred to keep the struct as it was and just add a default implementation for all since it's unused

PR #238, opened 2019-08-22
chore(get_account_id.sh): get_account_id.sh is required for examples

dora-gt:
Though `get_account_id.sh` is required for examples, it seems that I forgot to add this file when I
was making the PR #232. I'm sorry.

Signed-off-by: Taiga Nakayama <dora@dora-gt.jp>

### DIFF
--- a/scripts/get_account_id.sh
+++ b/scripts/get_account_id.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+# this script could be more generalized but just for now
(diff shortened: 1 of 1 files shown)


gakonst:
I think this is unneeded since we'll be changing everything to use usernames https://github.com/interledger-rs/interledger-rs/pull/233/commits/19c4ec1272b8f2c3a7891b3babeaac2450f90742

dora-gt:
oops, thank you for fixing!

PR #242, opened 2019-08-23
ETH Engine tweaks

### LINKED ISSUE #149: Improve Ethereum Ledger Settlement Engine robustnessss
The engine implemented in #125 works but can be improved. In particular:

- [x] When asking for a peer's address, send a challenge which they must sign so that they prove ownership of the key. Otherwise, it's pretty trivial to pretend you are someone else and get the benefit of their settlements https://github.com/emschwartz/interledger-rs/pull/125#discussion_r305142371
 - [x] extend speculative nonces middleware
 - [x] reduce callback hell https://github.com/emschwartz/interledger-rs/pull/125#discussion_r304152825,
 - [x]  return futures and do join_all (or similar) instead of spawning: https://github.com/emschwartz/interledger-rs/pull/125#discussion_r304679388, https://github.com/emschwartz/interledger-rs/pull/125#discussion_r304679066
 - [x] estimate gas and gas price before sending a transaction https://github.com/emschwartz/interledger-rs/pull/125#discussion_r304045049, allow node operator to specify urgency similar to ethgasstation
 - [x] If a transaction is known to have failed, add a a mechanism to persist the amount that account is owed so it could be retried late.
  - [x] Save gas fees by querying the sender's balance before signing a transaction and sending it only if the sender has sufficient balance (ethers.js does that automatically when it calls eth_estimateGas under the hood, need to check if web3-rust does that check) https://github.com/emschwartz/interledger-rs/pull/125#discussion_r304045693
  - [ ] add test: on what happens if the SE dies after making some of the POST requests to the connector but before it saves the current block number? It should be expected to start again and will try to resubmit the same blocks, however it will not submit the same transaction twice, since the check_tx_credited assertion is sufficient to ensure that a transaction is not processed again. https://github.com/emschwartz/interledger-rs/pull/125#discussion_r304676627
  - [ ] optimize calls to connector in case multiple incoming transactions are found from the same account: https://github.com/emschwartz/interledger-rs/pull/125#discussion_r304677692
  - [x] document edge cases: e.g. If the SE goes down for many blocks and then restarts, this will result in a burst of simultaneous requests, which may fail if using Infura or other ETH providers if they're rate limited. https://github.com/emschwartz/interledger-rs/pull/125#discussion_r304706744
 -  [x] Ensure that the `if new_balance > last_observed_balance` check works as intended even when the operator moves their funds to another address when an incoming payment is received (otherwise remove that `if` since it might be a redundant optimization) https://github.com/emschwartz/interledger-rs/pull/125#discussion_r304040412
  - [x] save recently observed block/balance only when all other operations succeed. If any of the tx settlements fails, it will fetch the blocks since the last successful checkpoint and reexecute all the transactions. The check_tx_credited check will ensure that no transaction gets sent twice.
  - [x] Manage incoming ERC20 transfers with event filtering, example: https://github.com/tomusdrw/rust-web3/blob/master/examples/simple_log_filter.rs. 
- [x] Test that `settle_to` works with both sending ETH and an ERC20 token amount

ERC20 Filter guide:
1. Setup filter builder with from/to block https://tomusdrw.github.io/rust-web3/web3/types/struct.FilterBuilder.html
2. Add the filter for `transfer(address,uint256)` where we filter for `address == our_address`
3. For each hit, notify the connector

gakonst:
Adds test for ERC20s, closes #149
Also connector now returns leftovers as per the latest state of the RFC


crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -509,7 +509,7 @@ where
                     error!("{:?}", error_msg);
                 }))
                 .and_then(move |amount| {
-                    debug!("Got uncredited amount {}", amount);
+                    debug!("Got uncredited amount {}", uncredited_settlement_amount);
```
```suggestion
                    debug!("Loaded uncredited amount {}", uncredited_settlement_amount);
```

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -618,19 +618,28 @@ where
                 .and_then(move |connector_amount: BigUint| {
                     // Scale the amount settled by the
                     // connector back up to our scale
+                    debug!("Scaling amount received from connector back to engine scale: {:?}", quantity);
```
emschwartz: This message is always going to print the same thing. It would be better to take it out or print the specific amounts we're dealing with here

crates/interledger-settlement-engines/src/engines/ethereum_ledger/erc20.code
```diff
@@ -0,0 +1 @@
...ffffffffffffffffffffffffff16815260200190815260200160002054610c1c90919063ffffffff16565b6000803373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002081905550610ae4826000808673ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002054610c3590919063ffffffff16565b6000808573ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001908152602001600020819055508273ffffffffffffffffffffffffffffffffffffffff163373ffffffffffffffffffffffffffffffffffffffff167fddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef846040518082815260200191505060405180910390a36001905092915050565b6000600160008473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002060008373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002054905092915050565b6000828211151515610c2a57fe5b818303905092915050565b6000808284019050838110151515610c4957fe5b809150509291505056fea165627a7a723058201bf0e65af1d88015199e6878e954b688318bdd24ac86c7b473120e8e235996060029
```
emschwartz: If this is just for tests, it seems like it should go in a `fixtures` directory

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -1031,9 +1046,119 @@ mod tests {
             String::from("cc96601bc52293b53c4736a12af9130abf347669b3813f9ec4cafdf6991b087e");
     }
 
+    #[test]
+    fn test_send_erc20() {
```
emschwartz: nit: tests that use external services (arguably even `ganache` should be considered integration tests rather than unit tests and be put in the `tests` directory. I also prefer that because fixtures like the ERC20 contract code can go under `tests/fixtures/...`

crates/interledger-settlement-engines/src/engines/ethereum_ledger/test_helpers.rs
```diff
@@ -266,6 +266,8 @@ pub fn test_engine<Si, S, A>(
     key: Si,
     confs: u8,
     connector_url: &str,
+    ganache_port: u16,
```
emschwartz: This file should also go under `tests/`

crates/interledger-settlement/src/api.rs
```diff
@@ -330,24 +330,26 @@ mod tests {
             let api = test_api(store.clone(), false);
 
             let ret: Response<_> = api
-                .receive_settlement(id.clone(), Quantity::new(200, 18), IDEMPOTENCY.clone())
+                .receive_settlement(id.clone(), Quantity::new(200, 11), IDEMPOTENCY.clone())
```
emschwartz: Why is this being changed? (If there's a good reason for it, please put a comment to explain where this is coming from and/or use a named constant declared at the top of the `settlement_tests` module)

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
gakonst: Yeah but ganache is only used because I did not find a good rust-native EVM-simulator (could probably use parity).

crates/interledger-settlement-engines/src/engines/ethereum_ledger/test_helpers.rs
gakonst: I've moded all of these files under a `test_helpers` directory. In the future we should look into changing from ganache to an actual rust-evm implementation.

crates/interledger-settlement-engines/src/engines/ethereum_ledger/test_helpers/fixtures.rs
```diff
@@ -1,4 +1,4 @@
-use super::test_helpers::TestAccount;
+use super::utils::TestAccount;
```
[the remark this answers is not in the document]
gakonst: This has been refactored to be under `ethereum_ledger/test_helpers/utils.rs`. Figured it'd be nicer to call it `utils` rather than having `test_helpers/test_helpers.rs`

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
gakonst: Addressed in https://github.com/interledger-rs/interledger-rs/pull/242/commits/57b128a01f5e3deb865b0fb83d3d3177ad8b9d61. The `utils.rs` file had to be duplicated because it's needed for the dummy engine store (and cannot be imported from the src directory)

Issue #212, opened 2019-08-15
Don't require accounts with settlement details to be created at the same time


emschwartz:
As mentioned [here](https://github.com/emschwartz/interledger-rs/pull/190#discussion_r314154530), accounts with settlement engine details currently need to be created at the same time. This is fine for testing scripts on the same machine but untenable if these are being configured by two real operators.

The problem is that the SEs try to send their handshakes immediately when an account is configured and they do not retry the call if the first one fails (for example, because the account doesn't yet exist on the other system).

When this is fixed, we should also update the examples to create the accounts in the foreground rather than in the background.

gakonst:
Note, that the engines [DO retry if the first call fails, for `MAX_RETRIES`](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-api/src/routes/accounts.rs#L122). The account insertion process just "hangs", until the other party responds (due to waiting for the future to complete, instead of spawning).

There's a few approaches for this:
- Introduce some form of handshake, i.e. when I add an account to my connector, I also send a ping packet (or similar) to added account. If they respond (hence acknowledging that they have added us to their connector), we make the post request to our engine, and reply with our engine's address.

Hence this would be:
_A adds B to their connector_
A -> B: SYN
B -> A: NACK since they don't have A in their account
_B adds A to their connector_
B -> A: SYN
A -> B: ACK
B -> B_engine: Add A
B -> A: ACK
A -> A_engine: Add B
_end_

If for any reason the above fails, we can add some option in the connector CLI like `interledger-settlement-engines sync-addresses <account_id>` which would make a POST to the engine manually (same as last option basically).

- Spawn the POST to the engine with a fixed interval without a max number of retries (costs a lot of bandwidth, probably not good)
- Do it manually after the 2 connectors have communicated that they're peered (simple, requires more actions from node operator)

emschwartz:
I think the adding a way to manually sync certain accounts (or all of them if no ids are provided) sounds like a good idea but I don't want to introduce another manual step into the normal peering flow.

Having the insert account call hang sounds bad (I know I was the the one that originally suggested doing it that way) because it means that the HTTP request will timeout unless the other party inserts the account within 30 seconds.

What if we do the following:
1. A adds B to their connector
2. A -> A_engine  (exp backoff, 10 retries ONLY on temporary errors, stops on final errors but returns `202 Accepted` if the error was the equivalent of a `404 Not Found`)
3. B adds A to their connector
[...]

PR #244, opened 2019-08-25
Allow accounts to be created at different times

gakonst:
Closes https://github.com/interledger-rs/interledger-rs/issues/212

It always does exponential backoff now. 

@emschwartz described the following behavior:
> exp backoff, 10 retries ONLY on temporary errors, stops on final errors but returns 202 Accepted if the error was the equivalent of a 404 Not Found)

[RetryIf](https://docs.rs/tokio-retry/0.2.0/tokio_retry/struct.RetryIf.html) might be usable for the above, or although futures-retry probably has a better API: https://gitlab.com/mexus/futures-retry/blob/master/examples/tcp-client.rs. Which status codes do we want to retry and which not?

### DIFF
--- a/crates/interledger-api/src/client.rs
+++ b/crates/interledger-api/src/client.rs
@@ -1,12 +1,13 @@
 // Adapted from the futures-retry example: https://gitlab.com/mexus/futures-retry/blob/master/examples/tcp-client-complex.rs
 use futures::future::Future;
 use futures_retry::{ErrorHandler, FutureRetry, RetryPolicy};
-use log::{debug, error, trace};
+use log::trace;
 use reqwest::r#async::Client as HTTPClient;
 use serde_json::json;
 use std::fmt::Display;
 use std::time::Duration;
 use url::Url;
+use http::StatusCode;
 
 // The account creation endpoint set by the engines in [RFC536](https://github.com/interledger/rfcs/pull/536)
 static ACCOUNTS_ENDPOINT: &str = "accounts";
@@ -29,7 +30,7 @@ impl Client {
         &self,
         engine_url: Url,
         id: T,
-    ) -> impl Future<Item = (), Error = reqwest::Error> {
+    ) -> impl Future<Item = StatusCode, Error = reqwest::Error> {
         let mut se_url = engine_url.clone();
         let timeout = self.timeout_ms;
         se_url
@@ -50,9 +51,10 @@ impl Client {
                 .json(&json!({"id" : id.to_string()}))
                 .send()
                 .and_then(move |response| {
-                    trace!("Engine responded with status code: {}", response.status());
-                    // TODO: Do something with the success status code?
-                    Ok(())
+                    // If the account is not found on the peer's connector, the
+                    // retry logic will not get triggered. When the counterparty
+                    // tries to add the account, they will complete the handshake.
+                    Ok(response.status())
                 })
         };
 
@@ -115,8 +117,9 @@ where
             // Retry timeouts and 5xx every 5 seconds
             RetryPolicy::WaitRetry(Duration::from_secs(5))
         } else {
-            // Instantly retry other errors since they may be HTTP/redirect related
-            RetryPolicy::Repeat
+            // Retry other errors slightly more frequently since they may be
+            // related to the engine not having started yet
+            RetryPolicy::WaitRetry(Duration::from_secs(1))
         }
     }
 }
--- a/crates/interledger-api/src/client.rs
+++ b/crates/interledger-api/src/client.rs
@@ -1,13 +1,13 @@
 // Adapted from the futures-retry example: https://gitlab.com/mexus/futures-retry/blob/master/examples/tcp-client-complex.rs
 use futures::future::Future;
 use futures_retry::{ErrorHandler, FutureRetry, RetryPolicy};
+use http::StatusCode;
 use log::trace;
 use reqwest::r#async::Client as HTTPClient;
 use serde_json::json;
 use std::fmt::Display;
 use std::time::Duration;
 use url::Url;
-use http::StatusCode;
 
 // The account creation endpoint set by the engines in [RFC536](https://github.com/interledger/rfcs/pull/536)
 static ACCOUNTS_ENDPOINT: &str = "accounts";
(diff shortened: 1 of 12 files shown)


gakonst:
> If the problem is either of the first two, the account creation step should probably fail (meaning we can't spawn the call to create the account in all cases).

I agree with this. We should not spawn for these 2 cases.

kincaidoneil:
> To handle the case where the account wasn't created on the other side, it seems more like a change to the SE and the send message API than to the account creation API. The SE needs to understand whatever error code it gets back from trying to send the handshake message as meaning "the account hasn't been created yet" and return a 202 or something when the account is created.

For this case, why does the SE need to know the account wasn't created on the peer's end?

Within SEs, I think this behavior would suffice:

1. On account setup, perform a handshake/attempt to send the message to exchange identifiers (if it fails, that's fine)
2. On sending an outgoing settlement, if no identifier is linked, try to perform the handshake before settling (if it fails again, the settlement amount will be tracked and retried later)
3. As new settlements are triggered (or when the peer sends its handshake message), step 2 can be repeated

(This is more or less the behavior that exists with the JavaScript ETH plugin; it never checks if it's connected to the peer or anything).

gakonst:
From call with Kincaid:

Error retry policy:
- retry timeouts
- retry 5xx
- do not retry 4xx

Account creation process handshake should be:
1. alice tries to add bob to her connector
1. alice's engine tries to send alice's address to bob's engine (via the connector)
1. bob has not configured alice in his connector so the call fails with a terminal error (alice's engine won't retry)
1. account creation moves forward w/o configuring alice's engine
1. bob tries to add alice to his connector
1. bob's engine sends bob's address to alice's engine (via the connector)
1. alice receives that, **and creates bob's account in her store** (this means that bob's data is now fully configured on alice's connector/engine)
1. alice's engine acks and sends bob her address
1. bob's engine saves alice's address
1. alice's account is now fully configured on bob's connector/engine

crates/interledger-api/src/client.rs
```diff
@@ -0,0 +1,125 @@
...
+use futures_retry::{ErrorHandler, FutureRetry, RetryPolicy};
+use http::StatusCode;
+use log::trace;
+use reqwest::r#async::Client as HTTPClient;
+use serde_json::json;
+use std::fmt::Display;
+use std::time::Duration;
+use url::Url;
+
+// The account creation endpoint set by the engines in [RFC536](https://github.com/interledger/rfcs/pull/536)
```
emschwartz: That isn't the RFC number, that's the PR number

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -39,15 +41,27 @@ use interledger_settlement::{Convert, ConvertDetails, Quantity};
...
+impl PaymentDetailsRequest {
+    fn new(challenge: Vec<u8>) -> Self {
+        PaymentDetailsRequest { challenge }
+    }
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 struct PaymentDetailsResponse {
     to: Addresses,
     sig: Signature,
```
emschwartz: nit: any reason not to call it `signature`?

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -840,13 +863,53 @@ where
...
+                            move |_| {
+                                // ACK BACK
+                                if let Some(challenge) = payment_details.challenge {
+                                    // if we were challenged, we must respond
+                                    let data = prefixed_message(challenge);
+                                    let signature = signer.sign_message(&data);
+                                    let resp = {
+                                        // Respond with our address, a signature,
+                                        // and no challenge, since we already sent
+                                        // them one earlier
```
emschwartz: Does this work even if the first request to do the handshake failed?

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -864,20 +927,72 @@ where
...
+            debug!("Received payment details: {:?}", resp);
+            let guard = self.challenges.read();
+            let fut = if let Some(challenge) = (*guard).get(&account_id) {
+                // if we sent them a challenge, we will verify the received
+                // signature and address, and if sig verification passes, we'll
+                // save them in our store
+                let data = prefixed_message(challenge.to_vec());
+                let challenge_hash = Sha3::digest(&data);
+                let recovered_address = resp.sig.recover(&challenge_hash);
+                Either::A(
```
emschwartz: This is synchronous code so it seems kind of strange to use futures here

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -864,20 +927,72 @@ where
...
+            } else {
+                Either::B(ok(()))
+            };
+
+            Box::new(
+                fut.and_then(move |_| Ok((StatusCode::from_u16(200).unwrap(), "OK".to_string()))),
+            )
+        } else {
+            Box::new(err((
+                StatusCode::from_u16(502).unwrap(),
```
emschwartz: Shouldn't this be a 400?

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -864,20 +930,74 @@ where
...
```
```suggestion
                StatusCode::from_u16(400).unwrap(),
```

crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -864,20 +930,74 @@ where
...
+                }
+            } else {
+                Either::B(ok(()))
+            };
+
+            Box::new(
+                fut.and_then(move |_| Ok((StatusCode::from_u16(200).unwrap(), "OK".to_string()))),
+            )
+        } else {
+            Box::new(err((
```
```suggestion
            error!("Ignoring message that was neither a PaymentDetailsRequest nor a PaymentDetailsResponse");
            Box::new(err((
```

PR #245, opened 2019-08-26
Hide secrets

### LINKED ISSUE #108: Improve key storage approach
Right now, the Store is responsible for managing secret values like API keys (for authenticating with peers via HTTP or BTP). The Redis store uses encryption and hmac keys derived from a [configured seed](https://github.com/emschwartz/interledger-rs/blob/21dd48b3007098aae6dc39aa3cada7add5fbda2f/crates/interledger-store-redis/src/store.rs#L135).

This approach isn't great because:
- It's not very secure (all of they key material is on the same machine)
- The seed needs to be passed in as an environment variable (is that even a good approach?)
- Cannot easily change the seed
- Every store implementation needs to re-implement similar crypto functionality or import some shared code (this issue was inspired by @cjimison asking about the right way to approach this for https://github.com/emschwartz/interledger-rs/issues/52) 

Also, the keys should be included in the `Account` type in such a way that they are never logged. For example, they could have special types that implement `Debug` and `Display` but only print `<redacted>` or something like that.

How should we improve the key storage? If necessary, we can also revisit the use of API keys (rather than more complex auth) if there is some decent solution that's more secure but not impossible to use.

gakonst:
will close #108

### DIFF
--- a/crates/interledger-settlement-engines/Cargo.toml
+++ b/crates/interledger-settlement-engines/Cargo.toml
@@ -37,8 +37,8 @@ sha3 = "0.8.2"
 num-bigint = "0.2.2"
 num-traits = "0.2.8"
 lazy_static = "1.3.0"
-secrecy = { git = "https://github.com/iqlusioninc/crates", branch = "develop", features = ["serde", "bytes"] }
-zeroize = { git = "https://github.com/iqlusioninc/crates", branch = "develop", features = ["bytes"] }
+secrecy = { version = "0.4.0", features = ["serde", "bytes"] }
+zeroize = { version = "0.10.1", features = ["bytes"] }
 
 [dev-dependencies]
 lazy_static = "1.3"
--- a/crates/interledger-settlement-engines/Cargo.toml
+++ b/crates/interledger-settlement-engines/Cargo.toml
@@ -37,6 +37,8 @@ sha3 = "0.8.2"
 num-bigint = "0.2.2"
 num-traits = "0.2.8"
 lazy_static = "1.3.0"
+secrecy = { git = "https://github.com/iqlusioninc/crates", branch = "develop", features = ["serde", "bytes"] }
+zeroize = { git = "https://github.com/iqlusioninc/crates", branch = "develop", features = ["bytes"] }
 
 [dev-dependencies]
 lazy_static = "1.3"
--- a/crates/interledger-store-redis/README.md
+++ b/crates/interledger-store-redis/README.md
@@ -25,9 +25,9 @@ This is tracked separately from the `balance` to avoid the ["settling back and f
 The `asset_code` and `asset_scale` for each of the accounts' balances can be found in the Account Details hash map. 
 Note that this means that accounts' balances are not directly comparable (for example if account 1's `balance` is 100 and account 2's `balance` is 1000, this does not necessarily mean that we owe accountholder 2 more than accountholder 1, because these values represent completely different assets).
 
-#### Outgoing Auth Tokens
+#### Incoming / Outgoing Auth Tokens
 
-Outgoing auth tokens are encrypted in the following manner:
+Auth tokens are encrypted in the following manner:
 - The encryption/decryption key is generated as `hmac_sha256(store_secret, "ilp_store_redis_encryption_key")`
 - Tokens are encrypted using the AES-256-GCM symmetric encryption scheme using 12-byte randomly generated nonces
 - The nonce is appended to the encrypted output (which includes the auth tag) and stored in the DB
@@ -41,15 +41,6 @@ Statically configured routes are stored as a hash map of prefix to account ID un
 ### Exchange Rates
 
 Exchange rates are stored as a hash map of currency code to rate under the key `rates:current`.
-
-### HTTP / BTP Auth Details
-
-Incoming HTTP and BTP authentication tokens are stored in hash maps (under `http_auth` and `btp_auth`, respectively) for fast lookup of which account corresponds to a given auth token.
-
-The auth tokens are stored in the following manner:
-- The HMAC key is generated as `hmac_sha256(store_secret, "ilp_store_redis_hmac_key")`
-- Only the output of `hmac_sha256(hmac_key, auth_token)` is stored in the database
-- The hash maps are mappings of the HMAC output to an account ID
 
 ### Rate Limiting
 
--- a/crates/interledger-store-redis/src/account.rs
+++ b/crates/interledger-store-redis/src/account.rs
@@ -198,23 +198,19 @@ impl AccountWithEncryptedTokens {
     pub fn decrypt_tokens(mut self, decryption_key: &aead::OpeningKey) -> Account {
         if let Some(ref encrypted) = self.account.btp_outgoing_token {
             self.account.btp_outgoing_token =
-                decrypt_token(decryption_key, &encrypted.expose_secret())
-                    .map(|decrypted| SecretBytes::from(decrypted));
+                decrypt_token(decryption_key, &encrypted.expose_secret()).map(SecretBytes::from);
         }
         if let Some(ref encrypted) = self.account.http_outgoing_token {
             self.account.http_outgoing_token =
-                decrypt_token(decryption_key, &encrypted.expose_secret())
-                    .map(|decrypted| SecretBytes::from(decrypted));
+                decrypt_token(decryption_key, &encrypted.expose_secret()).map(SecretBytes::from);
         }
         if let Some(ref encrypted) = self.account.btp_incoming_token {
             self.account.btp_incoming_token =
-                decrypt_token(decryption_key, &encrypted.expose_secret())
-                    .map(|decrypted| SecretBytes::from(decrypted));
+                decrypt_token(decryption_key, &encrypted.expose_secret()).map(SecretBytes::from);
         }
         if let Some(ref encrypted) = self.account.http_incoming_token {
             self.account.http_incoming_token =
-                decrypt_token(decryption_key, &encrypted.expose_secret())
-                    .map(|decrypted| SecretBytes::from(decrypted));
+                decrypt_token(decryption_key, &encrypted.expose_secret()).map(SecretBytes::from);
         }
 
         self.account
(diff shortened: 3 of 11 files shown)


crates/interledger-settlement-engines/Cargo.toml
```diff
@@ -37,6 +37,7 @@ sha3 = "0.8.2"
 num-bigint = "0.2.2"
 num-traits = "0.2.8"
 lazy_static = "1.3.0"
+secrecy = { git = "https://github.com/gakonst/crates", branch = "gakonst/cloneable-secret-string" }
```
gakonst: Tony just published a newer version including a slight change I made, will update to use that in the next commits.

crates/interledger-store-redis/src/account.rs
```diff
@@ -21,13 +21,37 @@ use std::fmt::Display;
...
 };
 use uuid::{parser::ParseError, Uuid};
 
 use url::Url;
 const ACCOUNT_DETAILS_FIELDS: usize = 21;
 
+use secrecy::{CloneableSecret, DebugSecret, ExposeSecret, Secret};
+use zeroize::Zeroize;
+
+/// An ApiToken is a wrapper over Bytes for which all the Secrecy related traits
```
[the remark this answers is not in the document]
gakonst: I'll do that.

crates/interledger-store-redis/src/account.rs
```diff
@@ -173,10 +223,24 @@ pub struct AccountWithEncryptedTokens {
 impl AccountWithEncryptedTokens {
     pub fn decrypt_tokens(mut self, decryption_key: &aead::OpeningKey) -> Account {
         if let Some(ref encrypted) = self.account.btp_outgoing_token {
-            self.account.btp_outgoing_token = decrypt_token(decryption_key, encrypted);
+            self.account.btp_outgoing_token =
+                decrypt_token(decryption_key, &encrypted.expose_secret().0)
```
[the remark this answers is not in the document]
gakonst: It does not, but since `AccountWithEncryptedTokens` has an `Account` inside it, the encrypted value must be wrapped in an `Secret` type. Otherwise, I can make `AccountWithEncryptedTokens` have all the same fields as `Account`, except for the tokens which will be set as `Bytes`, since we assume they're already encrypted. Do you think that change should be made? Seems unnecessary

crates/interledger-store-redis/src/store.rs
```diff
@@ -273,6 +261,46 @@ impl RedisStoreBuilder {
...
+// Note: This is probably bad, and will crash if the key is zero'd out and is
+// accessed again. Discuss other ways to forcibly clear the memory
+// of Sealing/Opening keys
+impl Zeroize for EncryptionKey {
+    fn zeroize(&mut self) {
+        let size = mem::size_of_val(self);
+        unsafe {
+            let ptr = self as *mut Self;
+            ptr::drop_in_place(ptr);
+            ptr::write_bytes(ptr as *mut u8, 0, size);
```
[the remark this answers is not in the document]
gakonst: I've manually implemented this, instead of clearing the memory, I overwrite it with a [0;32] private key.

crates/interledger-store-redis/src/account.rs
```diff
@@ -21,13 +21,37 @@ use std::fmt::Display;
...
+/// are implemented. ApiTokens should be used as Secret<ApiToken>.
+#[derive(Default, Debug, Serialize, Deserialize, Clone)]
+pub struct ApiToken(pub(crate) Bytes);
+
+impl DebugSecret for ApiToken {}
+impl CloneableSecret for ApiToken {}
+
+impl Zeroize for ApiToken {
+    fn zeroize(&mut self) {
+        self.0.clear();
```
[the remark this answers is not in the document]
gakonst: This is no longe relevant since we use the upstream `SecretBytes` from `secrecy`

crates/interledger-store-redis/src/crypto.rs
```diff
@@ -27,7 +20,8 @@ pub fn generate_keys(
         hmac::sign(&generation_key, ENCRYPTION_KEY_GENERATION_STRING).as_ref(),
     )
     .unwrap();
-    (hmac_key, encryption_key, decryption_key)
+    // is the generation key cleared from memory when it's dropped here?
```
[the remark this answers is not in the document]
gakonst: I've wrapped the generation key in a type that implements Zeroize/Drop, in the same way as Encryption/Decryption key. This should be sufficient to guarantee that it's cleared/overwritten with junk data once it's dropped.

crates/interledger-store-redis/src/account.rs
```diff
@@ -61,20 +64,22 @@ pub struct Account {
...
-
 fn address_to_string<S>(address: &Address, serializer: S) -> Result<S::Ok, S::Error>
 where
     S: Serializer,
 {
     serializer.serialize_str(str::from_utf8(address.as_ref()).unwrap_or(""))
 }
 
-fn optional_bytes_to_utf8<S>(bytes: &Option<Bytes>, serializer: S) -> Result<S::Ok, S::Error>
+fn optional_secret_bytes_to_utf8<S>(
```
emschwartz: I don't think we ever want to expose this. It should just print "secret" or something like that

crates/interledger-store-redis/src/crypto.rs
```diff
@@ -6,28 +6,107 @@ use ring::{
...
+#[derive(Debug)]
+pub struct GenerationKey(pub(crate) hmac::SigningKey);
+
+impl DebugSecret for EncryptionKey {}
+impl DebugSecret for DecryptionKey {}
+impl DebugSecret for GenerationKey {}
+
+impl Zeroize for EncryptionKey {
+    fn zeroize(&mut self) {
+        // Instead of clearing the memory, we create a key from an empty
```
emschwartz: It's not empty -- empty suggests a zero-length (the other copies of this comment should be updated too)

```suggestion
        // Instead of clearing the memory, we create overwrite the key with
```

crates/interledger-store-redis/src/crypto.rs
```diff
@@ -6,28 +6,107 @@ use ring::{
...
+        // slice as a secret.
```
emschwartz:
```suggestion
        // a slice filled with zeros.
```

crates/interledger-store-redis/src/crypto.rs
```diff
@@ -6,28 +6,107 @@ use ring::{
...
+    }
+}
+
+impl Drop for GenerationKey {
+    fn drop(&mut self) {
+        self.zeroize()
+    }
+}
+
+// this logic is taken from: https://github.com/iqlusioninc/crates/blob/develop/zeroize/src/lib.rs#L388-L400
```
emschwartz: Include a permalink

crates/interledger-store-redis/src/crypto.rs
```diff
@@ -6,28 +6,107 @@ use ring::{
...
+// Perform a volatile write to the destination
```
emschwartz: This description doesn't add anything to what the function name implies. Could you give a better sense of what it's used for and how we know its safe?

crates/interledger-store-redis/src/crypto.rs
```diff
@@ -6,28 +6,107 @@ use ring::{
...
+#[inline]
+fn volatile_write<T>(dst: &mut T, src: T) {
+    unsafe { ptr::write_volatile(dst, src) }
+}
+
+/// Use fences to prevent accesses from being reordered before this
```
emschwartz: Maybe include a link to more information about how this works?

crates/interledger-store-redis/src/crypto.rs
```diff
@@ -6,28 +6,107 @@ use ring::{
...
+/// point, which should hopefully help ensure that all accessors
```
emschwartz: "hopefully" doesn't inspire a lot of confidence

gakonst: Right, this was taken verbatim from Tony's crate: https://github.com/iqlusioninc/crates/blob/develop/zeroize/src/lib.rs#L388-L394

crates/interledger-store-redis/src/account.rs
gakonst: Good point, thanks.

crates/interledger-store-redis/src/crypto.rs
gakonst: The function being called has documentation describing it: https://doc.rust-lang.org/core/sync/atomic/fn.compiler_fence.html, do we need additional?

emschwartz: Maybe link to that in the comment because it's not a very commonly used feature

crates/interledger-settlement-engines/Cargo.toml
```diff
@@ -37,6 +37,8 @@ sha3 = "0.8.2"
 num-bigint = "0.2.2"
 num-traits = "0.2.8"
 lazy_static = "1.3.0"
+secrecy = { git = "https://github.com/iqlusioninc/crates", branch = "develop", features = ["serde", "bytes"] }
+zeroize = { git = "https://github.com/iqlusioninc/crates", branch = "develop", features = ["bytes"] }
```
tarcieri: I did releases of both of these (`secrecy` v0.4.0 and `zeroize` v0.10.1)

crates/interledger-store-redis/src/crypto.rs
```diff
@@ -6,28 +6,112 @@ use ring::{
...
+    }
+}
+
+impl Drop for GenerationKey {
+    fn drop(&mut self) {
+        self.zeroize()
+    }
+}
+
+// this logic is taken from [here](https://github.com/iqlusioninc/crates/blob/develop/zeroize/src/lib.rs#L388-L400)
```
emschwartz:
```suggestion
// this logic is taken from [here](https://github.com/iqlusioninc/crates/blob/b1937857418c3e2c46723d1687158b402a5e3f14/zeroize/src/lib.rs#L388-L400)
```

crates/interledger-store-redis/README.md
```diff
@@ -25,9 +25,9 @@ This is tracked separately from the `balance` to avoid the ["settling back and f
 The `asset_code` and `asset_scale` for each of the accounts' balances can be found in the Account Details hash map. 
 Note that this means that accounts' balances are not directly comparable (for example if account 1's `balance` is 100 and account 2's `balance` is 1000, this does not necessarily mean that we owe accountholder 2 more than accountholder 1, because these values represent completely different assets).
 
-#### Outgoing Auth Tokens
+#### Incoming / Outgoing Auth Tokens
 
-Outgoing auth tokens are encrypted in the following manner:
+Auth tokens are encrypted in the following manner:
```
tarcieri: Orthogonal to this PR, but this seems like a good use case for AES-(GCM-)SIV

emschwartz:
@gakonst please rebase and squash and/or rename the commits before merging

PR #247, opened 2019-08-27
fix(service, examples): reject invalid usernames during account creation

bstrie:
The new username feature doesn't allow names to contain special
characters other than underscore, but the simple example had usernames
of the form "node-a". However, this was only being detected during
account lookup, whereas account creation was bypassing the validation
check. This occurred because the intended function by which to construct
a valid instance of Username, `Username::from_str`, was being explicitly
invoked during account lookup; for account creation, the instance of
Username was constructed directly via the Deserialize implementation
generated by Serde. Serde has an attribute hook that allows types to run
custom validation logic via the TryFrom trait; we implement TryFrom for
Username here in terms of the existing FromStr impl (in some future
version of Rust this may be unnecessary, as specialization should allow
TryFrom and FromStr to be defined in terms of each other within libstd).

### DIFF
--- a/crates/interledger-service/src/auth/username.rs
+++ b/crates/interledger-service/src/auth/username.rs
@@ -1,5 +1,6 @@
 /// Wrapper around String to perform sanitization for usernames
 use regex::Regex;
+use std::convert::TryFrom;
 use std::fmt::Display;
 use std::str::FromStr;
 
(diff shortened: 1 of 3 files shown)


bstrie:
CI appears to be failing for an unrelated reason, is that known?

gakonst:
This has to do with the XRP testnet credentials. Looking..

crates/interledger-service/src/auth/username.rs
```diff
@@ -20,6 +21,7 @@ lazy_static! {
 /// 2. Checks the string is 2-32 word characters only (no special characters except `_`)
 /// 3. Uses [case folding](https://www.w3.org/International/wiki/Case_folding) to convert to lowercase in a language-aware manner
 #[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(try_from = "&str")]
```
gakonst: Ah, great, did not know of this feature. Thanks.

bstrie:
@gakonst Thanks, PR rebased, but do we need @emschwartz to approve it before it can actually be merged? I seem to recall that he may be on vacation this week.

PR #259, opened 2019-08-30
feat(eth se): port is deprecated, use http_address instead

dora-gt:
**This PR should be merged after #236 is merged**

Specifying a port for the settlement engine is not sufficient because sometimes we need to
specify the address, for example, running on Docker.

BREAKING CHANGE: Now --port option is obsolete, use http_address instead.

Signed-off-by: Taiga Nakayama <dora@dora-gt.jp>


dora-gt:
I'll `fmt` in #236 .

dora-gt:
oh wow... `cargo fmt` makes the file uncompilable. how...

Issue #216, opened 2019-08-15
Remove Interledger Store Memory


gakonst:
Currently it's only used in the CLI, where I'd expect that we would pass a redis instance instead. Do we need it or can we remove it?

emschwartz:
The CLI functions aside from the `node` are completely stateless so it wouldn't make sense to require Redis for them.

That said, we _may_ want to remove those functions and just point people to use the node. In the future we could add a different type of CLI that makes HTTP calls to the `node`.

emschwartz:
> That said, we _may_ want to remove those functions and just point people to use the node. In the future we could add a different type of CLI that makes HTTP calls to the `node`.

https://github.com/interledger-rs/interledger-rs/issues/228

gakonst:
OK so once #228 is in, we should remove the in-memory store along with `cli.rs` related functionality.

bstrie:
Closed by #321.

PR #267, opened 2019-08-31
feature: publish payment packets to pub/sub

bstrie:
In the interest of breaking up PRs into smaller chunks, here's the publish half of the pub/sub feature that's incoming.

### DIFF
--- a/crates/interledger-service/src/lib.rs
+++ b/crates/interledger-service/src/lib.rs
@@ -27,13 +27,14 @@
 //! HttpServerService --> ValidatorService --> StreamReceiverService
 
 use futures::{Future, IntoFuture};
-use interledger_packet::{Fulfill, Prepare, Reject};
+use interledger_packet::{Address, Fulfill, Prepare, Reject};
 use std::{
     cmp::Eq,
     fmt::{Debug, Display},
     hash::Hash,
     marker::PhantomData,
     str::FromStr,
+    sync::Arc,
 };
 
 use serde::Serialize;
@@ -118,6 +119,22 @@ pub trait AccountStore {
         dyn Future<Item = <<Self as AccountStore>::Account as Account>::AccountId, Error = ()>
             + Send,
     >;
+}
+
+/// A trait representing the Publish side of a pub/sub store
+pub trait PubStore {
+    fn publish_payment_notification(
+        self: Arc<Self>,
+        _sender: Username,
+        _receiver: Username,
+        _sender_address: Address,
+        _amount: u64,
+        _timestamp: i64,
+    ) {
+        // Publishing a notification is an operation we care about only for its side effects,
+        // so for ease of mocking we give it a default implementation that is entirely empty.
+        // In other words: This Space Intentionally Left Blank.
+    }
 }
 
 /// Create an IncomingService that calls the given handler for each request.
--- a/crates/interledger-store-memory/src/store.rs
+++ b/crates/interledger-store-memory/src/store.rs
@@ -8,7 +8,7 @@ use interledger_btp::{BtpOpenSignupAccount, BtpOpenSignupStore, BtpStore};
 use interledger_http::HttpStore;
 use interledger_ildcp::IldcpAccount;
 use interledger_router::RouterStore;
-use interledger_service::{Account as AccountTrait, AccountStore, Username};
+use interledger_service::{Account as AccountTrait, AccountStore, PubStore, Username};
 use parking_lot::{Mutex, RwLock};
 use std::collections::HashMap;
 use std::{
@@ -137,6 +137,8 @@ impl AccountStore for InMemoryStore {
         Box::new(ok(1))
     }
 }
+
+impl PubStore for InMemoryStore {}
 
 impl HttpStore for InMemoryStore {
     type Account = Account;
--- a/crates/interledger-store-redis/src/store.rs
+++ b/crates/interledger-store-redis/src/store.rs
@@ -19,7 +19,7 @@ use super::account::*;
 use super::crypto::{encrypt_token, generate_keys, DecryptionKey, EncryptionKey};
 use bytes::Bytes;
 use futures::{
-    future::{err, ok, result, Either},
+    future::{err, lazy, ok, result, Either},
     Future, Stream,
 };
 use log::{debug, error, trace, warn};
@@ -31,8 +31,9 @@ use interledger_api::{AccountDetails, AccountSettings, NodeStore};
 use interledger_btp::BtpStore;
 use interledger_ccp::RouteManagerStore;
 use interledger_http::HttpStore;
+use interledger_packet::Address;
 use interledger_router::RouterStore;
-use interledger_service::{Account as AccountTrait, AccountStore, Username};
+use interledger_service::{Account as AccountTrait, AccountStore, PubStore, Username};
 use interledger_service_util::{BalanceStore, ExchangeRateStore, RateLimitError, RateLimitStore};
 use interledger_settlement::{IdempotentData, IdempotentStore, SettlementStore};
 use lazy_static::lazy_static;
@@ -612,6 +613,21 @@ impl RedisStore {
                 ),
         )
     }
+
+    fn redis_publish(
+        &self,
+        channel: &str,
+        message: &str,
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        Box::new(
+            redis::cmd("PUBLISH")
+                .arg(channel)
+                .arg(message)
+                .query_async(self.connection.as_ref().clone())
+                .map_err(|_| ())
+                .and_then(|(_, _): (_, i32)| Ok(())),
+        )
+    }
 }
 
 impl AccountStore for RedisStore {
@@ -637,6 +653,24 @@ impl AccountStore for RedisStore {
                 .map_err(move |err| error!("Error getting account id: {:?}", err))
                 .and_then(|(_connection, id): (_, AccountId)| Ok(id)),
         )
+    }
+}
+
+impl PubStore for RedisStore {
+    fn publish_payment_notification(
+        self: Arc<Self>,
+        receiver: Username,
+        sender: Username,
+        sender_address: Address,
+        amount: u64,
+        timestamp: i64,
+    ) {
+        let channel = format!("p:{}", receiver);
+        let message = format!("{}:{}:{}:{}", sender, sender_address, amount, timestamp);
+        spawn(lazy(move || {
+            debug!("Publishing message {} to channel {}", message, channel);
+            self.redis_publish(&channel, &message)
+        }));
     }
 }
 
--- a/crates/interledger-stream/src/server.rs
+++ b/crates/interledger-stream/src/server.rs
@@ -2,17 +2,20 @@ use super::crypto::*;
 use super::packet::*;
 use base64;
 use bytes::Bytes;
+use chrono::Utc;
 use futures::future::result;
+use futures::Future;
 use hex;
 use interledger_ildcp::IldcpAccount;
 use interledger_packet::{
     Address, ErrorCode, Fulfill, FulfillBuilder, PacketType as IlpPacketType, Prepare, Reject,
     RejectBuilder,
 };
-use interledger_service::{Account, BoxedIlpFuture, OutgoingRequest, OutgoingService};
+use interledger_service::{Account, BoxedIlpFuture, OutgoingRequest, OutgoingService, PubStore};
 use log::debug;
 use std::convert::TryFrom;
 use std::marker::PhantomData;
+use std::sync::Arc;
 
 const STREAM_SERVER_SECRET_GENERATOR: &[u8] = b"ilp_stream_secret_generator";
 
@@ -92,30 +95,33 @@ impl ConnectionGenerator {
 ///
 /// This does not currently support handling data sent via STREAM.
 #[derive(Clone)]
-pub struct StreamReceiverService<O: OutgoingService<A>, A: Account> {
+pub struct StreamReceiverService<S, O: OutgoingService<A>, A: Account> {
     connection_generator: ConnectionGenerator,
     next: O,
     account_type: PhantomData<A>,
+    store: Arc<S>,
 }
 
-impl<O, A> StreamReceiverService<O, A>
+impl<S, O, A> StreamReceiverService<S, O, A>
 where
+    S: PubStore,
     O: OutgoingService<A>,
     A: Account,
 {
-    pub fn new(server_secret: Bytes, next: O) -> Self {
+    pub fn new(server_secret: Bytes, store: S, next: O) -> Self {
         let connection_generator = ConnectionGenerator::new(server_secret);
         StreamReceiverService {
             connection_generator,
             next,
             account_type: PhantomData,
+            store: Arc::new(store),
         }
     }
 }
 
-// TODO should this be an OutgoingService instead so the balance logic is applied before this is called?
-impl<O, A> OutgoingService<A> for StreamReceiverService<O, A>
+impl<S, O, A> OutgoingService<A> for StreamReceiverService<S, O, A>
 where
+    S: PubStore + Send + Sync + 'static,
     O: OutgoingService<A>,
     A: Account + IldcpAccount,
 {
@@ -128,14 +134,31 @@ where
     /// the server to check whether the Prepare packet was created with STREAM parameters
     /// that this server would have created or not.
     fn send_request(&mut self, request: OutgoingRequest<A>) -> Self::Future {
+        let to_name = request.to.username().clone();
+        let from_name = request.from.username().clone();
+        let from_address = request.from.client_address().clone();
+        let amount = request.prepare.amount();
+        let store = self.store.clone();
+
         let destination = request.prepare.destination();
-        let to = request.to.client_address();
+        let to_address = request.to.client_address();
         let dest: &[u8] = destination.as_ref();
-        if dest.starts_with(to.as_ref()) {
+        if dest.starts_with(to_address.as_ref()) {
             if let Ok(shared_secret) = self.connection_generator.rederive_secret(&destination) {
-                {
-                    return Box::new(result(receive_money(&shared_secret, &to, request.prepare)));
-                }
+                return Box::new(
+                    result(receive_money(&shared_secret, &to_address, request.prepare)).and_then(
+                        move |val| {
+                            store.publish_payment_notification(
+                                to_name,
+                                from_name,
+                                from_address,
+                                amount,
+                                Utc::now().timestamp_millis(),
+                            );
+                            Ok(val)
+                        },
+                    ),
+                );
             }
         }
         Box::new(self.next.send_request(request))
@@ -458,6 +481,7 @@ mod stream_receiver_service {
 
         let mut service = StreamReceiverService::new(
             server_secret.clone(),
+            DummyStore,
             outgoing_service_fn(|_: OutgoingRequest<TestAccount>| -> BoxedIlpFuture {
                 panic!("shouldn't get here")
             }),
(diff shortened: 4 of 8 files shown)


crates/interledger-stream/src/server.rs
```diff
@@ -92,30 +94,34 @@ impl ConnectionGenerator {
 ///
 /// This does not currently support handling data sent via STREAM.
 #[derive(Clone)]
-pub struct StreamReceiverService<O: OutgoingService<A>, A: Account> {
+pub struct StreamReceiverService<S, O: OutgoingService<A>, A: Account> {
     connection_generator: ConnectionGenerator,
     next: O,
     account_type: PhantomData<A>,
+    store: Arc<S>,
```
gakonst: Might be a better idea to make the store an Option? Should remove the need for the `impl PubStore for ()` as well.

gakonst: Might be a better idea to make the store an Option? Should remove the need for the impl PubStore for () as well.

crates/interledger-stream/src/server.rs
```diff
@@ -458,6 +486,7 @@ mod stream_receiver_service {
 
         let mut service = StreamReceiverService::new(
             server_secret.clone(),
+            (),
```
gakonst: e.g. here we'd make it None since there's no pubsub needed

crates/interledger-store-redis/src/store.rs
```diff
@@ -565,6 +565,21 @@ impl RedisStore {
...
         )
     }
+
+    fn redis_publish(
+        &self,
+        channel: &str,
+        message: &str,
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        Box::new(
+            redis::cmd("PUBLISH")
```
gakonst: https://github.com/mitsuhiko/redis-rs/pull/109/files#diff-a807dcb040d2e42176f752e520f7f519L449 there's a command for this we could directly use

crates/interledger-service/src/lib.rs
```diff
@@ -120,6 +120,21 @@ pub trait AccountStore {
...
+    fn publish(
+        &self,
+        channel: &str,
+        message: &str,
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
+}
+
+impl PubStore for () {
+    fn publish(&self, _: &str, _: &str) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        unimplemented!()
```
emschwartz: If this isn't implemented here, I think the trait should be defined wherever it is actually used

crates/interledger-store-memory/src/store.rs
```diff
@@ -138,6 +138,12 @@ impl AccountStore for InMemoryStore {
     }
 }
 
+impl PubStore for InMemoryStore {
+    fn publish(&self, _: &str, _: &str) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        unimplemented!()
```
emschwartz: Why implement the trait for the store if it's actually going to panic when used? Part of the point of using traits for store methods is to ensure at compile time that the store implements all of the necessary functionality

crates/interledger-store-redis/src/store.rs
emschwartz: Actually it doesn't currently support using the `PUBLISH` command with async connections, see https://github.com/mitsuhiko/redis-rs/issues/183

Using `PUBLISH` changes the mode of the redis connection so that it doesn't allow any other commands to be sent on the same connection (see https://redis.io/topics/pubsub), so I don't think this will work properly as implemented. If we want to use Redis PubSub we'll need at least 2 connections to the Redis instance

crates/interledger-stream/src/server.rs
```diff
@@ -92,30 +94,34 @@ impl ConnectionGenerator {
...
         StreamReceiverService {
             connection_generator,
             next,
             account_type: PhantomData,
+            store: Arc::new(store),
         }
     }
 }
 
 // TODO should this be an OutgoingService instead so the balance logic is applied before this is called?
```
```suggestion
```

crates/interledger-stream/src/server.rs
```diff
@@ -129,13 +135,35 @@ where
...
+                    request.to.username(),
+                    &format!(
+                        "{}:{}:{}",
+                        from_address,
+                        from_name,
+                        request.prepare.amount()
+                    ),
+                ));
+                trace!(
+                    "Sending money: {} ({}) -> {} ${}",
```
emschwartz: This is log line is confusing because this function is receiving rather than sending money

crates/interledger-stream/src/server.rs
```diff
@@ -129,13 +135,35 @@ where
...
+        let from_address = request.from.client_address();
         let dest: &[u8] = destination.as_ref();
-        if dest.starts_with(to.as_ref()) {
+        if dest.starts_with(to_address.as_ref()) {
             if let Ok(shared_secret) = self.connection_generator.rederive_secret(&destination) {
-                {
-                    return Box::new(result(receive_money(&shared_secret, &to, request.prepare)));
-                }
+                // TODO: NAMESPACES! redis doesn't namespace pub/sub, so multiple nodes on the same redis instance might be sad
+                tokio::spawn(self.store.publish(
```
emschwartz: Why does this need to be spawned?

Also, if we do want to `spawn` something, we can depend on `tokio_executor` instead of the full `tokio` crate

crates/interledger-stream/src/server.rs
```diff
@@ -129,13 +135,35 @@ where
...
+                    request.to.username(),
+                    &format!(
```
emschwartz: Instead of making `store.publish` a generic method and then passing in a string here, it would be more useful to make the store method more specific to what it's doing here and pass it structured data. It would then be up to the store implementation to namespace things correctly, rather than the caller

crates/interledger-store-redis/src/store.rs
bstrie: @emschwartz I'm unclear on what specific limitation this is alluding to, AFAICT publishing works in an asynchronous context and it's only subscribing that lacks dedicated async support. In addition I believe that while issuing a SUBSCRIBE command locks the client into a subscription state, the same is not true of the PUBLISH command, which works the same as any other.

Indeed, the agent that eventually handles subscription broadcasts will have to have one Redis client for each subscription (unavoidable due to the design of Redis), with each client running in its own thread to isolate the blocking (which should hopefully be obviated by a future version of the driver).

crates/interledger-service/src/lib.rs
bstrie: The impl for `()` here is just a quick temporary hack to get the tests passing, since the tests require constructing things that implement PubStore but don't actually test any pubsub functionality. Using an Option as suggested above will obviate the need for this.

crates/interledger-store-memory/src/store.rs
bstrie: It's a common pattern of mine to use `unimplemented!()` when you want to stub out WIP code that successfully typechecks; I just wanted to make sure that everything could be built properly and that the tests passed in the redis store. I would have gone through and implemented this for real were it not for #216 giving me doubt about the future of the memory store. This would be resolved one way or the other before accepting this PR.

crates/interledger-stream/src/server.rs
bstrie: This is currently left in as debug output in case you wanted to run it yourself and see work being performed without having to fire up redis-cli and watch the subscription channel manually. I agree that "send" is the wrong word here, although I'd like to posit as well that there could be a better name for this whole function than `send_request`.

crates/interledger-stream/src/server.rs
bstrie: I used `spawn` here because we don't want the act of publishing to block the server from doing any of its usual duties; correct me if I'm wrong, but without using `spawn` we'd either need to wait until the publish completed successfully in order to continue processing the request, or else we'd need to wait do the publish after the payment receipt is complete which would prevent the task from getting on with the next step of its operation. This lets us do both operations concurrently.

crates/interledger-stream/src/server.rs
bstrie: Good idea, semantically it makes sense to have `fn publish_notification` and `fn publish_some_other_thing` rather than relying on callers to namespace everything properly, but that still leaves the underlying issue which is that we allow multiple nodes to operate on the same redis instance, which currently works due to the use of different logical "databases" for each node, but pub/sub uses its own single namespace which AFAICT is shared among all connections. This means that if node_a and node_b both were running on the same instance and each had a user named "bob", subscribing to one would get notifications about both accounts. Internally we'd need to figure out how the store could itself namespace all publishes with a unique ID of its own (or else limit each node to a single redis instance). Is there any convenient existing unique ID that we could leverage?

crates/interledger-store-redis/src/store.rs
emschwartz: My mistake, you're right that publishing is different than subscribing

crates/interledger-stream/src/server.rs
```diff
@@ -129,13 +135,35 @@ where
...
+                    ),
+                ));
+                trace!(
+                    "Sending money: {} ({}) -> {} ${}",
+                    from_name,
+                    from_address,
+                    to_name,
+                    request.prepare.amount()
+                );
+                return Box::new(result(receive_money(
```
emschwartz: The notification should only be published if `receive_money` returns `Ok`

crates/interledger-stream/src/server.rs
emschwartz: Most `OutgoingService`s are actually used for "sending" requests. However, the STREAM receiver in particular short circuits the normal flow by handling the request locally instead of forwarding it on

crates/interledger-stream/src/server.rs
emschwartz: You're right about the scoping (see [the redis pubsub docs](https://redis.io/topics/pubsub#database-amp-scoping)). That's pretty annoying but I don't think there's anything we can do about it.

We only use different Redis DBs within a single logical instance in the tests, so this shouldn't be a problem for real usage

crates/interledger-stream/src/server.rs
emschwartz: That's fair, I guess we want to return the fulfillment anyway, even if sending the notification fails

bstrie:
@emschwartz I've just pushed an updated version of the publish half that we can get started reviewing while the sub half is on its way.

crates/interledger-stream/src/server.rs
```diff
@@ -92,30 +95,33 @@ impl ConnectionGenerator {
...
     A: Account,
 {
-    pub fn new(server_secret: Bytes, next: O) -> Self {
+    pub fn new(server_secret: Bytes, store: S, next: O) -> Self {
         let connection_generator = ConnectionGenerator::new(server_secret);
         StreamReceiverService {
             connection_generator,
             next,
             account_type: PhantomData,
+            store: Arc::new(store),
```
emschwartz: The way we have most stores implemented, the store object itself wraps an `Arc` thing so that we can just pass in a clone of the `Store` here. Also the `Store` should implement `Clone` (and generally `Send` and `Sync`)

crates/interledger-stream/src/server.rs
```diff
@@ -128,14 +134,31 @@ where
...
-                {
-                    return Box::new(result(receive_money(&shared_secret, &to, request.prepare)));
-                }
+                return Box::new(
+                    result(receive_money(&shared_secret, &to_address, request.prepare)).and_then(
+                        move |val| {
+                            store.publish_payment_notification(
+                                to_name,
+                                from_name,
+                                from_address,
```
emschwartz: I think it's more useful to include the `destination` instead of the `from_address`. The `from_name` tells which account it came from

crates/interledger-stream/src/server.rs
```diff
@@ -458,6 +481,7 @@ mod stream_receiver_service {
 
         let mut service = StreamReceiverService::new(
             server_secret.clone(),
+            DummyStore,
```
```suggestion
            None,
```

crates/interledger-stream/src/server.rs
```diff
@@ -509,6 +533,7 @@ mod stream_receiver_service {
...
```
```suggestion
            None,
```

crates/interledger-service/src/lib.rs
```diff
@@ -120,6 +121,22 @@ pub trait AccountStore {
     >;
 }
 
+/// A trait representing the Publish side of a pub/sub store
+pub trait PubStore {
+    fn publish_payment_notification(
+        self: Arc<Self>,
```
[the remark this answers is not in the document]
bstrie: Only to allow us to do the actual publish step concurrently with the request processing rather than sequentially, which might not be that important (I was reluctant to impose mandatory I/O overhead on every single payment packet). Admittedly it's an unfortunate leak of implementation details...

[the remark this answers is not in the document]
emschwartz: Why wouldn't you be able to use it the same way with a `&self` reference?

crates/interledger-service/src/lib.rs
```diff
@@ -120,6 +121,22 @@ pub trait AccountStore {
...
+        _sender: Username,
+        _receiver: Username,
+        _sender_address: Address,
```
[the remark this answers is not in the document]
bstrie: With the assumption that if someone has an account on a node that they're authenticated to subscribe to notifications on, then they can also look up that account's address via its user ID on that same node if they desire to. Conversely I wasn't sure how much information is publicly accessible on any given node, so I wasn't sure if someone receiving a payment from a remote node would be able to look up the payer's address using only the username. If you have a concrete list of details that you think we should be logging then I'd be happy to have that.

crates/interledger-service/src/lib.rs
```diff
@@ -120,6 +121,22 @@ pub trait AccountStore {
...
+        _amount: u64,
+        _timestamp: i64,
```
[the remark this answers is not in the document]
bstrie: We can't make it a SystemTime directly because that type is deliberately opaque and thus doesn't implement Display or ToString etc, but we can use Duration to get the seconds or millis relative to the epoch that any SystemTime represents. Do we mind if this is only at the resolution of a second?

crates/interledger-service/src/lib.rs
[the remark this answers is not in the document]
emschwartz: That sender address isn't going to be the address of the actual packet sender, because that might be on another node. The `from_name` will tell you which account the packet was received through. The `receiver_address` includes the segments added to the address by the STREAM receiver.

For now, I would include `from_username`, `to_username`, `prepare_destination`, `amount`, `timestamp`.

crates/interledger-service/src/lib.rs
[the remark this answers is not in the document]
emschwartz: Look at how this is implemented for the ILP `Prepare` packet ([here](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-packet/src/packet.rs#L122) and [here](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-packet/src/packet.rs#L246-L249)).

Originally, I used `chrono` objects but @sentientwaffle made a good point that it's better to have the interfaces between things use `SystemTimes`, while the crates themselves can use `chrono` for parsing and displaying if needed.

crates/interledger-store-memory/src/store.rs
```diff
@@ -138,6 +138,8 @@ impl AccountStore for InMemoryStore {
     }
 }
 
+impl PubStore for InMemoryStore {}
```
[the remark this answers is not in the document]
bstrie: Because the `interledger` crate still uses the memory store in some places, and so without implementing PubStore for InMemoryStore the repo can't compile.

crates/interledger-service/src/lib.rs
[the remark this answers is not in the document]
bstrie: Because `tokio::spawn` is like `thread::spawn` in that the lifetimes of parent and child have to be independent, so the spawned task can't take `&self`.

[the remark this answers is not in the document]
emschwartz: Look at how this is implemented elsewhere in the repo, for example in the `interledger-api` ([here](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-api/src/routes/accounts.rs#L127)). The store methods take `&self` while the `store` object wraps an `Arc` or something that we can `Clone`. If we need to pass owned versions into different closures (mostly needed because we aren't yet using async/await), we call `clone` on the `store` and pass that into the closure

crates/interledger-service/src/lib.rs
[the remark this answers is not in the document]
bstrie: I'm happy to avoid adding further dependency on chrono, although in this case we'll need the interface to be something that can be stored in Redis so we'll have to convert the SystemTime to something relative to the epoch, and we won't be able to use that number to create a SystemTime on the other side.

crates/interledger-store-redis/src/store.rs
```diff
@@ -640,6 +656,24 @@ impl AccountStore for RedisStore {
...
+impl PubStore for RedisStore {
+    fn publish_payment_notification(
+        self: Arc<Self>,
+        receiver: Username,
+        sender: Username,
+        sender_address: Address,
+        amount: u64,
+        timestamp: i64,
+    ) {
+        let channel = format!("p:{}", receiver);
```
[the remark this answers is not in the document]
bstrie: This is just a premature optimization of mine to avoid sending a few extra bytes with every packet receipt. The `p:` here exists to namespace these notifications in Redis in the event that we want to publish some other type of thing someday.

crates/interledger-store-redis/src/store.rs
```diff
@@ -640,6 +656,24 @@ impl AccountStore for RedisStore {
...
+        let message = format!("{}:{}:{}:{}", sender, sender_address, amount, timestamp);
```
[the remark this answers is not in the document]
bstrie: Sure, although note that the data we're storing here isn't accessible to end-users, so the subscription side could be the one to do the work of translating from an offset to a proper datetime format.

crates/interledger-stream/src/server.rs
```diff
@@ -92,30 +95,33 @@ impl ConnectionGenerator {
 ///
 /// This does not currently support handling data sent via STREAM.
 #[derive(Clone)]
-pub struct StreamReceiverService<O: OutgoingService<A>, A: Account> {
+pub struct StreamReceiverService<S, O: OutgoingService<A>, A: Account> {
     connection_generator: ConnectionGenerator,
     next: O,
     account_type: PhantomData<A>,
+    store: Arc<S>,
```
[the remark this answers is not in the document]
bstrie: Storing an `Option` here (and instantiating all the mocked stores with `None`) would still require implementations on all the mocked stores, because the type system doesn't know that we're mocking and assumes that `Some`s are possible.

crates/interledger-stream/src/server.rs
bstrie: Ah ok, I suppose this is the `self.connection.as_ref().clone()` that I see all around. I'll transition to that (I think I got this current approach from the HttpStore).

crates/interledger-service/src/lib.rs
```diff
@@ -120,6 +121,22 @@ pub trait AccountStore {
     >;
 }
 
+/// A trait representing the Publish side of a pub/sub store
+pub trait PubStore {
```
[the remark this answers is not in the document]
bstrie: I don't mind defining it in interledger-stream instead, it'll just mean adding dependencies on that crate to the store-* crates.

crates/interledger-service/src/lib.rs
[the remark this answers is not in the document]
emschwartz: I would just send Redis an rfc 3339 timestamp as a string

crates/interledger-store-redis/src/store.rs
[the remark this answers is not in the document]
emschwartz: Better not to prematurely optimize

bstrie:
Per our new guidelines I'll close this and re-open on its own branch in this repo.

PR #269, opened 2019-09-02
[WIP] DO NOT MERGE: Distribute docker images and allow examples to use docker

### LINKED ISSUE #179: Distribute pre-built binaries and/or allow examples to use Docker
Compiling the whole project is a relatively slow part of running the examples (as mentioned in https://github.com/emschwartz/interledger-rs/pull/157#issuecomment-515553486). 

To improve the developer experience, we should either distribute pre-built binaries that can be downloaded and used with the examples, or we should make it so the examples can be run in either Docker or "native" mode. The latter option would change the command for running the node from `cargo run...` to `docker run...` based on an environment variable or a flag passed into the script.

dora-gt:
WIP, Fixes #179 Distribute pre-built binaries and/or allow examples to use Docker

**This should be merged after #236 is merged**

- Docker files are provided (`node.dockerfile`, `eth-se.dockerfile`)
    - based on `Dockerfile`
- Docker images will be available on [Docker Hub](https://cloud.docker.com/u/interledgerrs/repository/list)
    - Have to turn on Docker Hub builds for both `.dockerfile`
- Examples use `docker` command to run `node` or `ethereum-ledger`
- Don't use `docker-compose` because these are examples and the main objective is to train readers. If we use `docker-compose`, it is too easy to tell the **real** procedure and the settings.

Questions:
- Should we automatically obtain XRP ledger private key/address pair from [XRP Testnet Faucet](https://xrpl.org/xrp-testnet-faucet.html)? because it seems that the testnet addresses are reset every a few months (then examples will get broken)

TODO:
- [ ] Add an explanation for the docker network because host and port settings got complicated (Souldn't we?)
- [ ] Fix diagrams so that those indicate using Docker


dora-gt:
Close now, will open a draft PR.

PR #270, opened 2019-09-02
[WIP] DO NOT MERGE: Distribute docker images and allow examples to use docker

### LINKED ISSUE #179: Distribute pre-built binaries and/or allow examples to use Docker

dora-gt:
Fixes #179 Distribute pre-built binaries and/or allow examples to use Docker

**This should be merged after #236 is merged**

- Docker files are provided (`node.dockerfile`, `eth-se.dockerfile`)
    - based on `Dockerfile`
- Docker images will be available on [Docker Hub](https://cloud.docker.com/u/interledgerrs/repository/list)
    - Have to turn on [Docker Hub builds](https://docs.docker.com/docker-hub/builds/) for both `.dockerfile`
- Examples use `docker` command to run `node` or `ethereum-ledger`
- Don't use `docker-compose` because these are examples and the main objective is to train readers. If we use `docker-compose`, it is too easy to tell the **real** procedure and the settings.

Questions:
- Should we automatically obtain XRP ledger private key/address pair from [XRP Testnet Faucet](https://xrpl.org/xrp-testnet-faucet.html)? because it seems that the testnet addresses are reset every a few months (then examples will get broken)
    - That said, this could be detectable if we write `README` tests.
- Should `node.dockerfile` be named `Dockerfile`?


eth-se.dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+# Expose ports for HTTP
+# - 3000: HTTP - settlement
+EXPOSE 3000
+
+# To save the node's data across runs, mount a volume called "/data".
+# You can do this by adding the option `-v data-volume-name:/data`
+# when calling `docker run`.
+
+VOLUME [ "/data" ]
```
emschwartz: Doesn't this also need to connect to a Redis instance? (Or are we going to connect it to an instance that lives outside the container? If so, I'm not sure we need this volume)

examples/eth-settlement/README.md
```diff
@@ -8,169 +8,201 @@ This example shows how to configure Interledger.rs nodes and use an Ethereum net
 
 ## Prerequisites
 
-- [Rust](#rust)
+- [Docker](#docker)
 - [An Ethereum network](#an-ethereum-network) to connect to
-- [Redis](#redis)
 
-### Rust
+### Docker
```
emschwartz: I think we should try to make it so the examples can either be run with Docker or without. Some people may want to get more hands-on with the Rust code and would prefer not to interact with it through Docker

examples/eth-settlement/README.md
```diff
@@ -8,169 +8,201 @@ This example shows how to configure Interledger.rs nodes and use an Ethereum net
...
-if lsof -tPi :3001 ; then
-    kill `lsof -tPi :3001`
+### 1. Set up Docker
+First, let's set up a docker network so that containers could connect each other.
+
+<!--!
+NETWORK_ID=`docker network ls -f "name=interledger" --format="{{.ID}}"`
+if [ -z "${NETWORK_ID}" ]; then
+    printf "Creating a docker network...\n"
+    docker network create interledger
```
emschwartz: Won't this be called twice?

eth-se.dockerfile
gakonst: I believe we should package it with Redis (at least for this release).

dora-gt: Sorry, I just forgot to remove this. We don't need this because Redis is on another container.

@gakonst, What do you mean by package? Do you mean squashing Redis into SE's container?
Is your image like this?

- Node Container
    - Interledger Node
    - Redis for Interledger Node
- ETH SE Container
    - ETH SE
    - Redis for ETH SE

examples/eth-settlement/README.md
dora-gt: Hmm yes, indeed. I'm not sure I can cleanly realize it but anyway I'll try.

examples/eth-settlement/README.md
dora-gt: On my environment, it won't. Why do you think so?

eth-se.dockerfile
gakonst: I'd prefer it like that I think. In the future, that would mean we could have the Interledger node and the SE running a different type fo database. For now the current approach might be OK

dora-gt: I see. I intended that these `.dockerfile` s are used for examples and for normal uses, so in the latter case, I'm not sure we should package it with Redis because it could be SQL or something.

So what about mentioning it in the document.

examples/eth-settlement/README.md
dora-gt: I think there are 3 options,

1. Insert both sub-sections.
    1. for non-Docker users
    1. for Docker users
1. Provide both sections
    1. for non-Docker users
    1. for Docker users
1. Provide separate documents

Options 2 and 3 contain a lot of duplicated areas, so I'll try option 1 first. If that looks unexpectedly hard to read, I'll consider option 2 or 3.

dora-gt: hmmmmm, Option 1 looks not so good... this is confusing and hard to read.

I feel like option 3. Although this could increase the cost of maintenance, I think this is better for readers. If we prioritize DX, we should adopt option 3. I'll try option 3.

image

eth-se.dockerfile
dora-gt: I mentioned like this:
https://github.com/interledger-rs/interledger-rs/pull/270/files#diff-f32e61e560deb8bb8ef2b67caf3dfc1dR75

examples/eth-settlement/README.md
dora-gt: Now the directory structure looks like:
```
├── README.md
├── eth-settlement
│   ├── README.md
│   ├── procedure-docker.md
│   └── procedure.md
├── eth_xrp_three_nodes
│   ├── README.md
│   ├── procedure-docker.md
│   └── procedure.md
├── simple
│   ├── README.md
│   ├── procedure-docker.md
│   └── procedure.md
```
Each example `procedure.md` which is for non docker users, and `procedure-docker.md` which is for docker users. These 2 are linked from `README.md`s.

examples/eth-settlement/README.md
```diff
@@ -1,491 +1,23 @@
...
-```
-
-Now two nodes and its settlement engines are set and accounts for each node are also set up.
-
-Notice how we use Alice's settlement engine endpoint while registering Bob. This means that whenever Alice interacts with Bob's account, she'll use that Settlement Engine.
-
-The `settle_threshold` and `settle_to` parameters control when settlements are triggered. The node will send a settlement when an account's balance reaches the `settle_threshold`, and it will settle for `balance - settle_to`.
-
-### 7. Sending a Payment
+- [Compile from the source code and deploy](./procedure.md)⚙
```
emschwartz: If we offer both, we should make one of them the default so that you only need to click over to a different file if you're not using the default method

examples/eth-settlement/images/overview-docker.svg
```diff
@@ -0,0 +1,202 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
emschwartz: It would be helpful to show what is in each docker container, since there are multiple containers running

examples/eth-settlement/procedure-docker.md
```diff
@@ -0,0 +1,492 @@
...
+Advanced: You can run this against the Rinkeby Testnet by running a node that connects to Rinkeby (e.g. `geth --rinkeby --syncmode "light"`) or use a third-party node provider such as [Infura](https://infura.io/). You must also [create a wallet](https://www.myetherwallet.com/) and then obtain funds via the [Rinkeby Faucet](https://faucet.rinkeby.io/).
+
+## Instructions
+
+<!--!
+docker --version > /dev/null || printf "\e[31mUh oh! You need to install Docker before running this example\e[m\n"
+
+printf "Stopping Interledger nodes\n"
+
+REDIS_CONTAINER_ID=`docker ps -q -f "name=redis"`
```
emschwartz: We should just have each node and SE use a different Redis instance

examples/eth-settlement/README.md
emschwartz: Do you think it would be possible to have one instruction file and use a bash variable to store the run commands for the node and SE?

The idea would be that if you run `run-md.sh README.md` it would use `cargo run` by default but if you run `USE_DOCKER=1 run-md.sh README.md` it would use the appropriate `docker run...` command instead. I'm imagining that the script would set a local variable like `RUN_NODE` to be either `cargo run...` or `docker run...` depending on that environment variable.

If we could make this work, it seems like this would require the least amount of duplication

examples/eth-settlement/README.md
emschwartz: I would suggest that the non-docker method should be the default

examples/eth-settlement/README.md
dora-gt: Ah, that is what I tried first. The `.md` of the image(https://github.com/interledger-rs/interledger-rs/pull/270#discussion_r321070479) was like that.

I think it is possible but even if we do that, the explanations become too complex because there are many sub-sections.

Or... do you mean just changing command and leave the explanations for non-docker version?

examples/eth-settlement/README.md
dora-gt: Although it depends on the decision of https://github.com/interledger-rs/interledger-rs/pull/270#discussion_r321450828, what about this structure?

```
├── simple
│   ├── README.md (default procedure, add link to docker.md)
│   └── docker.md
```

examples/eth-settlement/procedure-docker.md
dora-gt: Then isn't it natural that `Simple` example has 2 redis, one for Node A and another for Node B? Should I change that too?

dora-gt: I change so for now, to make it consistent with the other examples.

examples/eth-settlement/images/overview-docker.svg
dora-gt: I agree but actually these are almost 1:1. So every box indicates that it is a container. So I think mentioning it on documents would be sufficient. Placing some explanation in the diagram makes it complicated, I think.

image

examples/eth-settlement/README.md
emschwartz: I think it would be fine to only have the non-docker explanation for now. If people find it confusing, we could change it but we don't really want to spend time explaining docker to people that don't know it. We just want to explain how to use the Interledger components and the instructions should be very similar with the exception of docker-specific things like setting up the networking

examples/eth-settlement/images/overview-docker.svg
emschwartz: I see, then maybe it's just the box around it that's confusing because it seems to suggest that everything is running in a single docker container rather than everything as a separate docker container

examples/eth-settlement/procedure-docker.md
emschwartz: I agree with that

examples/eth-settlement/README.md
dora-gt: I see. Explaining docker might be too much. I think we need to place prerequisites on explanations, which means deciding what we should explain and what we won't try to explain. If we define `docker` as prerequisites, that would be fine.

I'll try.

examples/eth-settlement/images/overview-docker.svg
dora-gt: (Resolve this because we no longer need this diagram https://github.com/interledger-rs/interledger-rs/pull/270#discussion_r322049158)

examples/eth-settlement/README.md
dora-gt: (docker version will be merged into non docker version without explanations)

dora-gt:
I think I've addressed most points including Ben's suggestions (#287).
I'm checking if there are any points that additionally I should change.

I'm not sure why I need this:
https://github.com/interledger-rs/interledger-rs/pull/270/files#diff-3301fa36a91ba4f24d42a74c47b0c213R43-R50
but it doesn't work with a pipe. 🤔

examples/eth_xrp_three_nodes/README.md
```diff
@@ -177,33 +249,100 @@ In this example, we'll connect 3 Interledger nodes and each node needs its own s
 Instead of using the LEDGER_ADDRESS and LEDGER_SECRET from the examples below, you can generate your own XRPL credentials at the [official faucet](https://xrpl.org/xrp-test-net-faucet.html).
 
 <!--!
-which ilp-settlement-xrp &> /dev/null || error_and_exit "You need to install \"ilp-settlement-xrp\"."
+printf "\nStarting settlement engines...\n"
+if [ "$USE_DOCKER" -eq 1 ]; then
+    # Start Alice's settlement engine (ETH)
+    docker start interledger-rs-se_a ||
```
emschwartz: nit: this way of starting the nodes prints a lot of errors in the terminal such as the following. It would be great to make sure it doesn't print that:

```
Error response from daemon: No such container: redis-alice_node
Error: failed to start containers: redis-alice_node
```

node.dockerfile
```diff
@@ -0,0 +1,39 @@
+# Build Interledger node into standalone binary
```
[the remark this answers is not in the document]
emschwartz: Also, if we do have a `docker` directory, can you also move the `redis.conf` and `run-node-and-redis.sh` files into there too?

bstrie:
@emschwartz Was this closed by accident?

dora-gt:
It seems to be an accident but I can't reopen this (Idk why), so I'll open another PR later.

dora-gt:
It seems to be an accident but I can't reopen this (maybe because `feature/cli-improvement/main` branch was deleted as the branch was merged into the master), so I'll open another PR later.

scripts/run-md.sh
```diff
@@ -9,6 +9,8 @@
 # Markdown file but run them when the whole file is executed.
 # This is largely intended for commands that make the output more readable
 
+TMP_SCRIPT=$(mktemp -t "md.sh")
```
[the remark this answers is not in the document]
dora-gt: Oops, I didn't notice because it works on my environment.

examples/eth-settlement/README.md
```diff
@@ -36,7 +36,7 @@ Advanced: You can run this against the Rinkeby Testnet by running a node that co
 Then you should be able to use `npm`. To install `ganache-cli`, run `npm install -g ganache-cli`.
 
 ### Redis
-The Interledger.rs nodes currently use [Redis](https://redis.io/) to store their data (SQL database support coming soon!)
+The Interledger.rs nodes and settlement engines currently use [Redis](https://redis.io/) to store their data (SQL database support coming soon!). Nodes and settlement engines don't need to share the Redis necessarily.
```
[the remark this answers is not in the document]
dora-gt: Fixed

scripts/run-md.sh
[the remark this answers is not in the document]
dora-gt: Fixed. (but I also try to figure out why a pipe doesn't work)

examples/eth-settlement/README.md
```diff
@@ -49,106 +49,277 @@ Make sure your Redis is empty. You could run `redis-cli flushall` to clear all t
...
-    kill `lsof -tPi :3000`
-fi
+printf "Stopping Interledger nodes\n"
 
-if lsof -tPi :3001 ; then
-    kill `lsof -tPi :3001`
+if [ "$USE_DOCKER" -eq 1 ]; then
+    docker --version > /dev/null || error_and_exit "Uh oh! You need to install Docker before running this example"
+    
+    docker stop \
```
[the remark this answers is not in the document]
dora-gt: Fixed.

node.dockerfile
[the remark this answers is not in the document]
dora-gt: > Should we move both of these to a docker folder?

Actually I thought it would be better because currently the top directory looks a bit messy.
If you don't mind, I would like to do so.

image

We can flexibly automate builds. That should be fine.

I'm wondering if having these 2 dockerfiles is confusing.

- Docker
- node.dockerfile

I'm not sure we want a packaged image like the current `Dockerfile`.

examples/eth-settlement/README.md
```diff
@@ -158,46 +329,116 @@ ILP_REDIS_URL=redis://127.0.0.1:6379/0 \
...
 ILP_SETTLEMENT_API_BIND_ADDRESS=127.0.0.1:7771 \
-cargo run --package interledger -- node &> logs/node-alice.log &
+cargo run --bin interledger -- node &> logs/node-alice.log &
 
 # Start Bob's node
 ILP_ADDRESS=example.bob \
 ILP_SECRET_SEED=1604966725982139900555208458637022875563691455429373719368053354 \
 ILP_ADMIN_AUTH_TOKEN=hi_bob \
-ILP_REDIS_URL=redis://127.0.0.1:6379/1 \
+ILP_REDIS_URL=redis://127.0.0.1:6381/0 \
```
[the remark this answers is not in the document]
dora-gt: Fixed.

examples/eth_xrp_three_nodes/README.md
dora-gt: Yeah... it looks not so nice but I hesitated to suppress the outputs because sometimes it fails because of the other reasons than `No such container`.

I'll try to figure out a better way.

dora-gt:
New PR is here: #293

examples/eth_xrp_three_nodes/README.md
dora-gt: I realized that we didn't need to do this because the containers are always cleaned. Now it must look better.

Issue #258, opened 2019-08-30
Should we change any of the HTTP API endpoints?


emschwartz:
For example:
- `POST /pay` is not very RESTful. A more RESTful version might look something like `POST /payments` or `PUT /payments/:id` or `PUT /accounts/:username/payments/:payment_id` (is the payment a resource created under the account that you could then query later?)
- We might want to change `GET /spsp/:username` to `GET /accounts/:username/spsp`, since the SPSP details are a property of that account
- Should the balance endpoint be separate from the rest of the account details?

bstrie:
> Should the balance endpoint be separate from the rest of the account details?

We could have a general endpoint for `GET /accounts/:username/:property`, which would give us `GET /accounts/:username/balance/`, plus e.g. `GET /accounts/:username/ilp_address`, `GET /accounts/:username/asset_code`, etc. We'd want to whitelist the properties visible to non-admins, of course. `GET /accounts/:username` could still get the full information for an account.

> We might want to change GET /spsp/:username to GET /accounts/:username/spsp, since the SPSP details are a property of that account

With the above suggestion, `GET /accounts/:username/spsp_address` would be natural.

> A more RESTful version might look something like...

...or, if we're feeling clever, `PUT /accounts/:username/balance`. :)

gakonst:
I agree on the generic endpoint for each properties, did some initial work for it in #271. Which properties do we want to have available publicly?

I've changed /pay to /payments, and combined the balances in the account's details.

emschwartz:
Why would `GET /accounts/:username/property` be more useful than a `GET /accounts/:username` that returns a JSON object with all of the properties? It seems like there's a decent chance that you'd want to see more than one property at a time. As for whether we should support both, I'd prefer to only have one if adding multiple methods doesn't add much new functionality.

emschwartz:
> ...or, if we're feeling clever, PUT /accounts/:username/balance. :)

That wouldn't work because you'd need to `PUT` the total amount, rather than the delta, which would be a nightmare if you have different processes making payments at the same time.

gakonst:
> Why would `GET /accounts/:username/property` be more useful than a `GET /accounts/:username` that returns a JSON object with all of the properties? It seems like there's a decent chance that you'd want to see more than one property at a time. As for whether we should support both, I'd prefer to only have one if adding multiple methods doesn't add much new functionality.

[...]

PR #271, opened 2019-09-02
Restful improvements

gakonst:
Part of #258 

- combines balances into account details
- slightly more restful endpoints
- adds `/:username/:property` endpoint to have more fine grained control over getting account details


crates/interledger-api/src/routes/spsp.rs
```diff
@@ -127,12 +127,12 @@ impl_web! {
 
         // TODO resolve payment pointers with subdomains to the correct account
         // also give accounts aliases to use in the payment pointer instead of the ids
-        #[get("/.well-known/pay")]
+        #[get("/.well-known/payments")]
```
emschwartz: This cannot be changed because it is defined as this in the SPSP spec

crates/interledger-api/src/routes/spsp.rs
```diff
@@ -59,7 +59,7 @@ impl_web! {
             self
         }
 
-        #[post("/pay")]
+        #[post("/payments")]
```
emschwartz: Should this be under accounts?

crates/interledger-api/src/lib.rs
```diff
@@ -66,6 +66,8 @@ pub struct AccountDetails {
     pub asset_scale: u8,
     #[serde(default = "u64::max_value")]
     pub max_packet_amount: u64,
+    pub balance: Option<i64>,
+    pub prepaid_amount: Option<i64>,
```
emschwartz: I'm not sure we want to expose this implementation detail. In the future, we may want to switch the implementation to use payable and receivable balances instead of the net balance and prepaid amount.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -217,6 +217,72 @@ impl_web! {
             })
         }
 
+        #[get("/accounts/:username/:property")]
```
emschwartz: I'm not sure how useful it is to add this to the API

crates/interledger-settlement-engines/tests/eth_ledger_settlement.rs
```diff
@@ -86,6 +86,8 @@ fn eth_ledger_settlement() {
                         http_incoming_token: Some("in_alice".to_string()),
                         http_outgoing_token: None,
                         max_packet_amount: 10,
+                        balance: None,
```
emschwartz: Hmm this seems strange and makes me think that the balance does not actually belong on the account details

crates/interledger-store-redis/src/account.rs
```diff
@@ -254,6 +258,10 @@ impl ToRedisArgs for AccountWithEncryptedTokens {
             .write_redis_args(&mut rv);
         "round_trip_time".write_redis_args(&mut rv);
         account.round_trip_time.write_redis_args(&mut rv);
+        "balance".write_redis_args(&mut rv);
+        account.balance.write_redis_args(&mut rv);
```
emschwartz: The balance should not be updated like this

gakonst: Why not? If the balance becomes a field of the account, then it should be part of the ToRedisArgs implementation. What we could do, is encrypt the field, in the same fashion as the auth tokens.

crates/interledger-api/src/routes/accounts.rs
gakonst: It's an API endpoint which gives more flexibility on the returned info, as per https://github.com/interledger-rs/interledger-rs/issues/258#issuecomment-526708766. It might be better to make it `/accounts/:username?properties=ilp_address,balance`

crates/interledger-api/src/routes/spsp.rs
gakonst: We could make it to `POST /accounts/:username/pay/:amount`. In that case the word `pay` is better IMO since you are making an action in which `username` is the receiver. `payments` makes more sense if you are a user and you want to check your past payments

gakonst: We can definitely make it `POST /accounts/payments`.

We could also try converting it to `POST /accounts/:receiver/pay/:amount`. However, the receiver is a full URL, eg `https://localhost:7770/accounts/alice/spsp`. Putting a URL inside of a URL clearly isn't very nice. Would it be possible to fetch all the receiver information from the store, allowing us to make it `POST /accounts/:username/pay/:amount`? I think this is not possible e.g in case we want to send a payment to an account that the connector does not have in their store.

crates/interledger-api/src/routes/accounts.rs
bstrie: I've left some comments in #258 regarding this question.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -217,6 +217,72 @@ impl_web! {
...
+        fn http_get_property(&self, username: String, property: String, authorization: String) -> impl Future<Item = Value, Error = Response<()>> {
+            let store = self.store.clone();
+            let self_clone = self.clone();
+            let is_admin = self.is_admin(&authorization);
+            let username_clone = username.clone();
+            let auth_clone = authorization.clone();
+            result(Username::from_str(&username))
+            .map_err(move |_| {
+                error!("Invalid username: {}", username);
+                Response::builder().status(500).body(()).unwrap()
```
bstrie: This common boilerplate can be replaced with the `Response::error` helper function defined and used elsewhere in this file.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -217,6 +217,72 @@ impl_web! {
...
         }
 
+        #[get("/accounts/:username/:property")]
+        #[content_type("application/json")]
+        fn http_get_property(&self, username: String, property: String, authorization: String) -> impl Future<Item = Value, Error = Response<()>> {
+            let store = self.store.clone();
+            let self_clone = self.clone();
+            let is_admin = self.is_admin(&authorization);
+            let username_clone = username.clone();
+            let auth_clone = authorization.clone();
```
bstrie: Looking elsewhere in this file a see a few other `_clone` variables, and while some of them are definitely necessary I think others can be completely removed. But I might just go through and do that myself as a separate PR to get the whole file at once.

gakonst: I think we can go through the codebase and remove all the unnecessary clones once we migrate to async/await. They are a big bi-product of the nested closures of futures.

crates/interledger-api/src/lib.rs
gakonst: It is no longer exposed as of https://github.com/interledger-rs/interledger-rs/pull/271/commits/8344a794317175fef48401664b059c2374621b6e

crates/interledger-api/src/routes/accounts.rs
gakonst: The query string approach has been implemented in https://github.com/interledger-rs/interledger-rs/pull/271/commits/27cb9ba72b72ed9615b0b27a031c1af0fb0874b9.

bstrie:
@kincaidoneil There's additional discussion in #258 , TL;DR I think that having fine-grained property access has some advantages to backwards compatibility, interoperability of implementations, and debugability (e.g. accidentally requesting an admin-only property without admin permissions can 401 only a specific request rather than the entire request). There's also the concern that we might do something like record every received payment packet in the account info someday, which could cause the response to grow without bound.

gakonst:
I mean, as implemented currently, it's the same endpoint as before, it just optionally takes a query string parameter to filter for the desired information. Don't see any harm with adding this functionality.

crates/interledger-api/src/routes/spsp.rs
emschwartz: I think we should make it `POST /accounts/:username/payments`. In the future, we can add a payment ID to the response that would allow you to query the details of payments sent by calling `GET /accounts/:username/payments/:payment_id` or `GET /accounts/:username/payments?limit=100`, etc.

The idea is that by sending this POST request, you are creating a payment resource that belongs to a particular account.

crates/interledger-store-redis/src/account.rs
emschwartz: It seems dangerous to allow overwriting the balance just by updating the account. Even if the admin wanted to override the balance it seems like it would probably be safer to atomically add/subtract from the balance rather than mess with the numbers while payments could be in flight.

The concern isn't about the privacy of the balance, so I wouldn't bother encrypting it

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -158,11 +163,17 @@ impl_web! {
 
         #[get("/accounts/:username")]
         #[content_type("application/json")]
-        fn http_get_account(&self, username: String, authorization: String) -> impl Future<Item = Value, Error = Response<()>> {
+        fn http_get_account(&self, username: String, query_string: Option<QueryString>, authorization: String) -> impl Future<Item = Value, Error = Response<()>> {
```
emschwartz: This seems like a bit of a premature optimization, because the account details aren't that big so it's not like you're saving a lot of data.

> I'm a little unclear, what's the rationale for this? I'm not sure this is a common pattern, and it's probably much simpler for clients to just use the fields they want rather than having to explicitly request particular fields

I would tend to agree with @kincaidoneil here. It's not a _very complicated_ feature to include but I don't think many APIs use that style and it doesn't seem like it adds much over simply extracting the details you need from the JSON response body.

Regarding @bstrie's points in https://github.com/interledger-rs/interledger-rs/pull/271#issuecomment-528006445:
> TL;DR I think that having fine-grained property access has some advantages to backwards compatibility, interoperability of implementations, and debugability

I'm not sure it helps with any of these qualities. We also support just getting the whole account object. If you have a statically typed parser for the response, it should be configured to ignore additional properties so that we can add fields to this without breaking the API clients.

> There's also the concern that we might do something like record every received payment packet in the account info someday, which could cause the response to grow without bound.

Received payments would not be part of the account record. They would be queried using something like `GET /accounts/:username/received_payments?limit=100`.

emschwartz:
Closing this for now. We can reopen this or do another PR to make these changes if we want later

PR #272, opened 2019-09-02
Modify Account Settings

### LINKED ISSUE #229: Allow users to modify certain account settings?
Right now, most of the HTTP API endpoints are admin-only. Should we allow users to modify certain settings on their account?

Some fields we may want to make modifiable:
- User password / API key?
- Username?
- Settlement details (came up in https://github.com/interledger-rs/interledger-rs/issues/121)
- HTTP or BTP connection details

Obviously fields like balances and limits must not be modifiable.

gakonst:
Should close #229

### DIFF
--- a/crates/interledger-api/src/lib.rs
+++ b/crates/interledger-api/src/lib.rs
@@ -41,6 +41,12 @@ pub trait NodeStore: Clone + Send + Sync + 'static {
         account: AccountDetails,
     ) -> Box<dyn Future<Item = Self::Account, Error = ()> + Send>;
 
+    fn modify_account_settings(
+        &self,
+        id: <Self::Account as AccountTrait>::AccountId,
+        settings: AccountSettings,
+    ) -> Box<dyn Future<Item = Self::Account, Error = ()> + Send>;
+
     // TODO limit the number of results and page through them
     fn get_all_accounts(&self) -> Box<dyn Future<Item = Vec<Self::Account>, Error = ()> + Send>;
 
@@ -57,6 +63,26 @@ pub trait NodeStore: Clone + Send + Sync + 'static {
         prefix: String,
         account_id: <Self::Account as AccountTrait>::AccountId,
     ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
+}
+
+/// AccountSettings is a subset of the user parameters defined in
+/// AccountDetails. Its purpose is to allow a user to modify certain of their
+/// parameters which they may want to re-configure in the future, such as their
+/// tokens (which act as passwords), their settlement frequency preferences, or
+/// their HTTP/BTP endpoints, since they may change their network configuration.
+#[derive(Debug, Extract, Response, Clone, Default)]
+pub struct AccountSettings {
+    pub http_incoming_token: Option<String>,
+    pub btp_incoming_token: Option<String>,
+    pub http_outgoing_token: Option<String>,
+    pub btp_outgoing_token: Option<String>,
+    pub http_endpoint: Option<String>,
+    pub btp_uri: Option<String>,
+    pub settle_threshold: Option<i64>,
+    // Note that this is intentionally an unsigned integer because users should
+    // not be able to set the settle_to value to be negative (meaning the node
+    // would pre-fund with the user)
+    pub settle_to: Option<u64>,
 }
 
 /// The Account type for the RedisStore.
--- a/crates/interledger-store-redis/src/store.rs
+++ b/crates/interledger-store-redis/src/store.rs
@@ -16,7 +16,7 @@
 //    hgetall <key>         the flattened list of every key/value entry within a hash
 
 use super::account::*;
-use super::crypto::{generate_keys, DecryptionKey, EncryptionKey};
+use super::crypto::{encrypt_token, generate_keys, DecryptionKey, EncryptionKey};
 use bytes::Bytes;
 use futures::{
     future::{err, ok, result, Either},
@@ -27,7 +27,7 @@ use std::collections::{HashMap, HashSet};
 
 use super::account::AccountId;
 use http::StatusCode;
-use interledger_api::{AccountDetails, NodeStore};
+use interledger_api::{AccountDetails, AccountSettings, NodeStore};
 use interledger_btp::BtpStore;
 use interledger_ccp::RouteManagerStore;
 use interledger_http::HttpStore;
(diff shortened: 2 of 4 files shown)


crates/interledger-api/src/lib.rs
```diff
@@ -39,6 +39,12 @@ pub trait NodeStore: Clone + Send + Sync + 'static {
         account: AccountDetails,
     ) -> Box<dyn Future<Item = Self::Account, Error = ()> + Send>;
 
+    fn modify_account_settings(
+        &self,
+        id: <Self::Account as AccountTrait>::AccountId,
+        settings: AccountSettings,
+    ) -> Box<dyn Future<Item = Self::Account, Error = ()> + Send>;
+
```
bstrie: Does this need to exist separately from `update_account` above? Theoretically we could devise a way to make the current `update_account` agnostic over whether it's being called by an admin or a user, and then re-use the logic for both purposes. In general I think this might be part of a larger discussion about the internal structure of AccountDetails (take for instance Evan's note on the other PR that it sometimes feels weird to see `balance` included within it).

gakonst: Allowing this function to take `AccountDetails` would mean the user has to provide all of their account details which seems like unhandy API design. As a user I want an API that I send it the parameter I want to update along with my key and it's updated. 

We could make `AccountDetails`' fields `Option`al but that doesn't seem right.

--
(`balance` has been removed from `AccountDetails` from that PR as of https://github.com/interledger-rs/interledger-rs/pull/271/commits/8344a794317175fef48401664b059c2374621b6e.)

bstrie: Indeed, I was thinking of making this function take something like an `AccountDiff`, whose fields could be optional without imposing `Option` on `AccountDetails` itself, and then implementing `Into<AccountDiff>` for `AccountDetails` and `AccountSettings` (which in this design might want to be named `LimitedAccountDetails` or something). The end result would be a new intermediate struct, but it would allow sharing the code for both functions (and in the process could improve the admin updating operation to only update those fields that have actually changed, rather than having it recreate every field... today it is itself basically a copy-paste of the insertion routine). But in the end I don't mind even more copy-pasting (in this case a copy-paste of a copy-paste), although we should register it as something to look at improving later.

gakonst: `impl From<AccountDetails> for AccountSettings` is doable, since `AccountSettings` is a subset of `AccountDetails`. Going from `AccountSettings` to `AccountDetails` should not be possible. 

It seems to me that Tower Web does not support this type of functionality though, or wrapping the 2 types with an Enum..

```rust
#[macro_use]
extern crate tower_web;
use tower_web::Extract;

#[derive(Extract, Debug)]
struct Bigger {
    x: u64,
    y: u64,
}

#[derive(Extract, Debug)]
struct Smaller {
    y: u64,
}

impl From<Bigger> for Smaller {
    fn from(src: Bigger) -> Self {
        Smaller { y: src.y }
    }
}

impl_web! {
    #[derive(Clone)]
    pub struct MyStruct;
    impl MyStruct {

        #[get("/")]
        fn hello_world<T: Into<Smaller>>(&self, body: T) -> Result<String, ()> {
        // fn hello_world(&self, body: impl Into<Smaller>) -> Result<String, ()> {
            let s: Smaller = body.into();
            Ok(format!("{:?}", s))
        }
    }
}
```

bstrie: Sure, I'd be fine with cleaning this up in the future. Just something to keep in mind regarding the overall structure of AccountDetails and friends.

crates/interledger-api/src/lib.rs
```diff
@@ -57,6 +63,14 @@ pub trait NodeStore: Clone + Send + Sync + 'static {
     ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
 }
 
+#[derive(Debug, Extract, Response, Clone)]
+pub struct AccountSettings {
+    pub http_incoming_token: Option<String>,
+    pub btp_incoming_token: Option<String>,
+    pub settle_threshold: Option<i64>,
+    pub settle_to: Option<u64>,
```
emschwartz:
```suggestion
    // Note that this is intentionally an unsigned integer because users should
    // not be able to set the settle_to value to be negative (meaning the node
    // would pre-fund with the user)
    pub settle_to: Option<u64>,
```

crates/interledger-store-redis/src/store.rs
```diff
@@ -454,52 +454,103 @@ impl RedisStore {
...
+                "http_incoming_token",
+                encrypt_token(&encryption_key.expose_secret().0, token.as_ref()).as_ref(),
+            );
+        }
+
+        if let Some(settle_threshold) = settings.settle_threshold {
+            pipe.hset(accounts_key(id), "settle_threshold", settle_threshold);
+        }
+
+        if let Some(settle_to) = settings.settle_to {
```
emschwartz: We should make sure this does not overflow the max i64 (that's all that redis can handle). I'm not sure what would happen if you set it to a value that's too large, but we probably don't want that (and should have a test to ensure that even if the user does set it to the max u64 value, nothing bad happens)

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -82,6 +82,32 @@ fn update_accounts() {
...
+            settle_threshold: Some(80),
+            settle_to: Some(10),
+        };
+        let account = accounts[0].clone();
+
+        let id = account.id();
+        store
+            .modify_account_settings(id, settings)
+            .and_then(move |account| {
+                assert_eq!(account.get_http_auth_token(), Some("bob:new_token2"));
```
emschwartz: Why is this `new_token2` rather than `new_token1`?

gakonst: Because `get_btp_token` and `get_http_auth_token` are used to fetch the outgoing tokens. The rest of the account's fields are marked as `pub(crate)` and hence are not visible to the integration tests. We could add functions for `get_incoming_X_auth_token`, but they wouldn't be used anywhere other than the tests.

crates/interledger-store-redis/src/store.rs
gakonst: Added test for that.

crates/interledger-api/src/lib.rs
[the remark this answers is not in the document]
gakonst: Agreed on `http_endpoint`. Should the same logic then apply for `btp_uri`?

re tokens:

**Account A**'s `outgoing` token on **Node B** is **Account B**'s `incoming` token on **Node A**. Since Node A's operator can change Account B's incoming auth token, it should follow that they should also be able to change their outgoing token in node B. 

That said, I agree that we should also allow changing the outgoing tokens.

crates/interledger-api/src/lib.rs
```diff
@@ -59,6 +65,26 @@ pub trait NodeStore: Clone + Send + Sync + 'static {
...
 }
 
+/// AccountSettings is a subset of the user parameters defined in
+/// AccountDetails. Its purpose is to allow a user to modify certain of their
+/// parameters which they may want to re-configure in the future, such as their
+/// tokens (which act as passwords), their settlement frequency preferences, or
+/// their HTTP/BTP endpoints, since they may change their network configuration.
+#[derive(Debug, Extract, Response, Clone, Default)]
+pub struct AccountSettings {
+    pub http_incoming_token: Option<String>,
```
emschwartz: Should all of these secrets be a `Secret` type?

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -82,6 +82,52 @@ fn update_accounts() {
...
+            .and_then(move |ret| {
+                assert_eq!(
+                    account.get_http_auth_token().unwrap(),
+                    ret.get_http_auth_token().unwrap()
+                );
+                assert_eq!(
+                    account.get_btp_token().unwrap(),
+                    ret.get_btp_token().unwrap()
+                );
+                // Cannot check other parameters since they are only pub(crate).
```
emschwartz: :frowning_face:

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -82,6 +82,52 @@ fn update_accounts() {
...
+                let _ = context;
+                Ok(())
+            })
+    }))
+    .unwrap();
+}
+
+use std::default::Default;
+#[test]
+fn modify_account_settings_unchanged() {
```
emschwartz: You should also include a test to make sure that the parameters you do want to update are updated

crates/interledger-api/src/lib.rs
gakonst: They're there just for the API request, they never touch disk so don't think so

PR #274, opened 2019-09-04
fix(examples): update simple example in light of #268

bstrie:


crates/interledger/src/node.rs
```diff
@@ -113,6 +117,7 @@ pub struct InterledgerNode {
     /// When SPSP payments are sent to the root domain, the payment pointer is resolved
     /// to <domain>/.well-known/pay. This value determines which account those payments
     /// will be sent to.
+    #[serde(default = "default_default_spsp_account")]
```
emschwartz: I think `Option` types default to `None`

bstrie: Ah, indeed. :)

Issue #195, opened 2019-08-07
Split the notion of a store into different roles


bstrie:
Currently we simply have an all-encompassing notion of a `store` as a general storage abstraction, but it may be useful to split this out into more well-defined roles such as "persistence layer", "cache", "pub/sub", and then allow different storage implementations to serve each of these roles rather than e.g. forcing Redis to do everything.

gakonst:
Thinking more about this, this refactor should definitely happen in order to abstract away any encryption/decryption of keys from a store implementer.

PR #276, opened 2019-09-05
Get ILP Address from Parent node

### LINKED ISSUE #186: Replace numeric account IDs and configured ILP addresses with usernames and dynamically generated addresses

### LINKED ISSUE #248: Get node ILP Address from parent
Right now the `node` needs to be configured with an ILP address. If you are running a node as a "child" of another, it would be easier and more reliable to automatically query the parent node for the ILP address. (This would also mean that with https://github.com/interledger-rs/interledger-rs/issues/186, the ILP address of each account on the node will be correct without any configuration, whereas right now you can mess it up by using an un-routable address)

- [ ] Make `ilp_address` optional in the `node` configuration
- [ ] If `ilp_address` is not provided, make an ILDCP query to get the address
- [ ] Figure out which account to send the ILDCP query to (Should there be an index of accounts listed as `Parent`s? If there are multiple, which one should we use?)
- [ ] What happens when the node is run with no accounts configured? Should there be a hook on the account creation API that would recognize when a "parent" is added that would send the ILDCP query and update the node's address?

### LINKED ISSUE #285: Connect to BTP server when account is added with a btp_uri field
See TODO [here](https://github.com/interledger-rs/interledger-rs/blob/a7088a41c5a38c3ba8acf34712e6a9e680905be6/crates/interledger-api/src/routes/accounts.rs#L84-L85)

gakonst:
Closes #186. 
Closes #248.
Closes #285

This PR modifies the address assignment procedure for nodes which get added as `Child`ren to a node.

1. `IldcpAccount` is merged with `Account`
1. When an account is added as `Child`, instead of using their provided ILP Address, we append their `username` to our node's ilp address.
1. When an account is added as `Parent` (**only 1 parent is allowed for now**) the API:
    1. Performs an `ILDCP` Request to Parent. The Parent responds with the ILP Address they have assigned the node, and the store's address is updated to that value. This is done so that any subsequent account insertions will take into account the node's updated ilp address
    1. Sends a RouteControlRequest to the Parent, the  Parent responds with a RouteUpdateRequest 
which has all the routes. This is done because when a node is started with a non-routable address, it rejects all incoming routing messages. This required to modify the CCP Service so that it's able to read the store's ILP Address on each broadcast loop, and update its `self.global_prefix`.
    1. Updates all `Child` accounts on the node to use the update node's address. This can be an expensive operation, but we assume that a node operator will not change their parent node's address often.
1. When an account is added, if it has a BTP Url configured, it'll try to connect to that account's BTP Server


crates/interledger-ildcp/src/server.rs
```diff
@@ -30,13 +32,19 @@ where
...
 {
     type Future = BoxedIlpFuture;
 
     fn handle_request(&mut self, request: IncomingRequest<A>) -> Self::Future {
         if is_ildcp_request(&request.prepare) {
-            let from = request.from.client_address();
+            let from = if request.from.routing_relation() == RoutingRelation::Child {
+                self.ilp_address
+                    .with_suffix(request.from.username().as_bytes())
+                    .unwrap()
```
emschwartz: There are other places that we may want to use the ILP address (in SPSP/STREAM, for example), so I think it might be better to include this behavior in the implementation of the `ilp_address` method on the `Account` trait.

The downside of doing this, however, is that we would need to reimplement the behavior in each implementation of `Account` (i.e. for each Store implementation).

crates/interledger-ildcp/src/server.rs
```diff
@@ -51,3 +59,66 @@ where
...
+    use std::str::FromStr;
+    use std::time::SystemTime;
+    use crate::packet::{ILDCP_DESTINATION, PEER_PROTOCOL_CONDITION};
+
+    #[test]
+    fn appends_username() {
+        let mut service = test_service();
+        let result = service
+            .handle_request(IncomingRequest {
+                from: USERNAME_ACC.clone(), // account without an ILP Address configured, implicitly considered a Child
```
emschwartz: nit: this test would be clearer if you just created the `TestAccount` here instead of using a fixture

crates/interledger-service/src/lib.rs
```diff
@@ -52,6 +52,9 @@ pub trait Account: Clone + Send + Sized + Debug {
 
     fn id(&self) -> Self::AccountId;
     fn username(&self) -> &Username;
+    fn client_address(&self) -> &Address;
```
emschwartz: If we're changing this already, I really want to rename this method. `client_address` is what it's called in ILDCP but this goes beyond ILDCP now

```suggestion
    fn ilp_address(&self) -> &Address;
```

crates/interledger-store-redis/src/store.rs
```diff
@@ -286,6 +291,7 @@ impl RedisStoreBuilder {
 /// future versions of it will use PubSub to subscribe to updates.
 #[derive(Clone)]
 pub struct RedisStore {
+    ilp_address: Arc<Address>,
```
emschwartz: Why does this need to be a property on the `RedisStore`?

crates/interledger-store-redis/src/store.rs
```diff
@@ -320,6 +327,9 @@ impl RedisStore {
                     let mut pipe = redis::pipe();
                     pipe.exists(accounts_key(account.id));
                     pipe.hexists("usernames", account.username().as_ref());
+                    if account.routing_relation == RoutingRelation::Parent {
+                        pipe.exists("parent");
```
emschwartz: nit:
```suggestion
                        pipe.exists("parent_node_account_id");
```

crates/interledger-store-redis/src/store.rs
```diff
@@ -346,6 +356,12 @@ impl RedisStore {
                     // Save map for Username -> Account ID
                     pipe.hset("usernames", account.username().as_ref(), id).ignore();
 
+                    if account.routing_relation == RoutingRelation::Parent {
+                        pipe.set("parent", id).ignore();
+                    } else if account.routing_relation == RoutingRelation::Child {
+                        account.ilp_address = node_address.with_suffix(account.username.as_bytes()).unwrap();
```
emschwartz: It seems odd for this to happen in the store when the account is being loaded. I think it would be better to make this a little more dynamic by adding a `configured_ilp_address: Option<Address>` on the Redis `Account` implementation and then to make the `ilp_address` method use the `configured_ilp_address` if it is `Some` or append the `Username` if not.

Also, it feels slightly more flexible to allow any account to have a `configured_ilp_address`, rather than tie this behavior to the `RoutingRelation`.

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -21,7 +20,10 @@ fn insert_accounts() {
             .and_then(move |account| {
                 assert_eq!(
                     *account.client_address(),
-                    Address::from_str("example.charlie").unwrap()
+                    // note that charlie was added as a child, hence his
+                    // username is appended to the node's address instead of
+                    // using any address
+                    Address::from_str("example.address.charlie").unwrap()
```
emschwartz: nit: this would be clearer if the base address used were not `example.address` but something more like `example.node` or `example.bob`

crates/interledger-stream/Cargo.toml
```diff
@@ -30,6 +30,7 @@ ring = "0.14.6"
 [dev-dependencies]
 env_logger = "0.6.1"
 interledger-router = { path = "../interledger-router", version = "0.2.1" }
+interledger-ccp = { path = "../interledger-ccp", version = "0.1" }
```
emschwartz: STREAM should not depend on CCP. Those are protocols at very different parts of the stack

crates/interledger-stream/src/server.rs
```diff
@@ -470,12 +470,14 @@ mod stream_receiver_service {
                     ilp_address: Address::from_str("example.sender").unwrap(),
                     asset_code: "XYZ".to_string(),
                     asset_scale: 9,
+                    routing_relation: RoutingRelation::Peer,
```
emschwartz: Why is this field needed? It doesn't seem to be used anywhere

crates/interledger/src/main.rs
```diff
@@ -118,6 +118,11 @@ pub fn main() {
                                 .long("redis_uri")
                                 .help("Redis database to add the account to")
                                 .default_value("redis://127.0.0.1:6379"),
+                            Arg::with_name("node_ilp_address")
+                                .long("ilp_address")
+                                .help("ILP Address of the node (used in case the account being inserted is a child in order to calculate their address properly)")
```
emschwartz: It should be able to determine this automatically from whatever the node's ILP address is set to

crates/interledger-store-redis/src/store.rs
gakonst: As we discussed on our call, the store will have the node's current ILP Address loaded in memory, so that all services utilizing the address can pull it from there.

crates/interledger-ildcp/src/server.rs
gakonst: It might be better if we push the behavior to the store. It would not be that big of a problem, since if we implement https://github.com/interledger-rs/interledger-rs/issues/195, we'd just configure it to adjust the address properly before calling the specific store implementation.

crates/interledger-store-redis/src/store.rs
gakonst: I think that adding it to the `ilp_address` method is equivalent to doing it at load-time, or even less flexible actually.

[`ilp_address` fetches the address from the Account.](https://github.com/bstrie/interledger-rs/blob/5d620abbe2ff68941284d282ff19c972502ae086/crates/interledger-store-redis/src/account.rs#L496). To give it the ability to append the account's `Username` to the node's address, the `Account` struct would need to have a `node_address` field. That is pretty cumbersome, so I'd rather avoid that.

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -30,6 +81,37 @@ fn insert_accounts() {
...
+#[test]
+fn only_one_parent_allowed() {
+    let mut acc = ACCOUNT_DETAILS_2.clone();
+    acc.routing_relation = Some("Parent".to_owned());
+    acc.username = Username::from_str("another_name").unwrap();
+    acc.configured_ilp_address = Some(Address::from_str("example.another_name").unwrap());
+    block_on(test_store().and_then(|(store, context, accs)| {
+        // check that the store's ilp address is <parent address>.username
+        // assert_eq!(store.ilp
+        store.insert_account(acc.clone()).then(move |res| {
```
emschwartz: Why not just overwrite which node is the parent?

crates/interledger/src/node.rs
```diff
@@ -90,8 +95,13 @@ where
 #[derive(Deserialize, Clone)]
 pub struct InterledgerNode {
     /// ILP address of the node
-    #[serde(deserialize_with = "deserialize_string_to_address")]
+    // Rename this one because the env vars are prefixed with "ILP_"
+    #[serde(alias = "address")]
```
emschwartz: I think @dora-gt intentionally removed this to replace it with a different way to parse the parameter

crates/interledger/src/node.rs
```diff
@@ -90,8 +95,13 @@ where
...
+    #[serde(default = "default_ilp_address")]
+    #[serde(deserialize_with = "deserialize_string")]
     pub ilp_address: Address,
+    #[serde(deserialize_with = "deserialize_string")]
+    pub username: Username,
```
emschwartz: This shouldn't be needed

crates/interledger/src/node.rs
```diff
@@ -165,10 +168,20 @@ impl InterledgerNode {
         let exchange_rate_poll_interval = self.exchange_rate_poll_interval;
         let exchange_rate_spread = self.exchange_rate_spread;
 
-        RedisStoreBuilder::new(self.redis_connection.clone(), redis_secret)
-        .connect()
+        RedisStoreBuilder::new(self.redis_connection.clone(), redis_secret, self.username.clone())
+            .node_ilp_address(self.ilp_address.clone())
```
emschwartz: Isn't the whole idea of this PR that it will get this address from the parent if it isn't configured on the node?

crates/interledger/src/node.rs
dora-gt: It accepts `ILP_ADDRESS`.

https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger/src/main.rs#L457-L461

crates/interledger-store-redis/src/store.rs
```diff
@@ -212,20 +225,50 @@ impl RedisStoreBuilder {
...
-                let connection_clone = Arc::downgrade(&store.connection);
-                let routing_table = store.routes.clone();
-                let poll_routes =
-                    Interval::new(Instant::now(), Duration::from_millis(poll_interval))
+                // Before initializing the store, check if we have a parent and
+                // use its address. If no parent was found, use the builder's
+                // provided address (local.host) or the one we decided to
+                // override with
+                redis::cmd("GET")
+                    .arg("parent_node_account_address")
```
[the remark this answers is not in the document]
gakonst: This would mean that for some reason the parent assigned a new address to us. This would happen if a node operator edited our ILP Address to a new value. Is this expected to happen? We can definitely add it. It would require adding a service to the store, similarly to how it's been added to the API and make an Ildcp request. This starts to feel a bit hacky as we're injecting services everywhere to make requests. Is there a way to make this slightly more stateless?

crates/interledger-store-redis/tests/accounts_test.rs
gakonst: We're testing that only 1 parent can be inserted, check the latest version of the test in the last commit

gakonst:
This started failing for no reason, the commit after which the CI started failing just modified a test. I cannot reproduce on my local machine.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -409,3 +449,59 @@ where
...
+            let ilp_address = match response {
+                Ok(info) => info.ilp_address(),
+                Err(_) => return err(()),
+            };
+            ok(ilp_address)
+        })
+        .and_then(move |ilp_address| store.set_ilp_address(ilp_address))
+        .and_then(move |_| {
+            // Get the parent's routes for us
+            let prepare = RouteControlRequest {
```
emschwartz: Instead of constructing a `RouteControlRequest` here, what do you think about having a handle to the `CcpServer` to trigger it to send the request? If that's a lot more complicated, we can keep this as is

crates/interledger-ccp/src/server.rs
```diff
@@ -182,6 +185,17 @@ where
         })
     }
 
+    fn update_ilp_address(&self) {
+        let ilp_address = self.store.get_ilp_address();
```
emschwartz: nit: don't bother getting the `write` lock if it hasn't changed

crates/interledger-api/src/lib.rs
```diff
@@ -80,9 +81,9 @@ pub struct AccountSettings {
 }
 
 /// The Account type for the RedisStore.
-#[derive(Debug, Clone, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct AccountDetails {
-    pub ilp_address: Address,
+    pub configured_ilp_address: Option<Address>,
```
emschwartz: I think I probably suggested this name change in the first place, but what do you think about just keeping it as `ilp_address` and making it optional? Would that cause any issues?

This is based on:
- Everything in this struct is "configuration" so it's a bit strange to explicitly say that one of the fields here is configuration
- I don't think it would be that confusing for the field on the call to create the account to be called `ilp_address` and then if you don't fill it in, to have the `ilp_address` field filled in on the response
- Loosely applying the principle of don't change/break the API if we don't have to (though I recognize that we've been making a number of breaking changes and I don't really consider it stable yet)

crates/interledger-settlement-engines/tests/eth_ledger_settlement.rs
```diff
@@ -156,49 +156,46 @@ fn eth_ledger_settlement() {
...
+                    http_incoming_token: Some("bob".to_string()),
+                    http_outgoing_token: Some("bob:alice".to_string()),
                     max_packet_amount: 10,
-                    min_balance: None,
-                    settle_threshold: None,
-                    settle_to: None,
-                    routing_relation: Some("NonRoutingAccount".to_owned()),
+                    min_balance: Some(-100),
+                    settle_threshold: Some(70),
+                    settle_to: Some(-10),
```
emschwartz: Why do these fields need to be changed?

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
```diff
@@ -114,56 +112,38 @@ fn eth_xrp_interoperable() {
...
+        "configured_ilp_address": "example.alice",
+        "username": "alice",
+        "asset_code": "ETH",
+        "asset_scale": 9,
+        "http_endpoint": format!("http://localhost:{}/ilp", node1_http),
+        "http_incoming_token" : "in_alice",
+    });
+
+    let bob_on_alice = json!({
+        "configured_ilp_address": "example.bob",
```
emschwartz: This should be an issue for a separate PR but I wonder if we need to configure the ILP address for Peers. We might be able to get their address from the `CcpServer` instead

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
```diff
@@ -263,157 +248,80 @@ fn eth_xrp_interoperable() {
...
+            .and_then(move |_| node3.serve())
+            .and_then(move |_| charlie_fut)
+            .and_then(move |_| Ok(())),
     );
 
     runtime
         .block_on(
             // Wait for the nodes to spin up
-            delay(500)
+            delay(5000)
```
emschwartz: Why does this need to be so long?

crates/interledger-store-redis/src/store.rs
```diff
@@ -416,17 +460,22 @@ impl RedisStore {
...
+        let ilp_address = self.get_ilp_address().clone();
 
         Box::new(
             // Check to make sure an account with this ID already exists
             redis::cmd("EXISTS")
                 .arg(accounts_key(id))
-                // TODO this needs to be atomic with the insertions later, waiting on #186
+                // TODO this needs to be atomic with the insertions later,
+                // waiting on #186
+                // TODO: Do not allow this update to happen if
```
emschwartz: This TODO should probably be implemented before this is merged in

crates/interledger-store-redis/src/store.rs
```diff
@@ -1141,6 +1189,78 @@ impl NodeStore for RedisStore {
...
+            cmd("SET")
+                .arg(PARENT_ILP_KEY)
+                .arg(ilp_address.as_bytes())
+                .query_async(self.connection.as_ref().clone())
+                .map_err(|err| error!("Error setting static ilp address {:?}", err))
+                .and_then(move |(_, _): (SharedConnection, Value)| {
+                    *(self_clone.ilp_address.write()) = ilp_address;
+                    Ok(())
+                })
+                .join(self.get_all_accounts().and_then(move |accounts| {
```
emschwartz: It might be slightly better to do this as a lua script so that we don't need to first have it query and serialize all of the accounts, but instead do it within the DB

crates/interledger-store-redis/src/store.rs
```diff
@@ -1141,6 +1189,78 @@ impl NodeStore for RedisStore {
...
+                    // TODO: This can be an expensive operation if this function
+                    // gets called often. This currently only gets called when
+                    // inserting a new parent account in the API. It'd be nice
+                    // if we could generate a child's ILP address on the fly,
+                    // instead of having to store the username appended to the
+                    // node's ilp address. Currently this is not possible, as
+                    // account.ilp_address() cannot access any state that exists
```
emschwartz: Could / should we give the `Account` a handle to the store? I think the only thing this would change is that instead of being able to use the automatic `FromRedisArgs` implementation, we would need to make calls that return values from redis return `redis::Value` and then explicitly call the parse command (with the store handle included)

crates/interledger-store-redis/src/store.rs
```diff
@@ -1141,6 +1189,78 @@ impl NodeStore for RedisStore {
...
+        )
+    }
+
+    fn clear_ilp_address(&self) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        let self_clone = self.clone();
+        Box::new(
+            cmd("DEL")
+                .arg(PARENT_ILP_KEY)
+                .query_async(self.connection.as_ref().clone())
+                .map_err(|err| error!("Error setting static route: {:?}", err))
```
emschwartz: Wrong error message

crates/interledger/src/node.rs
```diff
@@ -168,6 +168,7 @@ impl InterledgerNode {
         let exchange_rate_spread = self.exchange_rate_spread;
 
         RedisStoreBuilder::new(self.redis_connection.clone(), redis_secret)
+        .node_ilp_address(ilp_address.clone())
```
emschwartz: The `ilp_address` should be optional in the Node configuration and we should make it use the `local.host` or whatever automatic one we decide if the configured value is `None`

crates/interledger/tests/exchange_rates.rs
```diff
@@ -44,7 +44,7 @@ fn coincap() {
 
     runtime
         .block_on(
-            delay(1000)
+            delay(2000)
```
emschwartz: Why does this need to be longer?

crates/interledger-api/src/routes/accounts.rs
gakonst: We could but I'd prefer not. CcpRouteManager is generic over `Incoming`, `Outgoing` services and a `Store`, and that'd require adding extra generic parameters on the API. 

The 'benefit" would be being more explicit on the types of outgoing requests that we can make. I don't think that's really beneficial, since in the future we might want to perform other actions and send other types of requests, than just CCP requests via the outgoing handler.

crates/interledger-settlement-engines/tests/eth_ledger_settlement.rs
gakonst: I changed the order of insertion of the two accounts, but you're right that it's not needed.

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
gakonst: Was an initial ballpark to get the test going. Changed to 1000ms. 500 is too short as the previous account adding operations do not complete.

crates/interledger-store-redis/src/store.rs
gakonst: I'd rather keep it as is, this is an infrequent operation and I'd rather not do data manipulation in lua

crates/interledger-store-redis/src/store.rs
gakonst: That would require each account to have a handle to the store. Wouldn't this be expensive? If we did, then sure, we could do `Account.store.get_ilp_address()` and generate the ILP Address in the redis trait implementation. Not sure how I feel about adding store related stuff on the account, seems like bad a abstraction

crates/interledger-store-redis/src/store.rs
gakonst: This method needs a lot of refinement which I think should go in a separate PR.

- Ilp addresses must be unique
- should we allow the username to change?
- should we allow promoting an existing account to a parent?
- should we allow demotion of an existing account to child? if so, should we let them remove their ILP address to that it gets recalculated via their username+node ilp address?

IMO we should just disallow any modifications to the routing relationship and require that they delete/re-insert the account.

crates/interledger/src/node.rs
gakonst: Serde problems, @emschwartz said he'd look at it

crates/interledger-store-redis/src/store.rs
emschwartz: I don't think having a handle to the store is too expensive but I agree it breaks the abstraction a bit. The question is whether the Account is just data loaded from the store, or more like part of the store / a handle to some particular data in the store

crates/interledger-api/src/routes/accounts.rs
emschwartz: Sounds good for now. Mind adding a TODO to make a note about this?
```suggestion
            // TODO we may want to make this trigger the CcpRouteManager to request
            // routes instead of sending a RouteControlRequest here
            let prepare = RouteControlRequest {
```

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -409,3 +398,137 @@ where
...
+                        .map_err(move |err| {
+                            error!("Got error when trying to update routes {:?}", err)
+                        }),
+                ),
+            ])
+        })
+        .and_then(move |_| Ok(()))
+}
+
+fn connect_to_external_services<S, A, T, B>(
```
emschwartz: This is a very generic sounding name. It would be good to be more specific to what it's doing and/or put a comment explaining it

crates/interledger-btp/src/client.rs
```diff
@@ -33,77 +32,90 @@ where
...
+                            protocol_name: String::from("auth_token"),
+                            content_type: ContentType::TextPlainUtf8,
+                            data: token,
+                        },
+                    ],
+                })
+                .to_bytes(),
+            );
+
+            connection
```
emschwartz:
```suggestion
            // TODO check that the response is a success before proceeding
            // (right now we just assume they'll close the connection if the auth didn't work)
            connection
```

crates/interledger/src/node.rs
emschwartz: Implemented in https://github.com/interledger-rs/interledger-rs/pull/329

PR #287, opened 2019-09-07
fix(examples): spruce up all three examples

bstrie:
While fixing a bug in one of the examples I got caught up in fixing other bugs that I found, and did so for all three examples. This commit fixes the following:

1. Previously we were building with `cargo build --bins` and running with `cargo run --package foo`, but because --bin and --package represent different things, we were actually building nearly the whole repo twice in every example. I've switched to using --bin everywhere, which should have a dramatic effect on build times for new users.

2. Scripts weren't tearing down their servers at the end, presumably as a temporary solution to solve the problem of the user seeing error messages from processes as they were force-killed. I've addressed this with `exec 2> /dev/null` before we kill the processes and re-enabled the teardown at the end.

3. We intentionally tear down twice in each example: at the beginning to make sure we're starting fresh (in case a previous run of the script was killed prematurely, for instance), and at the end to be nice to our users. We were somewhat inconsistently silencing the output of various teardown steps; now we silence all teardown for the finishing stage, but don't silence teardown for the beginning stage (since tearing down in the beginning stage is unexpected, so if we see anything that may indicative of a bug).

4. Inserted a sleep in one spot so that Redis has a chance to start before trying to flush it, which avoids printing an error to the terminal.

5. Updates the two advanced examples so that they function following #268

### DIFF
--- a/examples/eth-settlement/README.md
+++ b/examples/eth-settlement/README.md
@@ -51,7 +51,7 @@ Make sure your Redis is empty. You could run `redis-cli flushall` to clear all t
 <!--!
 printf "Stopping Interledger nodes\n"
 
-if lsof -Pi :6379 -sTCP:LISTEN -t >/dev/null ; then
+if lsof -Pi :6379 -sTCP:LISTEN -t ; then
     redis-cli -p 6379 shutdown
 fi
 
@@ -85,12 +85,13 @@ First of all, let's build interledger.rs. (This may take a couple of minutes)
 
 <!--! printf "Building interledger.rs... (This may take a couple of minutes)\n" -->
 ```bash
-cargo build --bins
+cargo build --bin interledger --bin interledger-settlement-engines
 ```
 
 ### 2. Launch Redis
 
 <!--!
+printf "Starting Redis\n"
 redis-server --version > /dev/null || printf "\e[31mUh oh! You need to install redis-server before running this example\e[m\n"
 -->
 
@@ -100,6 +101,15 @@ mkdir -p logs
 
 # Start Redis
 redis-server &> logs/redis.log &
+```
+
+<!--!
+sleep 1
+-->
+
+To remove all the data in Redis, you might additionally perform:
+
+```bash
 redis-cli flushall
 ```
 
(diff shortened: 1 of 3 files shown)


dora-gt:
Thank you Ben,

Why I tried NOT to kill all the services after the procedure, was that readers might check whether the settlement has been done. If the services are not working, they can't check that. So we need to keep them alive.

Because I'm trying to change many things in #270, can I try to implement some of these improvements in that PR instead of merging this?

bstrie:
@dora-gt Yes, feel free to incorporate these changes into your own PR and I can close this one. The most important change is the one where we change `--package` to `-bin`, to avoid recompiling everything twice. Regarding leaving the services running, could we prompt the user with a yes/no to ask them if they want us to stop the services for them? I believe this would also be useful for us when testing.

dora-gt:
Thank you, I've almost done reflecting your suggestions on this PR #270.

> Regarding leaving the services running, could we prompt the user with a yes/no to ask them if they want us to stop the services for them?

It would be better. I'll try.

bstrie:
>> Regarding leaving the services running, could we prompt the user with a yes/no to ask them if they want us to stop the services for them?
> 
> It would be better. I'll try.

Thinking about it more, ideally the script itself could check whether the settlement has been done, Am I correct in thinking that the only reason the script does not do this today is because we don't want to require the user to have Geth installed? If so, then as an alternative to asking the user, we could test whether Geth is installed, and if Geth is installed then we attempt to check the settlement (and kill the services afterward), and if Geth is not installed then we print information about how to install Geth and kill the services immediately. Not sure which approach is best.

dora-gt:
Actually if we just only want to wait for the settlement, we could without `geth` because we could `curl` many times till it gets settled value. Currently we are printing `curl` commands to confirm settlements like:

image

One thing I'm concerned about is that we have to wait a few ... or several seconds. I'm not sure it is better.

hmmm IMO, printing "Do you want to kill the services? [y/N]" is sufficient.

bstrie:
I'll close this and we can continue conversation in #293.

Issue #277, opened 2019-09-05
Should RoutingRelation be used to determine whether to send/receive routes?


emschwartz:
Currently, each account is configured with the properties `routing_relation`, `send_routes`, and `receive_routes`. Should we remove `send_routes` and `receive_routes` and use the `routing_relation` to determine whether we want to send or receive route broadcasts?

### Routing Relation

The idea of the routing relationship was borrowed from the Internet and BGP. `Peer`s are two nodes (like ISPs) that are on equal footing with one another and generally do not charge one another for forwarding traffic. A `Parent` is an upstream provider that probably charges for any traffic you send their way and a `Child` indicates when you are the upstream provider for another node and are probably charging it for traffic. The main purpose of making this distinction is to avoid [valley-free routing](https://blog.ipspace.net/2018/09/valley-free-routing.html).

Right now in Interledger.rs, the `routing_relation` is only used to decide between competing route broadcasts that have the same path length (see the code in the [CCP server](https://github.com/interledger-rs/interledger-rs/blob/79593217b1e12692857098a011c59b37727a2c7c/crates/interledger-ccp/src/server.rs#L798-L799)).

### Scenarios to Consider

| `routing_relation` | `send_routes`? | `receive_routes`? |
|---|---|---|
| `Parent` | probably not worth it because they'll likely ignore the broadcast anyway | yes, if they broadcast routes |
| `Child` | maybe? | no |
| `Peer` | probably | probably |

#### `send_routes` to a `Child`?

On the Internet, your ISP doesn't send you route broadcasts because the assumption is that you will always forward it all of your traffic (so you don't need a real routing table).

If we want to support having multiple parents, we would _probably_ want to have the parents send route broadcasts to the child nodes (so the child could decide which parent to send traffic to). That said, the child node wouldn't need to use the routing table to make the determination of which parent to send particular packets to if they used multi-path STREAM (https://github.com/interledger-rs/interledger-rs/issues/44). That would work fine for _sending_ from that child node but would still leave the question open of how it would determine which parent to send particular packets through if it needed to route packets from other processes.

#### Would you ever not `send_routes` and `receive_routes` with a `Peer`?

I can't think of a reason why you wouldn't, but it's possible there is some use case.

### Questions
1. Should we get rid of the separate `send_routes` and `receive_routes` flags?
[...]

PR #288, opened 2019-09-09
Merge send/receive routes functionality in Peer Relationship

gakonst:
Closes https://github.com/interledger-rs/interledger-rs/issues/277

send/receive routes -> relation
false/false -> nonrouting
true/false -> parent
false/true -> child
true/true -> peer


crates/interledger-ccp/src/lib.rs
```diff
@@ -31,6 +31,7 @@ use serde::{Deserialize, Serialize};
 #[repr(u8)]
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub enum RoutingRelation {
```
dora-gt: Though this is not related to your change, sometimes I get confused when I read the code of relations.
i.e. Is `Parent` in point of us? or them? Am I `Parent` or is it `Parent`?

So I think putting some comments on this struct helps a lot for interledger-rs novices.

crates/interledger-ccp/src/lib.rs
```diff
@@ -41,6 +42,7 @@ impl FromStr for RoutingRelation {
 
     fn from_str(string: &str) -> Result<Self, ()> {
         match string.to_lowercase().as_str() {
+            "nonroutingaccount" => Ok(RoutingRelation::NonRoutingAccount),
```
dora-gt: Isn't this better? `non_routing_account` or `non-routing-account`?
(It looks a bit hard to read)

dora-gt:
JFYI: I'll fix examples according to this change.

PR #296, opened 2019-09-11
Import Ethereum Transaction Signing logic 

gakonst:
Instead of having to rely on an [external dependency](https://github.com/synlestidae/ethereum-tx-sign/) we just import the code.

We also add conditional compilation for the ethereum engine.


crates/interledger-settlement-engines/src/engines/ethereum_ledger/raw_transaction.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+fn ecdsa_sign(hash: &[u8], private_key: &[u8], chain_id: u8) -> EcdsaSig {
+    let s = Secp256k1::signing_only();
+    let msg = Message::from_slice(hash).unwrap();
+    let key = SecretKey::from_slice(private_key).unwrap();
+    let (v, sig_bytes) = s.sign_recoverable(&msg, &key).serialize_compact();
+
+    EcdsaSig {
+        v: vec![v.to_i32() as u8 + chain_id * 2 + 35],
```
dora-gt: I think we should basically avoid using magic numbers (because it is hard to understand what's happening), so adding comment referring the EIP would be better.
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

Honestly, I don't know much about Ethereum but is it OK just setting `35`?
Isn't it like the following?

```
v: vec![v.to_i32() as u8 + chain_id * 2 + 8 + v],
```

gakonst:
```suggestion
       // EIP155: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md
        v: vec![v.to_i32() as u8 + chain_id * 2 + 35],
```

crates/interledger-settlement-engines/Cargo.toml
```diff
@@ -53,3 +57,6 @@ rand = "0.7.0"
 interledger = { path = "../interledger", version = "^0.4.1-alpha.1"}
 interledger-packet = { path = "../interledger-packet", version = "^0.2.2-alpha.1"}
 interledger-store-redis = { path = "../interledger-store-redis", version = "^0.2.2-alpha.1"}
+
+[features]
+ethereum = ["tiny-keccak", "secp256k1", "rlp", "ethereum-types"]
```
emschwartz: Is this included by default? Wondering how this plays with the integration tests that depend on this being available

Issue #300, opened 2019-09-12
Configure tokio to kill the server when a worker panics


bstrie:
Right now a panic in a worker is logged and then the server continues on, but we probably want the server to go down instead so that we actually notice such errors.

PR #302, opened 2019-09-12
Propagate panics from tokio

emschwartz:
Resolves https://github.com/interledger-rs/interledger-rs/issues/300


crates/interledger/src/node.rs
```diff
@@ -17,6 +17,7 @@ use interledger_service_util::{
     MaxPacketAmountService, RateLimitService, ValidatorService,
 };
 // Re-export this
```
gakonst: This comment should be on the following line since it refers to re-exporting the ExchangeRateProvider

Issue #198, opened 2019-08-07
Make `.md`s testable


dora-gt:
I noticed that we made `.md`s runnable so we could even test it.
It has a pro that if the documents go outdated, we will be able to notice that.

That said, it will require more efforts, so... not now but sometime in the future.

emschwartz:
I think that's a great idea. It would probably take a bit more to include assertions but it might be worth just having CI run them as part of the normal flow

dora-gt:
I strongly feel the importance of this feature because interledger-rs is rapidly growing and there are many changes which affect documents.

I'm thinking of what is the right approach to implement this.

1. Insert test codes into README.md with a specific mark
```
some explanations

    ```bash #test
    # do some test
    ```
```
`#test` means this code is for test, should not be run by `run-md.sh`. This option seems good because **sometimes** it can reuse exit codes of commands of the procedure. The tests output some information to a file, CI process reads it.

2. Run tests after `run-md.sh`
    - CI runs `run-md.sh` for some markdown files.
    - After that, CI runs tests whether the final state is correct or not.

This option only tests whether the **result** is correct or not. Although it seems correct approach as an integrated test, may be inefficient because it cannot reuse exit codes and sometimes it is hard to define what is the **correct** state as the result.

or... should we support both?

What do you think? @interledger-rs/contributors

bstrie:
Rather than adding new tests within each file that are hidden from view, I would like to invoke `run-md.sh` as part of the test, and find a way to make the usual run of each README produce output that could be tested. For example, when the simple test produces this output at the end:
```
Alice's balance: {"balance":"-500"}
Node B's balance on Node A: {"balance":"500"}
Node A's balance on Node B: {"balance":"-500"}
Bob's balance: {"balance":"500"}
```
...we could also write each JSON response to a different file in the `logs` directory, and have a test that:

1. Deletes any existing output files
2. Executes `run-md.sh` with `simple/README.md`
3. Blocks while waiting for all four files to be created
4. Reads all four files and tests that the results are as expected

I haven't taken a look at the other examples to see if they produce testable output, but if not then perhaps they could be changed to do so.

dora-gt:
Thank you Ben.
(I've edited the original post because the expression was not hiding the test from readers)

I understand the approach. One thing I'm curious about is why you would like the approach.
It doesn't make the `.md`s messy? or.. it looks simpler?

bstrie:
[...]

PR #303, opened 2019-09-13
Make `.md`s testable

dora-gt:
This is a part of #198
I'm breaking the issue down into separate PRs this time.

What this PR realizes is to make `.md`s testable but it is NOT automated yet.
Integrating into CircleCI would be later.

- Just run `run-md-test.sh` to test markdowns.
- `run-md-test.sh` injects `hook_before_kill` functions into the script of markdowns.
    - The tests are actually written in `scripts/run-md-tests/examples`.
        - Each file defines a `hook_before_kill` function.
    - Why I adopted this approach is:
        - We wouldn't know when to test. i.e. when the payment is done. We have to wait for something without this.
        - This approach almost doesn't affect the original code.
    - Though this is slightly different from Ben's suggestion: https://github.com/interledger-rs/interledger-rs/issues/198#issuecomment-527599418 is this acceptable? @bstrie 
- `run-md-test.sh` exits with `0` when it succeeded, and with `1` when it failed. So integrating this script into CircleCI wouldn't be so hard.

In a case of success:
image

In a case of failure:
image

### DIFF
--- a/scripts/run-md-test.sh
+++ b/scripts/run-md-test.sh
@@ -1,0 +1,100 @@
+#!/bin/bash
+
+function init_test() {
+    clear_logs
+    clear_docker
+    free_ports
+    clear_redis_file
+}
+
+function clear_logs() {
+    rm -rf logs
+}
+
+function clear_docker() {
+    docker network rm interledger 2>/dev/null
+    docker stop $(docker ps -aq) 2>/dev/null
+    docker rm $(docker ps -aq) 2>/dev/null
+}
+
+function free_ports() {
+    # ports of redis
+    for port in $(seq 6379 6385); do
+        if lsof -Pi :${port} -sTCP:LISTEN -t >/dev/null ; then
+            redis-cli -p ${port} shutdown
+        fi
+    done
+
+    # ports of ganache + node + se
+    for port in 8545 7770 8770 9770 3000 3001 3002 3003; do
+        if lsof -tPi :${port} >/dev/null ; then
+            kill `lsof -tPi :${port}`
+        fi
+    done
+}
+
+function clear_redis_file() {
+    # redis file
+    if [ -f dump.rdb ] ; then
+        rm -f dump.rdb
+    fi
+}
+
+# $1 = example name
+# $2 = docker mode in [01]
+#
+# test_example eth-settlement
+#
+# This heavily depends on the naming convention.
+# - "run-md-tests/examples/$1.sh" MUST exist
+# - "$BASE_DIR/../examples/$1" MUST exist
+function test_example() {
+    local target_name=$1
+    local docker_mode=$2
+
+    # doing in a sub-shell because we want to load `hook_before_kill` many times
+    (
+        cd $BASE_DIR/../examples/$target_name
+        source ../../scripts/run-md-tests/examples/${target_name}.sh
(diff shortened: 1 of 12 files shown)


scripts/run-md-tests/README.md
```diff
@@ -0,0 +1,9 @@
+## Tests for Runnable Markdowns
+
+Should implement `hook_before_kill` function like:
```
gakonst: This is very clean, LGTM.

scripts/run-md-test.sh
```diff
@@ -0,0 +1,100 @@
...
+}
+
+# $1 = example name
+# $2 = docker mode in [01]
+#
+# test_example eth-settlement
+#
+# This heavily depends on the naming convention.
+# - "run-md-tests/examples/$1.sh" MUST exist
+# - "$BASE_DIR/../examples/$1" MUST exist
```
emschwartz: What is this one?

scripts/run-md-test.sh
```diff
@@ -0,0 +1,100 @@
...
+
+tests=()
+tests+=("test_example eth-settlement 0")
+tests+=("test_example eth-settlement 1")
+tests+=("test_example eth-xrp-three-nodes 0")
+tests+=("test_example eth-xrp-three-nodes 1")
+tests+=("test_example simple 0")
+tests+=("test_example simple 1")
+tests+=("test_example xrp-settlement 0")
+tests+=("test_example xrp-settlement 1")
```
emschwartz: Would it be possible to just make sure this runs on all of the directories in the `examples` directory instead of explicitly listing which it should run on?

scripts/run-md-tests/README.md
emschwartz:
```suggestion
In order to implement tests for a runnable markdown file, the function `hook_before_kill` should be implemented with the assertions to run after the example finishes. Bash variables can be set while the example is running, which can then be checked in this final hook.
```

scripts/run-md-tests/examples/eth-settlement.sh
```diff
@@ -0,0 +1,6 @@
+function hook_before_kill() {
```
emschwartz: Can this be included at the bottom of the markdown file instead of in a separate file?

scripts/run-md-test.sh
dora-gt: It is example directory itself. For example,`examples/eth-settlement`.

scripts/run-md-tests/README.md
dora-gt: Removed this `.md` because tests are merged into the original markdowns.

scripts/run-md-test.sh
dora-gt: Fixed.

scripts/run-md-tests/examples/eth-settlement.sh
dora-gt: Yes, I thought separating tests would be better but the tests were a few lines, so merged into the original files.

scripts/run-md-test.sh
dora-gt: Actually we might not need this one🤔 I'll fix the structure.

dora-gt:
I'm merging this.

PR #307, opened 2019-09-17
feat: API for receiving notifications upon payment receipt

bstrie:
The integration test isn't yet added, but other than that everything works from publishing to receiving the websocket notifications (however `wscat` doesn't want to follow redirects, so despite the HTTP endpoint serving up a proper redirect you'll need to connect to the websocket port directly to test it from CLI).


crates/interledger-api/src/lib.rs
```diff
@@ -151,6 +176,82 @@ where
...
+                            error!("Failed to identify valid request path");
+                            Either::B(ok(()))
+                        }
+                        Some(captures) => match captures.name("user") {
+                            None => {
+                                error!("Failed to locate username in path");
+                                Either::B(ok(()))
+                            }
+                            Some(username) => Either::A(
+                                // TODO Validate authorization first
```
emschwartz: This should be done before this gets merged

crates/interledger-api/src/lib.rs
```diff
@@ -151,6 +176,82 @@ where
...
+                                            ws_stream
+                                                .send_all(rx.map(Message::Text).map_err(|_| {
+                                                    warn!("Websocket connection aborted");
+                                                    WebSocketError::from(IoError::from(
+                                                        ErrorKind::ConnectionAborted,
+                                                    ))
+                                                }))
+                                                .then(move |_| {
+                                                    info!("Finished forwarding subscriber to account: {}", username);
+                                                    // TODO close connection
```
emschwartz: I think this will close the connection, no?

crates/interledger-api/src/lib.rs
```diff
@@ -151,6 +176,82 @@ where
...
+                                        Ok(())
+                                    })
+                                }),
+                            )
+                        }
+                    }
+                })
+            })
+            .map_err(|_| ())
+            .and_then(|_| Ok(()))
```
emschwartz: This doesn't do anything

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -57,10 +59,11 @@ impl_web! {
     A: Account + HttpAccount + Serialize + 'static,
 
     {
-        pub fn new(admin_api_token: String, store: T) -> Self {
+        pub fn new(admin_api_token: String, store: T, notifications_address: SocketAddr) -> Self {
```
emschwartz: nit: store should come last

crates/interledger-store-memory/Cargo.toml
```diff
@@ -17,5 +17,6 @@ interledger-router = { path = "../interledger-router", version = "^0.2.2-alpha.1
 interledger-service = { path = "../interledger-service", version = "^0.2.2-alpha.1"}
 interledger-packet = { path = "../interledger-packet", version = "^0.2.2-alpha.1"}
 interledger-service-util = { path = "../interledger-service-util", version = "^0.2.2-alpha.1"}
+interledger-stream = { path = "../interledger-stream", version = "^0.2.2-alpha.1"}
```
emschwartz: I don't think this dependency should be needed

crates/interledger-store-memory/src/store.rs
```diff
@@ -138,6 +139,8 @@ impl AccountStore for InMemoryStore {
     }
 }
 
+impl PubStore for InMemoryStore {}
```
emschwartz: I don't think this store should implement this trait because it doesn't actually implement the notifications functionality

crates/interledger-store-redis/src/store.rs
```diff
@@ -199,21 +203,33 @@ impl RedisStoreBuilder {
...
-                debug!("Connected to redis: {:?}", client);
+                debug!("Connected shared client to redis: {:?}", client);
                 client
                     .get_shared_async_connection()
-                    .map_err(|err| error!("Error connecting to Redis: {:?}", err))
+                    .map_err(|err| error!("Error connecting shared client to Redis: {:?}", err))
             })
-            .and_then(move |connection| {
+            .join(
+                result(Client::open(redis_info.clone()))
```
emschwartz: Seems like it may cause issues to make this synchronous IO call in a Future chain. Generally Tokio doesn't like that

Also, please put a comment explaining what this other connection is for

crates/interledger-store-redis/src/store.rs
```diff
@@ -241,6 +257,36 @@ impl RedisStoreBuilder {
                         });
                 spawn(poll_routes);
 
+                let subs_clone = store.subscriptions.clone();
```
emschwartz: Please write out the full words rather than abbreviating to "subs"

crates/interledger-store-redis/src/store.rs
```diff
@@ -241,6 +257,36 @@ impl RedisStoreBuilder {
...
+                std::thread::spawn(move || {
+                    let sub_status =
```
emschwartz: Please put comments explaining what is going on here

crates/interledger-store-redis/src/store.rs
```diff
@@ -241,6 +257,36 @@ impl RedisStoreBuilder {
...
+                                Ok(s) => s,
+                                Err(e) => {
+                                    error!("Failed to get payload from subscription: {}", e);
+                                    return ControlFlow::Continue;
+                                }
+                            };
+                            debug!("Subscribed message received for {:?}", channel[1]);
+                            match subs_clone.read().get(channel[1]) {
+                                Some(sender) => {
+                                    debug!("Mailboxes looking for {:?}: {:?}", channel[1], sender);
```
emschwartz: What's a mailbox?

crates/interledger-store-redis/src/store.rs
```diff
@@ -241,6 +257,36 @@ impl RedisStoreBuilder {
...
+                                        error!("Failed to send message: {}", e);
+                                    }
+                                }
+                                None => debug!("No mailboxes looking for {:?}", channel[1]),
+                            }
+                            ControlFlow::Continue
+                        });
+                    match sub_status {
+                        Err(e) => warn!("Could not issue psubscribe to Redis: {}", e),
+                        Ok(_) => info!("Successfully issued psubscribe to Redis"),
```
emschwartz: This should probably be debug or trace

crates/interledger-store-redis/src/store.rs
```diff
@@ -241,6 +257,36 @@ impl RedisStoreBuilder {
...
+                                    if let Err(e) = sender.unbounded_send(message) {
+                                        error!("Failed to send message: {}", e);
+                                    }
+                                }
+                                None => debug!("No mailboxes looking for {:?}", channel[1]),
+                            }
+                            ControlFlow::Continue
+                        });
+                    match sub_status {
+                        Err(e) => warn!("Could not issue psubscribe to Redis: {}", e),
```
emschwartz: If this means the service won't work as intended, this should at least log an `error` but more likely crash the node or something like that to indicate it isn't set up

crates/interledger-store-redis/src/store.rs
```diff
@@ -618,6 +665,23 @@ impl AccountStore for RedisStore {
     }
 }
 
+impl PubStore for RedisStore {
+    fn publish_payment_notification(&self, pmt: PaymentNotification) {
```
emschwartz: Same as above, please use the full words (helps with readability)

crates/interledger-stream/src/server.rs
```diff
@@ -85,37 +89,58 @@ impl ConnectionGenerator {
...
 
-impl<O, A> StreamReceiverService<O, A>
+impl<S, O, A> StreamReceiverService<S, O, A>
 where
+    S: PubStore,
     O: OutgoingService<A>,
     A: Account,
 {
-    pub fn new(server_secret: Bytes, next: O) -> Self {
+    pub fn new(server_secret: Bytes, store: S, next: O) -> Self {
```
emschwartz: Still think this should be an `Option<S>` so that we don't need the no-op implementation of `publish_payment_notification`

crates/interledger-stream/src/server.rs
```diff
@@ -85,37 +89,58 @@ impl ConnectionGenerator {
...
+pub struct PaymentNotification {
+    pub to_username: Username,
+    pub from_username: Username,
+    pub destination: Address,
+    pub amount: u64,
+    pub timestamp: String,
+}
+
+/// A trait representing the Publish side of a pub/sub store
+pub trait PubStore {
```
emschwartz: It would be better to make the name more specific to what it is. I would suggest something like:
```suggestion
pub trait StreamNotificationsStore {
```

crates/interledger-stream/src/server.rs
```diff
@@ -128,14 +153,32 @@ where
...
+
+        // The case where the request is bound for this server
+        if dest.starts_with(to_address.as_ref()) {
             if let Ok(shared_secret) = self.connection_generator.rederive_secret(&destination) {
-                {
-                    return Box::new(result(receive_money(&shared_secret, &to, request.prepare)));
-                }
+                return Box::new(
+                    result(receive_money(&shared_secret, &to_address, request.prepare)).and_then(
+                        move |val| {
```
```suggestion
                        move |fulfill| {
```

crates/interledger-stream/src/server.rs
```diff
@@ -85,37 +89,58 @@ impl ConnectionGenerator {
...
+    pub to_username: Username,
+    pub from_username: Username,
+    pub destination: Address,
+    pub amount: u64,
+    pub timestamp: String,
+}
+
+/// A trait representing the Publish side of a pub/sub store
+pub trait PubStore {
+    fn publish_payment_notification(&self, _pmt: PaymentNotification) {
```
emschwartz: If this is synchronous, it should probably return a `Result` so the caller can tell if it didn't work. We might not handle that error right now aside from logging it, but it's better to know if it doesn't work

crates/interledger/src/node.rs
```diff
@@ -49,6 +49,10 @@ fn default_exchange_rate_poll_interval() -> u64 {
     60000
 }
 
+fn default_notifications_bind_address() -> SocketAddr {
+    SocketAddr::from(([127, 0, 0, 1], 7766))
```
emschwartz:
```suggestion
    SocketAddr::from(([127, 0, 0, 1], 7772))
```

examples/simple/README.md
```diff
@@ -187,6 +187,7 @@ ILP_REDIS_URL=redis://127.0.0.1:6379/ \
 ILP_HTTP_BIND_ADDRESS=127.0.0.1:7770 \
 ILP_BTP_BIND_ADDRESS=127.0.0.1:7768 \
 ILP_SETTLEMENT_API_BIND_ADDRESS=127.0.0.1:7771 \
+ILP_NOTIFICATIONS_BIND_ADDRESS=127.0.0.1:7765 \
```
```suggestion
ILP_NOTIFICATIONS_BIND_ADDRESS=127.0.0.1:7772 \
```

examples/simple/README.md
```diff
@@ -196,6 +197,7 @@ ILP_REDIS_URL=redis://127.0.0.1:6380/ \
 ILP_HTTP_BIND_ADDRESS=127.0.0.1:8770 \
 ILP_BTP_BIND_ADDRESS=127.0.0.1:8768 \
 ILP_SETTLEMENT_API_BIND_ADDRESS=127.0.0.1:8771 \
+ILP_NOTIFICATIONS_BIND_ADDRESS=127.0.0.1:8765 \
```
```suggestion
ILP_NOTIFICATIONS_BIND_ADDRESS=127.0.0.1:8772 \
```

crates/interledger-api/src/lib.rs
```diff
@@ -151,6 +176,82 @@ where
...
+                                result(
+                                    Username::from_str(username.as_str())
+                                    .map_err(|e| error!("Failed to parse username: {}", e))
+                                )
+                                .and_then(move |username| {
+                                    let username_clone = username.clone();
+                                    store_clone.get_account_id_from_username(&username)
+                                    .map_err(move |_| error!("Failed to find account with username: {}", username_clone))
+                                    .and_then(move |_account_id| {
+                                        // TODO: use account_id rather than username
```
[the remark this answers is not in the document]
bstrie: I had a good reason why I didn't do this in this PR and typed up a few paragraphs and ended up rubber-ducking myself into what I think will be a reasonable solution, so sure. :)

crates/interledger-api/src/lib.rs
bstrie: I believe I left this here more as a reminder that we should probably be listening for if the client closes the websocket connection so that we can remove that sender from the subscribers map. However I don't know if trying to send to a listener who has since disconnected would actually cause any fatal errors as currently implemented, I'll test that.

crates/interledger-api/src/lib.rs
bstrie: I swear something was happening there at some point. :P

crates/interledger-api/src/lib.rs
```diff
@@ -59,6 +79,8 @@ pub trait NodeStore: Clone + Send + Sync + 'static {
         prefix: String,
         account_id: <Self::Account as AccountTrait>::AccountId,
     ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
+
+    fn add_payment_notification_subscription(&self, id: String, sender: UnboundedSender<String>);
```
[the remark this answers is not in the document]
bstrie: Can you elaborate on what the message type ought to be? Something generic perhaps?

crates/interledger-store-memory/Cargo.toml
bstrie: The repo doesn't build unless `PubStore` is implemented for `InMemoryStore`, which is why we need this dependency.

crates/interledger-store-memory/src/store.rs
bstrie: As I note above, we need this or else `cargo check` fails hard in the repo:
```
    Checking interledger v0.4.1-alpha.1 (/Users/bstriegel/Code/interledger-rs/crates/interledger)
error[E0277]: the trait bound `interledger_store_memory::store::InMemoryStore: interledger_stream::server::PubStore` is not satisfied
   --> crates/interledger/src/cli.rs:225:13
    |
225 |             StreamReceiverService::new(server_secret.clone(), store.clone(), outgoing_service);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `interledger_stream::server::PubStore` is not implemented for `interledger_store_memory::store::InMemoryStore`
    |
    = note: required by `interledger_stream::server::StreamReceiverService::<S, O, A>::new`

[clipped: another hundred lines of errors]
```

crates/interledger-store-redis/src/store.rs
bstrie: Can you elaborate on why tokio is fine with the first `Client::open` in the chain but not the second?

crates/interledger-store-redis/src/store.rs
bstrie: Er, I just wanted something unique to grep for in the logs while debugging. :P I'll reword this.

crates/interledger-store-redis/src/store.rs
bstrie: Can you briefly expound on what we want to be using each for? I've been using info for "here's what the code is currently doing" (basically equivalent to a code comment), debug for "here's the debug representation of something in memory", and no idea what trace ought to be used for.

crates/interledger-store-redis/src/store.rs
bstrie: I originally made this an error rather than a warn, but while running the test suite it causes a huge cascade of logged errors as nodes get rapidly spun up and torn down, obscuring all the output; warnings don't get logged to the console during test runs. Is there a solution we could use for that?

crates/interledger-stream/src/server.rs
bstrie: We'd still need those no-op implementations with an Option here, because Rust has no ability to know that certain stores will always be instantiated with None.

crates/interledger-stream/src/server.rs
bstrie: Did you have anything in mind for the type of error that ought to be returned?

bstrie:
Minor comments addressed, commits for the major comments on the way.

crates/interledger-stream/src/server.rs
emschwartz: Just `()` is fine for right now. We have a lot of places where we return `()`'s that will need to be cleaned up later

crates/interledger-api/src/lib.rs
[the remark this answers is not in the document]
emschwartz: Changed in https://github.com/interledger-rs/interledger-rs/pull/307/commits/fdca5af9d582bcdb5b7bc95b0816927ceafc6dcc

PR #308, opened 2019-09-17
docs(examples): fix log file names mismatch

darentuzi:
at some point the log file names got shortened in the example, step 2 doesn't match step 4

### DIFF
--- a/examples/eth-xrp-three-nodes/README.md
+++ b/examples/eth-xrp-three-nodes/README.md
@@ -314,7 +314,7 @@ cargo run --bin interledger-settlement-engines -- ethereum-ledger \
 --redis_url redis://127.0.0.1:6380/ \
 --asset_scale 6 \
 --settlement_api_bind_address 127.0.0.1:3000 \
-&> logs/node-alice-settlement-engine-eth.log &
+&> logs/redis-a-se-eth.log &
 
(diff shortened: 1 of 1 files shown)


examples/eth-xrp-three-nodes/README.md
```diff
@@ -314,7 +314,7 @@ cargo run --bin interledger-settlement-engines -- ethereum-ledger \
 --redis_url redis://127.0.0.1:6380/ \
 --asset_scale 6 \
 --settlement_api_bind_address 127.0.0.1:3000 \
-&> logs/node-alice-settlement-engine-eth.log &
+&> logs/redis-a-se-eth.log &
```
dora-gt: I don't understand why you are amending `node-alice-settlement-engine-eth.log` to `redis-a-se-eth.log`.
It's a log of a settlement engine. Could you explain why `redis`?

darentuzi: I got the names from step 2, the goal was to match the file names to have the logs in the same file but I might have understood it wrongly, in that case feel free to close the pull.

dora-gt: Thank you for the explanation 🙇 
My intention is to write into separate files, so I'll close this PR.

dora-gt:
As discussed above, I'll close this PR.

PR #311, opened 2019-09-18
Remove 26 transitive dependencies and shave 10 seconds off of `cargo build`

bstrie:
Just something I threw together while on the community call, it adds "default-features = false" to every dependency that we can get away with (many are currently no-ops, but guard against future feature additions), and removes a config line that was limiting our codegen-units for a reason that I couldn't determine. From looking at the development of various prominent crates (regex, tokio, etc) I believe that this will become a trend going forward for binaries that wish to shed transitive dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,7 +2,7 @@
 # It is not intended for manual editing.
 [[package]]
 name = "adler32"
-version = "1.0.3"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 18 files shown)


bstrie:
Regardless of whether or not this is something that we want to pursue, one thing to note is that now that we commit Cargo.lock we'll want to be careful when making any commits that add dependencies, since there's a good chance that will conflict with anyone else who's done the same in their branch.

bstrie:
I'll close this for now since it needs a little work to get all the tests passing and it's not a particular priority right now, but let's keep it in mind for when we start doing crate cleanup (and also after we do a thorough update of our dependencies).

PR #316, opened 2019-09-23
Separate crypto operations from redis

gakonst:
Removes any cryptography related operations from the underlying redis store implementation. The NodeStore implementer must perform any encryption related operations. The underlying store must operate only on encrypted data.

Part of https://github.com/interledger-rs/interledger-rs/issues/195 & https://github.com/interledger-rs/interledger-rs/issues/108

### DIFF
--- a/crates/interledger-api/src/lib.rs
+++ b/crates/interledger-api/src/lib.rs
@@ -80,6 +80,22 @@ pub struct AccountSettings {
     pub settle_to: Option<u64>,
 }
 
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct EncryptedAccountSettings {
+    // TODO these should be Secret types
+    pub http_incoming_token: Option<Bytes>,
+    pub btp_incoming_token: Option<Bytes>,
+    pub http_outgoing_token: Option<Bytes>,
+    pub btp_outgoing_token: Option<Bytes>,
+    pub http_endpoint: Option<String>,
+    pub btp_uri: Option<String>,
+    pub settle_threshold: Option<i64>,
+    // Note that this is intentionally an unsigned integer because users should
+    // not be able to set the settle_to value to be negative (meaning the node
+    // would pre-fund with the user)
+    pub settle_to: Option<u64>,
+}
+
 /// The Account type for the RedisStore.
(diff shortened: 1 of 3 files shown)


crates/interledger-api/src/lib.rs
```diff
@@ -80,6 +80,22 @@ pub struct AccountSettings {
     pub settle_to: Option<u64>,
 }
 
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct EncryptedAccountSettings {
```
gakonst: FWIW: I'd rather have this datatype and be explicit than overloading the `AccountSettings` struct's tokens to be used also for encrypted values.

emschwartz: Sounds good. Can you put a comment explaining what this struct is for?

crates/interledger-api/src/lib.rs
```diff
@@ -80,6 +80,22 @@ pub struct AccountSettings {
...
+    pub btp_incoming_token: Option<Bytes>,
+    pub http_outgoing_token: Option<Bytes>,
+    pub btp_outgoing_token: Option<Bytes>,
+    pub http_endpoint: Option<String>,
+    pub btp_uri: Option<String>,
+    pub settle_threshold: Option<i64>,
+    // Note that this is intentionally an unsigned integer because users should
+    // not be able to set the settle_to value to be negative (meaning the node
+    // would pre-fund with the user)
+    pub settle_to: Option<u64>,
```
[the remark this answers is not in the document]
gakonst: This is not encrypted, only the tokens are

gakonst:
> My original thought for separating the crypto from the store was that if we actually pulled it out of interledger-store-redis and didn't even give the store the encryption/decryption key, they store implementer wouldn't have to think about encrypting and decrypting at all. The values passed to the store traits would be encrypted already. That could make the API clunkier but it would ensure we standardize how crypto is being handled across DB integrations.

Yes that's exactly the goal. I have a follow up PR which completely refactors this (which we briefly discussed on our last call). This PR was meant just to separate the crypto functionality out of the `redis_*` functions.

emschwartz:
I'm going to merge this in. Sorry @gakonst if you wanted to make other changes to this first

Issue #39, opened 2019-03-08
Listen for incoming packets over BTP and HTTP on same port


emschwartz:
The BTP and HTTP servers both accept incoming packets/connections from some TCP listener. It should be possible to share the listener and expose one port for all incoming connections.

emschwartz:
I think if we want to do this, we'd need to switch the API, BTP server, and ILP-Over-HTTP server over to [`actix-web`](https://actix.rs/). It looks like that's the only web framework that currently supports handling WebSocket and HTTP messages over the same port.

emschwartz:
Actually, [`warp`](https://github.com/seanmonstar/warp) supports this as well and looks like it would be a smaller change than switching to `actix-web`

PR #318, opened 2019-09-23
Switch BTP to warp

emschwartz:
Resolves https://github.com/interledger-rs/interledger-rs/issues/39. BTP now listens on the same port as the HTTP. This removes the need for the `btp_bind_address` configuration option too.

This also removes the `BtpOpenSignupStore` because we are not going to be supporting BTP signup in the Interledger.rs node.

~~Note the base is set to https://github.com/interledger-rs/interledger-rs/pull/276 so that one should be merged first.~~

EDIT: The base is now set to https://github.com/interledger-rs/interledger-rs/pull/316, which should be merged first.


emschwartz:
I'm not sure why that test is failing but I'll look into that more tomorrow

crates/interledger-btp/src/server.rs
```diff
@@ -33,121 +21,133 @@ const MAX_MESSAGE_SIZE: usize = 40000;
 /// The separation is designed to enable the returned BtpOutgoingService to be passed
 /// to another service like the Router, and _then_ for the Router to be passed as the
 /// IncomingService to the BTP server.
-pub fn create_server<T, U, A>(
+pub fn btp_server<O, S, A>(
```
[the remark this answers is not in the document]
dora-gt: Maybe because it just returns a filter without binding? so what about `create_btp_service_and_filter`?

crates/interledger/src/node.rs
```diff
@@ -198,117 +192,122 @@ impl InterledgerNode {
...
-                                    if let Some(username) = default_spsp_account {
-                                        api.default_spsp_account(username);
-                                    }
-                                    spawn(api.bind(http_bind_address));
+                            // Mount the BTP endpoint at /btp
+                            let btp_endpoint = warp::path("btp")
+                                .and(warp::path::end())
+                                .and(btp_filter);
+                            // Note that other endpoints added to the API must come first
+                            // because the API includes error handling and consumes the request.
```
dora-gt: Just a question, what does `error handling` mean? like "not found" (the path was not found) error?

crates/interledger-store-redis/src/store.rs
```diff
@@ -714,13 +714,16 @@ impl AccountStore for RedisStore {
...
+        let username = username.clone();
         Box::new(
             cmd("HGET")
                 .arg("usernames")
                 .arg(username.as_ref())
                 .query_async(self.connection.as_ref().clone())
                 .map_err(move |err| error!("Error getting account id: {:?}", err))
-                .and_then(|(_connection, id): (_, AccountId)| Ok(id)),
+                .and_then(|(_connection, id): (_, Option<AccountId>)| {
+                    id.ok_or_else(move || debug!("Username not found: {}", username))
```
gakonst: Good catch thanks.

crates/interledger/src/node.rs
```diff
@@ -198,117 +192,122 @@ impl InterledgerNode {
...
+                            // TODO should we just make BTP part of the API?
```
gakonst: I believe this is a good call, since now they are in the same port.

crates/interledger-btp/src/server.rs
```diff
@@ -1,30 +1,18 @@
...
+use warp::{
+    self,
+    ws::{Message, WebSocket, Ws2},
+    Error, Filter,
 };
 
-type WsStream = WebSocketStream<MaybeTlsStream<TcpStream>>;
-
-const MAX_MESSAGE_SIZE: usize = 40000;
+// const MAX_MESSAGE_SIZE: usize = 40000;
```
gakonst: What is this for? Can we remove it?

crates/interledger-btp/src/server.rs
```diff
@@ -33,121 +21,133 @@ const MAX_MESSAGE_SIZE: usize = 40000;
...
+    BtpOutgoingService<O, A>,
+    warp::filters::BoxedFilter<(impl warp::Reply,)>,
+)
 where
-    T: OutgoingService<A> + Clone + Send + Sync + 'static,
-    U: BtpStore<Account = A> + Clone + Send + Sync + 'static,
-    A: BtpAccount + 'static,
+    O: OutgoingService<A> + Clone + Send + Sync + 'static,
+    S: BtpStore<Account = A> + Clone + Send + Sync + 'static,
+    A: BtpAccount + Account + 'static,
```
gakonst: Since BtpAccount extends Account this should be possible:
 
```suggestion
    A: BtpAccount + 'static,
```

crates/interledger-btp/src/server.rs
```diff
@@ -33,121 +21,133 @@ const MAX_MESSAGE_SIZE: usize = 40000;
...
+                            Err(WsError::Tungstenite(tungstenite::Error::Utf8))
+                        }
+                    }
+                    Ok(AsyncSink::Ready) => Ok(AsyncSink::Ready),
+                    Err(err) => Err(WsError::from(err)),
+                }
+            }
+            // Ignore other message types because warp's WebSocket type doesn't
+            // allow us to send any other types of messages
+            // TODO make sure warp's websocket responds to pings and/or sends them to keep the
```
gakonst: This should probably be checked in this PR, we wouldn't want to have a BTP connection drop if it's been inactive for a while

crates/interledger-btp/src/server.rs
```diff
@@ -303,45 +225,47 @@ fn get_auth(
...
+                    let mut token: Option<String> = None;
+                    for protocol_data in message.protocol_data.iter() {
+                        match (
+                            protocol_data.protocol_name.as_ref(),
+                            !protocol_data.data.is_empty(),
+                        ) {
+                            ("auth_token", _) => {
+                                token = String::from_utf8(protocol_data.data.clone()).ok()
+                            }
+                            ("auth_username", true) => {
```
gakonst: `auth_username` is not used anywhere at the moment, do we need it?

gakonst:
FYI: The test is failing due to some flakiness I've introduced in the ildcp-parent branch. I am investigating it, it's unrelated to any of your changes.

gakonst:
Rebased this on #316

crates/interledger/src/node.rs
```diff
@@ -198,117 +192,122 @@ impl InterledgerNode {
...
-                                        store.clone(),
-                                        incoming_service.clone(),
-                                        outgoing_service.clone(),
-                                        btp.clone(),
-                                    );
-                                    if let Some(username) = default_spsp_account {
-                                        api.default_spsp_account(username);
-                                    }
-                                    spawn(api.bind(http_bind_address));
+                            // Mount the BTP endpoint at /btp
```
bstrie: Should we consider namespacing these protocol-level endpoints under something other than root?

crates/interledger-btp/src/server.rs
```diff
@@ -33,121 +21,133 @@ const MAX_MESSAGE_SIZE: usize = 40000;
...
+        .boxed();
+    (service, filter)
+}
 
-        Ok(service)
-    })
+/// This wraps a warp Websocket connection to make it act like a
+/// tungstenite Websocket connection. It is needed for
+/// compatibility with the BTP service that interacts with the
+/// websocket implementation from warp and tokio-tungstenite
```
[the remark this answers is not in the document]
emschwartz: The problem is that `warp` is only on the server side, while the `service` deals with client and server connections. We can only make the client connections using `tokio-tungstenite`. I couldn't find a cleaner way of turning one set of errors into the other

crates/interledger-api/Cargo.toml
```diff
@@ -21,6 +21,8 @@ interledger-service-util = { path = "../interledger-service-util", version = "^0
 interledger-settlement = { path = "../interledger-settlement", version = "^0.1.1-alpha.1"}
 interledger-spsp = { path = "../interledger-spsp", version = "^0.2.2-alpha.1"}
 interledger-stream = { path = "../interledger-stream", version = "^0.2.2-alpha.1" }
+interledger-ccp = { path = "../interledger-ccp", version = "^0.1.1-alpha.1"}
+interledger-btp = { path = "../interledger-btp", version = "^0.2.2-alpha.1"}
```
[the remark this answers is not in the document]
bstrie: I'll resolve this in favor of tracking discussion at https://github.com/interledger-rs/interledger-rs/issues/325.

crates/interledger/src/node.rs
emschwartz: [This](https://github.com/interledger-rs/interledger-rs/blob/2a12b20a9bde92b0e1fc1929823f6d738554c82f/crates/interledger-api/src/lib.rs#L206-L220) is what I'm referring to. Before, I put the API before the `btp_filter` and ran into [this error](https://github.com/seanmonstar/warp/blob/c605336891547d2093d8c0c6b4871c7c69c8d4d4/src/filters/body.rs#L28-L29)

crates/interledger/src/node.rs
emschwartz: I think we might actually want to go the other way and move things out of the API: https://github.com/interledger-rs/interledger-rs/issues/325

crates/interledger/src/node.rs
emschwartz: It _could_ be `/ilp/btp`. That would make more sense to me if we implemented https://github.com/interledger-rs/interledger-rs/issues/326 and put it under `/ilp/ws` or `/ilp/websocket`

crates/interledger-btp/src/server.rs
emschwartz: This is dependent on https://github.com/seanmonstar/warp/issues/269 (either enabling auto-ping or exposing the ping and pong message types)

crates/interledger-btp/src/server.rs
emschwartz: We should parse the username out of it if it's there

crates/interledger/src/node.rs
bstrie: Would you leave `/ilp` where it is, or would it become `/ilp/http`?

emschwartz: I would leave it as `/ilp` because it's an HTTP API so it seems redundant to put that in the name

dora-gt:
🎉

PR #324, opened 2019-09-24
Improve Error Responses

dora-gt:
Update HTTP APIs including ILP over HTTP to use `ApiError` to output formal RFC7807 errors, which improves the developer experience of `interledger-rs`, providing more understandable error messages.

<details>
<summary>The original message</summary>

**This is not actually working**, just a concept level code to confirm the direction.

My thoughts:
- APIs return `ApiErrorStruct` (would be renamed to `ApiError`) which represents `Problem` of the [RFC7807](https://tools.ietf.org/html/rfc7807).
    - `ApiErrorStruct` could be reused in the `recover` section as it implements `Reply`.
- The top-level lib's knowing every error looks a bit weird. They should not know every API specific errors. So only a very primitive struct is defined in the top-level lib and APIs utilize it.
- `ApiErrorStruct` can not handle [extension members](https://tools.ietf.org/html/rfc7807#section-3.2) but it looks sufficient so far.

What do you think?
</details>

### DIFF
--- a/crates/interledger-api/Cargo.toml
+++ b/crates/interledger-api/Cargo.toml
@@ -30,6 +30,10 @@ reqwest = { version = "0.9.21", default-features = false }
 url = { version = "2.1.0", default-features = false, features = ["serde"] }
 warp = { version = "0.1.20", default-features = false }
 secrecy = { version = "0.4.0", default-features = false, features = ["serde"] }
+chrono = { version = "0.4.9", default-features = false }
+serde_path_to_error = { version = "0.1", default-features = false }
+regex = { version ="1.3.1", default-features = false, features = ["std"] }
+lazy_static = { version ="1.4.0", default-features = false }
 
 [badges]
 circle-ci = { repository = "interledger-rs/interledger-rs" }
--- a/crates/interledger-api/Cargo.toml
+++ b/crates/interledger-api/Cargo.toml
@@ -35,6 +35,17 @@ serde_path_to_error = { version = "0.1", default-features = false }
 regex = { version ="1.3.1", default-features = false, features = ["std"] }
 lazy_static = { version ="1.4.0", default-features = false }
 
+[dev-dependencies]
+ilp-node = { path = "../ilp-node", version = "^0.4.1-beta.2"}
+interledger = { path = "../interledger", version = "^0.4.1-alpha.1" }
+env_logger = "0.6.1"
+net2 = "0.2.33"
+rand = "0.6.5"
+redis = "0.12.0"
+tokio = "0.1.20"
+base64 = "0.10.1"
+approx = "0.3.2"
+
 [badges]
 circle-ci = { repository = "interledger-rs/interledger-rs" }
 codecov = { repository = "interledger-rs/interledger-rs" }
--- a/crates/interledger-api/Cargo.toml
+++ b/crates/interledger-api/Cargo.toml
@@ -35,13 +35,14 @@ secrecy = { version = "0.4.0", default-features = false, features = ["serde"] }
 [dev-dependencies]
 ilp-node = { path = "../ilp-node", version = "^0.4.1-beta.2"}
 interledger = { path = "../interledger", version = "^0.4.1-alpha.1" }
-env_logger = "0.6.1"
-net2 = "0.2.33"
-rand = "0.6.5"
-redis = "0.13.0"
-tokio = "0.1.20"
-base64 = "0.10.1"
-approx = "0.3.2"
+env_logger = { version = "0.6.1", default-features = false }
+net2 = { version = "0.2.33", default-features = false }
+rand = { version = "0.6.5", default-features = false }
+redis = { version = "0.13.0", default-features = false }
+tokio = { version = "0.1.20", default-features = false }
+base64 = { version = "0.10.1", default-features = false }
+approx = { version = "0.3.2", default-features = false }
+ring = { version = "0.14.6", default-features = false }
 
 [badges]
 circle-ci = { repository = "interledger-rs/interledger-rs" }
--- a/crates/interledger-api/Cargo.toml
+++ b/crates/interledger-api/Cargo.toml
@@ -26,14 +26,11 @@ interledger-btp = { path = "../interledger-btp", version = "^0.2.2-alpha.1", def
 log = { version = "0.4.8", default-features = false }
 serde = { version = "1.0.101", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.41", default-features = false }
+serde_path_to_error = { version = "0.1", default-features = false }
 reqwest = { version = "0.9.21", default-features = false }
 url = { version = "2.1.0", default-features = false, features = ["serde"] }
 warp = { version = "0.1.20", default-features = false }
 secrecy = { version = "0.4.0", default-features = false, features = ["serde"] }
-chrono = { version = "0.4.9", default-features = false }
-serde_path_to_error = { version = "0.1", default-features = false }
-regex = { version ="1.3.1", default-features = false, features = ["std"] }
-lazy_static = { version ="1.4.0", default-features = false }
 
 [dev-dependencies]
 ilp-node = { path = "../ilp-node", version = "^0.4.1-beta.2"}
--- a/crates/interledger-api/Cargo.toml
+++ b/crates/interledger-api/Cargo.toml
@@ -38,7 +38,7 @@ interledger = { path = "../interledger", version = "^0.4.1-alpha.1" }
 env_logger = "0.6.1"
 net2 = "0.2.33"
 rand = "0.6.5"
-redis = "0.12.0"
+redis = "0.13.0"
 tokio = "0.1.20"
 base64 = "0.10.1"
 approx = "0.3.2"
--- a/crates/interledger-api/src/error.rs
+++ b/crates/interledger-api/src/error.rs
@@ -1,0 +1,356 @@
+use chrono::{DateTime, Local};
+use http::header::HeaderValue;
+use lazy_static::lazy_static;
+use regex::Regex;
+use serde::{ser::Serializer, Serialize};
+use serde_json::error::Category;
+use serde_json::{Map, Value};
+use std::{
+    error::Error as StdError,
+    fmt::{self, Display},
+};
+use warp::{reject::custom, reply::json, reply::Response, Rejection, Reply};
+
+/// This struct represents the fields defined in [RFC7807](https://tools.ietf.org/html/rfc7807).
+/// The meaning of each field could be found at [Members of a Problem Details Object](https://tools.ietf.org/html/rfc7807#section-3.1) section.
+/// ApiError implements Reply so that it could be used for responses.
+#[derive(Clone, Debug, Serialize)]
+pub(crate) struct ApiError {
+    /// `type` is a URI which represents an error type. The URI should provide human-readable
+    /// documents so that developers could solve the problem easily.
+    #[serde(serialize_with = "serialize_type")]
+    pub r#type: &'static ProblemType,
+    /// `title` is a short, human-readable summary of the type.
+    /// SHOULD NOT change from occurrence to occurrence of the problem, except for purposes
+    /// of localization.
+    pub title: &'static str,
+    /// `status` is a HTTP status of the problem.
+    #[serde(serialize_with = "serialize_status_code")]
+    pub status: http::StatusCode,
+    /// `detail` explains the problem in human-readable detail.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub detail: Option<String>,
+    /// `instance` is a URI reference that identifies the specific occurrence of the problem.
+    /// We should be careful of how we provide the URI because if it provides very detailed
+    /// information about the error, it might expose some vulnerabilities of the node.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub instance: Option<String>,
+    /// `extension_members` is a Map of JSON values which will be flatly injected into response
+    /// JSONs. For example, if we specify `extension_members` like:
+    /// ```JSON
+    /// "invalid-params": [
+    ///     { "name": "Username", "type": "missing" }
+    /// ]
+    /// ```
+    /// then this map is merged into the response JSON and will look like:
+    /// ```JSON
+    /// {
+    ///     "type": "about:blank",
+    ///     "title": "Missing Fields Error",
+    ///     "status": 400,
+    ///     "detail": "foo bar",
+    ///     "invalid-params": [
+    ///         { "name": "Username", "type": "missing" }
+    ///     ]
+    /// }
+    /// ```
+    #[serde(flatten, skip_serializing_if = "Option::is_none")]
+    pub extension_members: Option<Map<String, Value>>,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub enum ProblemType {
+    /// `Default` is a [pre-defined value](https://tools.ietf.org/html/rfc7807#section-4.2) which is
+    /// going to be serialized as `about:blank`.
+    Default,
+    /// InterledgerHttpApi is a API specific error type which is going to be serialized like
+    /// `https://errors.interledger.org/http-api/foo-bar`. Variant means path, in the example,
+    /// `foo-bar` is the path.
+    InterledgerHttpApi(&'static str),
+}
+
+pub struct ApiErrorType {
+    pub r#type: &'static ProblemType,
+    pub title: &'static str,
+}
+
+// default errors
+const ERROR_TYPE_PREFIX: &str = "https://errors.interledger.org/http-api";
+const DEFAULT_BAD_REQUEST_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Bad Request",
+};
+const DEFAULT_INTERNAL_SERVER_ERROR_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Internal Server Error",
+};
+const DEFAULT_UNAUTHORIZED_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Unauthorized",
+};
+const DEFAULT_NOT_FOUND_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Not Found",
+};
+const DEFAULT_METHOD_NOT_ALLOWED_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Method Not Allowed",
+};
+
+// This should be OK because serde serializer MUST be `fn<S>(&T, S)`
+#[allow(clippy::trivially_copy_pass_by_ref)]
+fn serialize_status_code<S>(status: &http::StatusCode, s: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    s.serialize_u16(status.as_u16())
+}
+
+fn serialize_type<S>(r#type: &ProblemType, s: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    match r#type {
+        ProblemType::Default => s.serialize_str("about:blank"),
+        ProblemType::InterledgerHttpApi(custom_type) => {
+            s.serialize_str(&format!("{}/{}", ERROR_TYPE_PREFIX, custom_type))
+        }
+    }
+}
+
+impl ApiError {
+    #[allow(dead_code)]
+    pub fn default_bad_request() -> Self {
+        ApiError::bad_request(&DEFAULT_BAD_REQUEST_TYPE, None, None, None)
+    }
+
+    pub fn bad_request(
+        problem_type: &ApiErrorType,
+        detail: Option<String>,
+        instance: Option<String>,
+        extension_members: Option<Map<String, Value>>,
+    ) -> Self {
+        ApiError {
+            r#type: problem_type.r#type,
+            title: problem_type.title,
+            status: http::StatusCode::BAD_REQUEST,
+            detail,
+            instance,
+            extension_members: Some(ApiError::merge_default_extension_members(extension_members)),
+        }
+    }
+
+    pub fn default_internal_server_error() -> Self {
+        ApiError::internal_server_error(&DEFAULT_INTERNAL_SERVER_ERROR_TYPE, None, None, None)
+    }
+
+    pub fn internal_server_error(
+        problem_type: &ApiErrorType,
+        detail: Option<String>,
+        instance: Option<String>,
+        extension_members: Option<Map<String, Value>>,
+    ) -> Self {
+        ApiError {
+            r#type: problem_type.r#type,
+            title: problem_type.title,
+            status: http::StatusCode::INTERNAL_SERVER_ERROR,
+            detail,
+            instance,
+            extension_members: Some(ApiError::merge_default_extension_members(extension_members)),
+        }
+    }
+
+    pub fn default_unauthorized() -> Self {
+        ApiError::unauthorized(&DEFAULT_UNAUTHORIZED_TYPE, None, None, None)
+    }
+
+    pub fn unauthorized(
+        problem_type: &ApiErrorType,
+        detail: Option<String>,
+        instance: Option<String>,
+        extension_members: Option<Map<String, Value>>,
+    ) -> Self {
+        ApiError {
+            r#type: problem_type.r#type,
+            title: problem_type.title,
+            status: http::StatusCode::UNAUTHORIZED,
+            detail,
+            instance,
+            extension_members: Some(ApiError::merge_default_extension_members(extension_members)),
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn default_not_found() -> Self {
+        ApiError::not_found(&DEFAULT_NOT_FOUND_TYPE, None, None, None)
+    }
+
+    pub fn not_found(
+        problem_type: &ApiErrorType,
+        detail: Option<String>,
+        instance: Option<String>,
+        extension_members: Option<Map<String, Value>>,
+    ) -> Self {
+        ApiError {
+            r#type: problem_type.r#type,
+            title: problem_type.title,
+            status: http::StatusCode::NOT_FOUND,
+            detail,
+            instance,
+            extension_members: Some(ApiError::merge_default_extension_members(extension_members)),
+        }
+    }
+
+    pub fn default_method_not_allowed() -> Self {
+        ApiError::method_not_allowed(&DEFAULT_METHOD_NOT_ALLOWED_TYPE, None, None, None)
+    }
+
+    pub fn method_not_allowed(
+        problem_type: &ApiErrorType,
+        detail: Option<String>,
+        instance: Option<String>,
+        extension_members: Option<Map<String, Value>>,
+    ) -> Self {
+        ApiError {
+            r#type: problem_type.r#type,
+            title: problem_type.title,
+            status: http::StatusCode::METHOD_NOT_ALLOWED,
+            detail,
+            instance,
+            extension_members: Some(ApiError::merge_default_extension_members(extension_members)),
+        }
+    }
+
+    fn get_base_extension_members() -> Map<String, Value> {
+        // TODO Should implement request wide time
+        let datetime: DateTime<Local> = Local::now();
+        let mut map = serde_json::Map::new();
+        map.insert("datetime".to_owned(), Value::from(datetime.to_rfc3339()));
+        map
+    }
+
+    fn merge_default_extension_members(
+        extension_members: Option<Map<String, Value>>,
+    ) -> Map<String, Value> {
+        let mut merged_extension_members = ApiError::get_base_extension_members();
+        if let Some(map) = extension_members {
+            for (k, v) in map {
+                merged_extension_members.insert(k, v);
+            }
+        }
+        merged_extension_members
+    }
+}
+
+impl Display for ApiError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_fmt(format_args!("{:?}", self))
+    }
+}
+
+impl Reply for ApiError {
+    fn into_response(self) -> Response {
+        let res = json(&self);
+        let mut res = res.into_response();
+        *res.status_mut() = self.status;
+        res.headers_mut().insert(
+            "Content-Type",
+            HeaderValue::from_static("application/problem+json"),
+        );
+        res
+    }
+}
+
+impl StdError for ApiError {}
+
+impl From<ApiError> for Rejection {
+    fn from(from: ApiError) -> Self {
+        custom(from)
+    }
+}
+
+lazy_static! {
+    static ref MISSING_FIELD_REGEX: Regex = Regex::new("missing field `(.*)`").unwrap();
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct JsonDeserializeError {
+    pub(crate) category: Category,
+    pub(crate) detail: String,
+    pub(crate) path: serde_path_to_error::Path,
+}
+
+impl StdError for JsonDeserializeError {}
+
+impl Display for JsonDeserializeError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_fmt(format_args!("{:?}", self))
+    }
+}
+
+// JSON deserialization errors
+const JSON_SYNTAX_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("json-syntax"),
+    title: "JSON Syntax Error",
+};
+const JSON_DATA_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("json-data"),
+    title: "JSON Data Error",
+};
+const UNKNOWN_JSON_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("json-unknown"),
+    title: "Unknown JSON Error",
+};
+
+impl Reply for JsonDeserializeError {
+    fn into_response(self) -> Response {
+        let mut extension_member = Map::new();
+
+        // invalid-params should be a plural form even if it is always an array with a single value
+        // for the future extendability.
+
+        // if `path` has segments and the first value is not Unknown
+        if let Some(segment) = self.path.iter().next() {
+            match segment {
+                serde_path_to_error::Segment::Unknown => {}
+                _ => {
+                    let invalid_params = serde_json::json!([ { "name": self.path.to_string() } ]);
+                    extension_member.insert("invalid-params".to_string(), invalid_params);
+                }
+            }
+        }
+
+        // if detail contains missing field error
+        // it seems that there is no way to handle this cleanly
+        if let Some(captures) = MISSING_FIELD_REGEX.captures(&self.detail) {
+            if let Some(r#match) = captures.get(1) {
+                let invalid_params =
+                    serde_json::json!([ { "name": r#match.as_str(), "type": "missing" } ]);
+                extension_member.insert("invalid-params".to_string(), invalid_params);
+            }
+        }
+
+        let api_error_type = match self.category {
+            Category::Syntax => &JSON_SYNTAX_TYPE,
+            Category::Data => &JSON_DATA_TYPE,
+            _ => &UNKNOWN_JSON_TYPE,
+        };
+
+        ApiError::bad_request(
+            api_error_type,
+            Some(self.detail),
+            None,
+            match extension_member.keys().len() {
+                0 => None,
+                _ => Some(extension_member),
+            },
+        )
+        .into_response()
+    }
+}
+
+impl From<JsonDeserializeError> for Rejection {
+    fn from(from: JsonDeserializeError) -> Self {
+        custom(from)
+    }
+}
--- a/crates/interledger-api/src/error/error_types.rs
+++ b/crates/interledger-api/src/error/error_types.rs
@@ -1,0 +1,63 @@
+// APIs should implement their own `ApiErrorType`s to provide more detailed information
+// about what were the problem, for example, `JSON_SYNTAX_TYPE` or `ACCOUNT_NOT_FOUND_TYPE`.
+
+use super::{ApiErrorType, ProblemType};
+
+// Default errors
+#[allow(dead_code)]
+pub(crate) const DEFAULT_BAD_REQUEST_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Bad Request",
+    status: http::StatusCode::BAD_REQUEST,
+};
+pub(crate) const DEFAULT_INTERNAL_SERVER_ERROR_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Internal Server Error",
+    status: http::StatusCode::INTERNAL_SERVER_ERROR,
+};
+pub(crate) const DEFAULT_UNAUTHORIZED_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Unauthorized",
+    status: http::StatusCode::UNAUTHORIZED,
+};
+pub(crate) const DEFAULT_NOT_FOUND_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Not Found",
+    status: http::StatusCode::NOT_FOUND,
+};
+pub(crate) const DEFAULT_METHOD_NOT_ALLOWED_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Method Not Allowed",
+    status: http::StatusCode::METHOD_NOT_ALLOWED,
+};
+
+// JSON deserialization errors
+pub(crate) const JSON_SYNTAX_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("json-syntax"),
+    title: "JSON Syntax Error",
+    status: http::StatusCode::BAD_REQUEST,
+};
+pub(crate) const JSON_DATA_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("json-data"),
+    title: "JSON Data Error",
+    status: http::StatusCode::BAD_REQUEST,
+};
+pub(crate) const UNKNOWN_JSON_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("json-unknown"),
+    title: "Unknown JSON Error",
+    status: http::StatusCode::BAD_REQUEST,
+};
+
+// Account specific errors
+pub(crate) const ACCOUNT_NOT_FOUND_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("accounts/account-not-found"),
+    title: "Account Not Found",
+    status: http::StatusCode::NOT_FOUND,
+};
+
+// Node settings specific errors
+pub(crate) const INVALID_ACCOUNT_ID_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("settings/invalid-account-id"),
+    title: "Invalid Account Id",
+    status: http::StatusCode::BAD_REQUEST,
+};
--- a/crates/interledger-api/src/lib.rs
+++ b/crates/interledger-api/src/lib.rs
@@ -8,20 +8,18 @@ use interledger_service_util::{BalanceStore, ExchangeRateStore};
 use interledger_settlement::{SettlementAccount, SettlementStore};
 use interledger_stream::StreamNotificationsStore;
 use serde::{de, Deserialize, Serialize};
-use std::{
-    collections::HashMap,
-    error::Error as StdError,
-    fmt::{self, Display},
-    net::SocketAddr,
-    str::FromStr,
-};
+use std::{fmt::Display, net::SocketAddr, str::FromStr, collections::HashMap};
 use warp::{self, Filter};
 mod routes;
+use bytes::Buf;
 use interledger_btp::{BtpAccount, BtpOutgoingService};
 use interledger_ccp::CcpRoutingAccount;
 use secrecy::SecretString;
+use serde::de::DeserializeOwned;
 use url::Url;
+use warp::{filters::body::FullBody, Rejection, Reply};
 
+pub(crate) mod error;
 pub(crate) mod http_retry;
 
 // This enum and the following functions are used to allow clients to send either
@@ -207,27 +205,6 @@ pub struct AccountDetails {
     pub settlement_engine_url: Option<String>,
 }
 
-#[derive(Clone, Copy, Debug)]
-pub(crate) enum ApiError {
-    AccountNotFound,
-    BadRequest,
-    InternalServerError,
-    Unauthorized,
-}
-
-impl Display for ApiError {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.write_str(match self {
-            ApiError::AccountNotFound => "Account not found",
-            ApiError::BadRequest => "Bad request",
-            ApiError::InternalServerError => "Internal server error",
-            ApiError::Unauthorized => "Unauthorized",
-        })
-    }
-}
-
-impl StdError for ApiError {}
-
 pub struct NodeApi<S, I, O, B, A: Account> {
     store: S,
     admin_api_token: String,
@@ -306,16 +283,12 @@ where
             ))
             .or(routes::node_settings_api(self.admin_api_token, self.store))
             .recover(|err: warp::Rejection| {
-                if let Some(&err) = err.find_cause::<ApiError>() {
-                    let code = match err {
-                        ApiError::AccountNotFound => warp::http::StatusCode::NOT_FOUND,
-                        ApiError::BadRequest => warp::http::StatusCode::BAD_REQUEST,
-                        ApiError::InternalServerError => {
-                            warp::http::StatusCode::INTERNAL_SERVER_ERROR
-                        }
-                        ApiError::Unauthorized => warp::http::StatusCode::UNAUTHORIZED,
-                    };
-                    Ok(warp::reply::with_status(warp::reply(), code))
+                if let Some(api_error) = err.find_cause::<error::ApiError>() {
+                    Ok(api_error.clone().into_response())
+                } else if let Some(json_error) = err.find_cause::<error::JsonDeserializeError>() {
+                    Ok(json_error.clone().into_response())
+                } else if err.status() == http::status::StatusCode::METHOD_NOT_ALLOWED {
+                    Ok(error::ApiError::default_method_not_allowed().into_response())
                 } else {
                     Err(err)
                 }
@@ -326,6 +299,22 @@ where
     pub fn bind(self, addr: SocketAddr) -> impl Future<Item = (), Error = ()> {
         warp::serve(self.into_warp_filter()).bind(addr)
     }
+}
+
+fn deserialize_json<T: DeserializeOwned + Send>(
+) -> impl Filter<Extract = (T,), Error = Rejection> + Copy {
+    warp::header::exact("content-type", "application/json")
+        .and(warp::body::concat())
+        .and_then(|buf: FullBody| {
+            let deserializer = &mut serde_json::Deserializer::from_slice(&buf.bytes());
+            serde_path_to_error::deserialize(deserializer).map_err(|err| {
+                warp::reject::custom(error::JsonDeserializeError {
+                    category: err.inner().classify(),
+                    detail: err.inner().to_string(),
+                    path: err.path().clone(),
+                })
+            })
+        })
 }
 
 #[cfg(test)]
--- a/crates/interledger-api/src/lib.rs
+++ b/crates/interledger-api/src/lib.rs
@@ -8,10 +8,11 @@ use interledger_service_util::{BalanceStore, ExchangeRateStore};
 use interledger_settlement::{SettlementAccount, SettlementStore};
 use interledger_stream::StreamNotificationsStore;
 use serde::{de, Deserialize, Serialize};
-use std::{fmt::Display, net::SocketAddr, str::FromStr, collections::HashMap};
+use std::{collections::HashMap, fmt::Display, net::SocketAddr, str::FromStr};
 use warp::{self, Filter};
 mod routes;
 use bytes::Buf;
+use error::*;
 use interledger_btp::{BtpAccount, BtpOutgoingService};
 use interledger_ccp::CcpRoutingAccount;
 use secrecy::SecretString;
@@ -283,12 +284,15 @@ where
             ))
             .or(routes::node_settings_api(self.admin_api_token, self.store))
             .recover(|err: warp::Rejection| {
-                if let Some(api_error) = err.find_cause::<error::ApiError>() {
+                if let Some(api_error) = err.find_cause::<ApiError>() {
                     Ok(api_error.clone().into_response())
-                } else if let Some(json_error) = err.find_cause::<error::JsonDeserializeError>() {
+                } else if let Some(json_error) = err.find_cause::<JsonDeserializeError>() {
                     Ok(json_error.clone().into_response())
                 } else if err.status() == http::status::StatusCode::METHOD_NOT_ALLOWED {
-                    Ok(error::ApiError::default_method_not_allowed().into_response())
+                    Ok(
+                        ApiError::from_api_error_type(&DEFAULT_METHOD_NOT_ALLOWED_TYPE)
+                            .into_response(),
+                    )
                 } else {
                     Err(err)
                 }
@@ -308,7 +312,7 @@ fn deserialize_json<T: DeserializeOwned + Send>(
         .and_then(|buf: FullBody| {
             let deserializer = &mut serde_json::Deserializer::from_slice(&buf.bytes());
             serde_path_to_error::deserialize(deserializer).map_err(|err| {
-                warp::reject::custom(error::JsonDeserializeError {
+                warp::reject::custom(JsonDeserializeError {
                     category: err.inner().classify(),
                     detail: err.inner().to_string(),
                     path: err.path().clone(),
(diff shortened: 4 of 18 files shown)


crates/interledger-api/src/lib.rs
```diff
@@ -126,6 +126,31 @@ impl Display for ApiError {
 
 impl StdError for ApiError {}
 
+use warp::{Reply, reply::Response};
+
+#[derive(Clone, Debug)]
+pub struct ApiErrorStruct {
+    pub r#type: &'static str,
```
emschwartz: What about calling it `error_type` so we don't need to use `r#` everywhere?

emschwartz:
> The top-level lib's knowing every error looks a bit weird. They should not know every API specific errors. So only a very primitive struct is defined in the top-level lib and APIs utilize it.

I'm not sure I agree with this. Having every API error defined in one place would make it really easy to see the different errors the API can produce. Also, that would allow us to make each error an Enum variant so we could write `ApiError::InvalidParameter("missing username")` in the code and the implementation of the `ApiError` would put the correct `type`, `title`, and `code`. It's better to write that mapping once rather than need to put the right fields every time an error is created

gakonst:
> Also, that would allow us to make each error an Enum variant so we could write ApiError::InvalidParameter("missing username")

I agree with this. Let's have an API error enum which has specific `type`, `title` and `code`, based on the enum variant, and takes a string as input for the error message.

Also, could you rebase this?

dora-gt:
OK, I'll try to implement what you guys suggest. I'm not sure it works cleanly but let me think loosely coupling and useful structure. I mean, I think it is true that some kind of common errors should be pre-defined but some cannot be pre-defined as they are known only to some specific APIs.

dora-gt:
Hmmm I tried a few approaches but could not come up with a beautiful structure.

Because `enum` cannot be extended later after the definition, I had to place many enum variants. That looked almost the same as defining `structure`. So I finally end up with the pure `structure` with some utils which generate the struct.

I think this is the most extendable and loosely coupled structure.

dora-gt:
I tried to use [path-to-error](https://github.com/dtolnay/path-to-error) but I couldn't figure out how to make the trait compatible...anyway, the current implementation outputs like:

```
{
  "type": "https://errors.interledger.org/http-api/json-data",
  "title": "JSON Data Error",
  "status": 400,
  "detail": "missing field `username` at line 8 column 20",
  "datetime": "2019-09-28T16:15:09.659122+09:00"
}

{
  "type": "https://errors.interledger.org/http-api/json-syntax",
  "title": "JSON Syntax Error",
  "status": 400,
  "detail": "trailing comma at line 8 column 21",
  "datetime": "2019-09-28T16:15:40.066371+09:00"
}

{
  "type": "https://errors.interledger.org/http-api/unauthorized",
  "title": "Unauthorized",
  "status": 401,
  "datetime": "2019-09-28T16:16:18.347859+09:00"
}
```

crates/interledger-api/src/routes/node_settings.rs
```diff
@@ -62,14 +63,16 @@ where
...
-        .and(warp::body::json())
+        .and(deserialize_json())
         .and(with_store.clone())
         .and_then(|rates: HashMap<String, f64>, store: S| {
             if store.set_exchange_rates(rates.clone()).is_ok() {
                 Ok(warp::reply::json(&rates))
             } else {
                 error!("Error setting exchange rates");
-                Err(warp::reject::custom(ApiError::InternalServerError))
+                Err(warp::reject::custom(
```
emschwartz: What do you think about implementing `warp::Rejection` for `ApiError` so that we don't need to write out `warp::reject::custom` everywhere?

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,196 @@
...
+            extension_members: Some(ApiError::merge_default_extension_members(extension_members)),
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn create_default_not_found() -> Self {
+        ApiError::create_unauthorized(ApiErrorType::Default, "Not Found", None, None, None)
+    }
+
+    pub fn create_not_found(
```
gakonst: `type` should be "Default and `title` should be a "not found" message. Not sure what `instance` means. Also, what is the use of `detail`, if the error variant is `Custom`?

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -141,15 +144,17 @@ where
...
         .and(with_store.clone())
         .and_then(move |account_details: AccountDetails, store: S| {
             let store_clone = store.clone();
             let handler = outgoing_handler_clone.clone();
             let btp = btp_clone.clone();
             store
                 .insert_account(account_details)
-                .map_err(|_| warp::reject::custom(ApiError::InternalServerError))
+                // TODO more understandable error?
+                // Internal Server Error doesn't tell so much
```
gakonst: Yeah good point, currently our store errors do not return something meaningful. We'll be able to resolve this once our futures return better errors instead of just `()`

crates/interledger-api/src/routes/node_settings.rs
```diff
@@ -174,3 +189,19 @@ where
         .or(put_static_route)
         .boxed()
 }
+
+impl ApiError {
+    fn create_invalid_account_id(invalid_account_id: Option<&str>) -> Self {
+        let detail = match invalid_account_id {
+            Some(invalid_account_id) => format!("{} is an invalid account ID", invalid_account_id),
+            None => "Invalid string was given as an account ID".to_owned(),
```
gakonst: If it was `None`, shouldn't the error be "No account id was provided"?

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,196 @@
...
+use warp::{reply::json, reply::Response, Reply};
+
+#[derive(Clone, Debug, Serialize)]
+pub struct ApiError {
+    #[serde(serialize_with = "serialize_type")]
+    pub r#type: ApiErrorType,
+    pub title: &'static str,
+    #[serde(serialize_with = "serialize_status_code")]
+    pub status: http::StatusCode,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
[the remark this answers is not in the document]
emschwartz: Don't think so, I think it will serialize it as `null` values for JSON

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,196 @@
...
+            ApiErrorType::Default,
+            "Internal Server Error",
+            None,
+            None,
+            None,
+        )
+    }
+
+    pub fn create_internal_server_error(
+        r#type: ApiErrorType,
```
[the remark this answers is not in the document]
dora-gt: Let me think. I can't answer at this point whether it is possible or not.
What I'm thinking about is realizing these:
1. Loosely coupling structure
2. Extendable
3. But useful

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+}
+
+// top level errors
+const ERROR_TYPE_PREFIX: &str = "https://errors.interledger.org/http-api";
+
+impl ApiError {
+    #[allow(dead_code)]
+    pub fn create_default_bad_request() -> Self {
```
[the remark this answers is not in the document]
dora-gt: I personally think I'm not a big fan of this format because it looks like a static variable.
**Programs** do something which means every program could be expressed as a verb.
That said, it actually returns just an instance (struct), this might be a special case that allows the method names to be a simple noun.

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,196 @@
...
+        // about:blank is a pre-defined value
+        ApiErrorType::Default => s.serialize_str("about:blank"),
+        ApiErrorType::Custom(custom_type) => {
+            s.serialize_str(&format!("{}/{}", ERROR_TYPE_PREFIX, custom_type))
+        }
+    }
+}
+
+// top level errors
+const ERROR_TYPE_PREFIX: &str = "https://errors.interledger.org/http-api";
```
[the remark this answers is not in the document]
dora-gt: Yes, I think we should provide some actual documents here so that developers easily solve the problem.

crates/interledger-api/src/routes/node_settings.rs
dora-gt: That might be nicer. Let me try.

crates/interledger-api/src/error.rs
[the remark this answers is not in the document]
dora-gt: Yes, I think option fields should be omitted when they are just `null`.

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,196 @@
...
+
+#[derive(Clone, Debug, Serialize)]
+pub struct ApiError {
+    #[serde(serialize_with = "serialize_type")]
+    pub r#type: ApiErrorType,
+    pub title: &'static str,
+    #[serde(serialize_with = "serialize_status_code")]
+    pub status: http::StatusCode,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub detail: Option<String>,
```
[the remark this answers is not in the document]
dora-gt: Unfortunately, it is defined as `detail` in [RFC7807](https://tools.ietf.org/html/rfc7807#section-3.1). I think we should comply with that.

crates/interledger-api/src/error.rs
dora-gt: I think it should not because we could provide a more clear type for it.

RFC7807 defines a default error type like `about:blank` which looks a bit weird but actually it is. This tells nothing for developers, so I think we should basically avoid using this. I provided `default_` methods just for ease of OUR matters.

`instance` means just an error instance, for example, we could provide an identifier of the error and developers query to us with the identifier, then we could investigate the error cause on our machines looking for with the identifier.

It is useful when we run interledger.rs on the production environment where we should not provide so much information for the error because that might mean leak of information for security matters.

`detail` always means the details of the error. We could fill the field with whatever we want. RFC7807 defines `detail` as `"detail" (string) - A human-readable explanation specific to this
 occurrence of the problem.`

crates/interledger-api/src/routes/node_settings.rs
dora-gt: hmmmm it is used here:
https://github.com/interledger-rs/interledger-rs/pull/324/files#diff-56b7912c05b5d66132ffce0b4e35f674L170
and this means, something was actually given, but it doesn't look like an account id. So... it doesn't mean `No account id was provided `.

crates/interledger-api/src/routes/node_settings.rs
dora-gt: I tried but it seemed impossible because the `Reject` trait was private (I believe you meant `Reject` trait instead of `Rejection` because it is struct).

Tracked here: https://github.com/seanmonstar/warp/issues/280

dora-gt: Nooooo I'm so sorry I just realized you meant a bit different thing.
You mean, make ApiError return `Custom` of `warp::Rejections`, right?
Maybe that works, let me try.

emschwartz: I don't think it should return a `warp::Rejection` but there should be an easy way to convert it. Maybe `impl From<ApiError> for warp::Rejection`?

dora-gt: I see, I'll try.

dora-gt: I tried to use `into`, and it worked in some cases for example:
https://github.com/interledger-rs/interledger-rs/pull/324/files#diff-b6ee23b4722a9bdbc19e7bc41b67c64cR75

however, in some cases I couldn't make it work because it says like:
```
error[E0283]: type annotations required: cannot resolve `_: warp::reject::sealed::CombineRejection<warp::reject::Rejection>`
  --> crates/interledger-api/src/routes/accounts.rs:89:10
   |
89 |         .and_then(|username: Username, store: S| {
   |          ^^^^^^^^
```
I tried to fix this but could not come up with how to. It seemed that I had to specify some type annotations but I could not understand what is required.

Could you please help me? @interledger-rs/contributors

dora-gt: I tried but I could not completely work it. It worked in some cases for example,
https://github.com/interledger-rs/interledger-rs/pull/324/files#diff-b6ee23b4722a9bdbc19e7bc41b67c64cR75

but it didn't work in some cases for example,
https://github.com/interledger-rs/interledger-rs/pull/324/files#diff-b6ee23b4722a9bdbc19e7bc41b67c64cR95

It says
```
error[E0283]: type annotations required: cannot resolve `_: warp::reject::sealed::CombineRejection<warp::reject::Rejection>`
  --> crates/interledger-api/src/routes/accounts.rs:89:10
   |
89 |         .and_then(|username: Username, store: S| {
   |
```

It seems that I have to specify some type parameters but I could not figure what was required. I pushed a branch where we can reproduce the compile error, so could you please take a look at and could you help me? @interledger-rs/contributors 

branch: `feature/dx/improve-http-responses-err`

dora-gt: Seems that this works:

image

Is this correct?

dora-gt: I think I could solve.

crates/interledger-api/src/error.rs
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,196 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub instance: Option<String>,
```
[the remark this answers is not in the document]
dora-gt: Added.

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,196 @@
...
+            "Internal Server Error",
+            None,
+            None,
+            None,
+        )
+    }
+
+    pub fn create_internal_server_error(
+        r#type: ApiErrorType,
+        title: &'static str,
```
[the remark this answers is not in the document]
dora-gt: I realized, in the first place, making this static was wrong because `title` should be filled depending on the case like `Username "xxx" was not found` which looks better for users. I would like to change this.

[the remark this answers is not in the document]
dora-gt: ^ ... just please forget this. `title` is defined as `It SHOULD NOT change from occurrence to occurrence of the problem, ` 😔

dora-gt:
I think I've fixed most points but I'm now trying to add tests.

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+}
+
+impl ApiError {
+    #[allow(dead_code)]
+    pub fn default_bad_request() -> Self {
+        ApiError::bad_request(&DEFAULT_BAD_REQUEST_TYPE, None, None, None)
+    }
+
+    pub fn bad_request(
```
emschwartz: Instead of defining two functions for every type of error, I think it might be better to have a single function called `bad_request` (in this case) that would do what `default_bad_request` currently does. `detail`, `instance`, and `option`, could be modified directly or, better yet, there could be methods on `ApiError` like `detail(&mut self, details: String)` that set the values for the other.

Calling this would look like:
```rust
ApiError::bad_request().detail("no good".to_string())
```

(Those methods could also take `self` instead of `&mut self`)

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,356 @@
...
+};
+const DEFAULT_UNAUTHORIZED_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Unauthorized",
+};
+const DEFAULT_NOT_FOUND_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Not Found",
+};
+const DEFAULT_METHOD_NOT_ALLOWED_TYPE: ApiErrorType = ApiErrorType {
```
emschwartz: Shouldn't these also have status codes, since each corresponds to a particular code?

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -134,10 +138,10 @@ where
     // POST /accounts
     let btp_clone = btp.clone();
     let outgoing_handler_clone = outgoing_handler.clone();
-    let post_accounts = warp::post2()
-        .and(accounts_index)
+    let post_accounts = accounts_index
+        .and(warp::post2())
```
emschwartz: It should be alright for the method to come before the path, no?

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -193,23 +200,26 @@ where
...
         .boxed();
 
     // GET /accounts/:username/balance
-    let get_account_balance = warp::get2()
-        .and(admin_or_authorized_account.clone())
+    let get_account_balance = admin_or_authorized_account
+        .clone()
         .and(warp::path("balance"))
         .and(warp::path::end())
+        .and(warp::get2())
```
emschwartz: This doesn't make sense. If the `admin_or_authorized_account` comes first, aren't we going to get a lot of incorrect errors saying the request is not authorized?

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -233,48 +243,51 @@ where
...
 
     // DELETE /accounts/:username
-    let delete_account = warp::delete2()
-        .and(admin_only.clone())
-        .and(account_username_to_id.clone())
+    let delete_account = account_username_to_id
+        .clone()
         .and(warp::path::end())
+        .and(warp::delete2())
+        .and(admin_only.clone())
```
emschwartz: Like above, I think the order should always be:
1. Method
2. Path
3. Auth
4. Body
5. Store

crates/interledger-api/src/error.rs
```diff
@@ -0,0 +1,356 @@
...
+};
+use warp::{reject::custom, reply::json, reply::Response, Rejection, Reply};
+
+/// This struct represents the fields defined in [RFC7807](https://tools.ietf.org/html/rfc7807).
+/// The meaning of each field could be found at [Members of a Problem Details Object](https://tools.ietf.org/html/rfc7807#section-3.1) section.
+/// ApiError implements Reply so that it could be used for responses.
+#[derive(Clone, Debug, Serialize)]
+pub(crate) struct ApiError {
+    /// `type` is a URI which represents an error type. The URI should provide human-readable
+    /// documents so that developers could solve the problem easily.
```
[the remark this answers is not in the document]
dora-gt: May I ask the reason? I'm just curious.
Does that sound like too **uncertain**?

crates/interledger-api/src/error.rs
dora-gt: I think it is correct and I feel like I made this too complicated. It seems that I could make this more simple 😔 Let me try.

crates/interledger-api/src/error.rs
dora-gt: Should I take `mut self` like Clap?
https://github.com/clap-rs/clap/blob/v2.31.2/src/app/mod.rs#L178-L181

I'm not sure which is more useful...

emschwartz: This may be somewhat helpful: https://doc.rust-lang.org/1.0.0/style/ownership/builders.html

**`&mut self`**
- makes it easier to conditionally call these methods:
```rust
let error = ApiError::bad_request();
if some_condition {
  error.detail("no good".to_string());
}
error
```
- makes it more cumbersome to return the error normally:
```rust
let error = ApiError::bad_request();
error.detail("no good".to_string());
return error;
```

**`self`**
- more cumbersome to call conditionally:
```rust
let error = ApiError::bad_request();
let error = if some_condition {
  error.detail("no good".to_string())
} else {
  error
};
```

- very natural to call normally:
```rust
return ApiError::bad_request().detail("no good".to_string());
```

In this case, I think I'd have a slight preference for taking `self`, unless you think we'll be calling it conditionally in most cases.

crates/interledger-api/src/routes/accounts.rs
dora-gt: I think placing method filters first looks a bit weird because it causes "method not allowed" errors many times. If we place path filters first, it causes one "method not allowed" error.

Isn't it more natural?

The priority of causes is:
I think path filters should be the first because it is on the lowest priority:
https://github.com/seanmonstar/warp/blob/master/src/reject.rs#L502-L516

> Not Found < Method Not Allowed < Status Code

crates/interledger-api/src/routes/accounts.rs
dora-gt: My bad. I'll fix.

crates/interledger-api/src/routes/accounts.rs
dora-gt: I don't understand why Method comes first. Could you explain why?

crates/interledger-api/src/routes/accounts.rs
emschwartz: I don't think that's right.. If you put the filters in the order I suggested and try POSTing to `/accounts` with an incorrect body, that will give a 400 error. According to those rules, the 400 error would take precedence over the 405

dora-gt: Maybe what I meant was a bit different, if I'm correct (actually I'm not so confident though...).

Imagine there are APIs like

- [GET] /foo/bar
- [POST] /abc
- [DELETE] /def

and if the order is "method -> path -> etc"
and I send a request of `[DELETE] /foo/bar`, it would cause

- Method Not Allowed for "[GET] /foo/bar"
- Method Not Allowed for "[POST] /abc"
- Not Found for "[DELETE] /def"

If the order is "path -> method -> etc", it would cause

- Method Not Allowed for "[GET] /foo/bar"
- Not Foud for "[POST] /abc"
- Not Foud for "[DELETE] /def"

I felt the latter was more natural.

crates/interledger-api/src/routes/accounts.rs
dora-gt: Noooo I'm so trapped into the **GENERICS HELL** of filters... I'm so sorry that I'm spending so long for this but it seems that I need some more time. I'm trying hard...

crates/interledger-api/src/routes/accounts.rs
dora-gt: But... actually, the final outputs would not differ even if we adopt "method -> path -> etc".
I'm ok with changing.

crates/interledger-api/src/routes/accounts.rs
emschwartz: Oh no! What's the issue?

crates/interledger-api/src/routes/accounts.rs
emschwartz: That's an interesting problem. If you call `DELETE` on a path that doesn't support that, it definitely seems like it should return 405 instead of 404.

crates/interledger-api/src/routes/accounts.rs
dora-gt: I'm trying to separate `path` filter and `auth` filter like
https://github.com/interledger-rs/interledger-rs/blob/feature/dx/improve-http-responses-refactor/crates/interledger-api/src/routes/accounts.rs#L212
or
https://github.com/interledger-rs/interledger-rs/blob/feature/dx/improve-http-responses-refactor/crates/interledger-api/src/routes/accounts.rs#L234
(these are just wip)

I don't come up with how to separate these filters and apply later with `admin_only` filter.

it would look like (if possible)

- method filter
- path filter (extracting `:username`)
- path end filter
- admin filter OR authorized user filter (using extracted username) <- this is difficult for me 😔
- store

dora-gt: I think I came up with a way, I'm trying.

dora-gt: Really really really^100 struggled but finally came up with this solution:
https://github.com/interledger-rs/interledger-rs/commit/119459cf64eefd7e7b475b943a7429296c23fddc

I'm not sure this is the most essential way but I'm fixing with this.

emschwartz: It might be worth implementing these changes in separate PRs. We could keep the filter order as is for now and integrate the error changes. Then, we could change the filter order later

crates/interledger-api/src/routes/accounts.rs
emschwartz: Actually you're right. From a style perspective, I thought that made more sense and seems more like other web frameworks. But that may not be the best way to do it with warp

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -66,19 +66,24 @@ where
...
         // success to the next filter, it just gets rid of it
         .untuple_one()
         .boxed();
     let with_store = warp::any().map(move || store.clone()).boxed();
+    let admin_auth_header = format!("Bearer {}", admin_api_token);
+    let with_admin_auth_header = warp::any().map(move || admin_auth_header.clone()).boxed();
     let with_incoming_handler = warp::any().map(move || incoming_handler.clone()).boxed();
+    // Note that the following `accounts*` and `account*` filters are based on the path of `accounts`.
+    // These filters should be applied before anything so that some other functions or filters
+    // are NOT kicked for any paths.
```
emschwartz: What do you mean by "kicked for any paths"?

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,57 +92,91 @@ where
...
-                .map_err(move |_| {
-                    error!(
-                        "Invalid authorization provided for user: {}",
-                        auth.username()
-                    );
-                    warp::reject::custom(ApiError::Unauthorized)
-                })
+        .and(with_admin_auth_header.clone())
+        .and_then(move |path_username: Username, auth_string: String, store: S, admin_auth_header: String| -> Result<_, Rejection> {
+            let account_id = match store.clone().get_account_id_from_username(&path_username).wait() {
```
emschwartz: We should never use `.wait` outside of tests -- Tokio doesn't like when you block the thread

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,57 +92,91 @@ where
...
-        .clone()
-        .and(account_username_to_id.clone())
-        .clone()
-        .or(authorized_account_from_path
-            .clone()
-            .map(|account: A| account.id()))
-        .unify()
+
+    // Receives Result<A::AccountId, ApiError> and rejects if Err if given.
+    let admin_or_authorized_user_only = |id: Result<A::AccountId, ApiError>| -> Result<_, Rejection> {
```
emschwartz: Huh? This function doesn't seem like it does anything

crates/interledger-api/src/error/error_types.rs
```diff
@@ -0,0 +1,63 @@
+// APIs should implement their own `ApiErrorType`s to provide more detailed information
+// about what were the problem, for example, `JSON_SYNTAX_TYPE` or `ACCOUNT_NOT_FOUND_TYPE`.
+
+use super::{ApiErrorType, ProblemType};
+
+// Default errors
+#[allow(dead_code)]
+pub const DEFAULT_BAD_REQUEST_TYPE: ApiErrorType = ApiErrorType {
```
emschwartz: I don't think these need to be public. We should just use the functions on the `ApiError` struct

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,57 +92,91 @@ where
...
+        .and(warp::header::<String>("authorization"))
+        .and(with_store.clone())
+        .and_then(move |path_username: Username, auth_string: String, store: S| -> Result<_, Rejection> {
+            let auth = match AuthToken::from_str(&auth_string) {
+                Ok(auth) => auth,
+                Err(_) => return Ok(Err(ApiError::bad_request().detail(Some("Auth token is invalid.".to_owned())))),
+            };
+            let authorized_account = match store.clone().get_account_from_http_auth(auth.username(), auth.password()).wait() {
+                Ok(authorized_account) => authorized_account,
+                Err(_) => return Ok(Err(ApiError::from_api_error_type(&ACCOUNT_NOT_FOUND_TYPE).not_found_account(Some(&path_username)))),
```
emschwartz: I think it's considered better practice to return the same response whether the auth is invalid or its for a different account

crates/interledger-api/tests/accounts.rs
```diff
@@ -0,0 +1,711 @@
...
+
+    let mut runtime = RuntimeBuilder::new().build().unwrap();
+
+    let redis_test_context = redis_test_helpers::TestContext::new();
+    let node_http_port = redis_test_helpers::get_open_port(Some(7770));
+    let node_settlement_port = redis_test_helpers::get_open_port(Some(3000));
+    let mut connection_info = redis_test_context.get_client_connection_info();
+    connection_info.db = 1;
+
+    let node = InterledgerNode {
```
emschwartz: Instead of running a full node, you can use warp's testing functionality to write these tests (and they can be unit instead of integration tests): https://docs.rs/warp/0.1.20/warp/test/index.html

crates/interledger-api/src/routes/accounts.rs
dora-gt: That is based on an idea that "We should ensure that the other filters never return `Rejection`s which are going to be prioritized than NotFound one.". Not Found is the safest error because it's least prioritized.

If we place other filters before path filters, it might return Unauthorized or Bad Request or something, which could possibly cause a confusing error that we sent a request to a specific path like `/foo/bar` and then the actual error was spawned by another path like `/abc`. This is really confusing when debugging.

crates/interledger-api/src/routes/accounts.rs
dora-gt: I'd really struggled to realize this without `.wait` but finally I could not make it.
Is there an easy way to do that? Because I'm really noob at `Future`s and `Filter`s, I don't come up with make this realize without `.wait`... 😞  anyway, I'll try.

crates/interledger-api/src/routes/accounts.rs
dora-gt: It's the main trick.

`is_admin_or_authorized_user_with_id` or `is_authorized_user_with_account` never return `Err`s or `Rejection`s. It wraps `Err` with `Ok`. `Ok(Err(xxx))`.

This is the way to handle errors without spawning `Rejection`s, and handle later when we SHOULD actually return it. Imagine that, we have an API endpoint of:

- `/accounts/:user/payments/list/page/:num` (just an example)

and we have filters like:

1. Auth filter based on `:user` path segment.
1. Path filter for remaining `payments/list/xxxx`
1. Body filter
1. etc

We should NOT return any `Rejection`s at the point of **1**. It is described here:
https://github.com/interledger-rs/interledger-rs/pull/324#discussion_r333806840
Filter **2** should be processed in advance of auth filters. So, after **2**, this function **unwraps** the error returned by **1**, then a `Rejection` is spawned.

dora-gt: But there may be a more smart way, I feel so.

crates/interledger-api/src/routes/accounts.rs
dora-gt: If the path is correct, it will return 405. So I think no problem.

crates/interledger-api/src/error.rs
dora-gt: Fixed.

crates/interledger-api/src/error.rs
dora-gt: Fixed.

crates/interledger-api/src/routes/accounts.rs
dora-gt: Anyway, fixed, placing methods first.

crates/interledger-api/src/routes/accounts.rs
dora-gt: I think I could without using `.wait`. Please look at later.

crates/interledger-api/src/routes/accounts.rs
dora-gt: I think either work. Fixed placing method filter first, for now.

crates/interledger-api/src/routes/accounts.rs
dora-gt: Fixed without using `wait`. That was really difficult... and it might contain some other weird point, please let me know then.

crates/interledger-api/src/error/error_types.rs
dora-gt: I'll try.

dora-gt: I fixed a little bit but I think we should not implement methods on the top-level module, which use these consts because it breaks the Open/Close principle. I really hesitated to place this file `error_types` but I just put this file so that developers just easily look at errors.

If we provide error methods in the top level module and if we add some more API services, we have to add methods on top level module everytime. It definitely breaks `Close` principle. So I beliveve we should not do that.

crates/interledger-api/src/routes/accounts.rs
dora-gt: Fixed.

crates/interledger-api/src/routes/accounts.rs
emschwartz: Got it. I would rewrite that sentence to "Path filters should be applied before others (such as method and authentication) to avoid triggering unexpected errors for requests that do not match this path."

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,57 +95,116 @@ where
...
+                        let auth = match AuthToken::from_str(&auth_string) {
+                            Ok(auth) => auth,
+                            Err(_) => return Either::A(ok(Err(ApiError::account_not_found()))),
+                        };
+                        Either::B(
+                            store
+                                .get_account_from_http_auth(auth.username(), auth.password())
+                                .then(move |authorized_account: Result<A, _>| {
+                                    if authorized_account.is_err() {
+                                        return ok(Err(ApiError::account_not_found()));
```
emschwartz: nit: you can return `Ok(Err...` here because `Result` implements `IntoFuture`

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,57 +95,116 @@ where
...
-                    Ok(authorized_account)
-                } else {
-                    Err(warp::reject::custom(ApiError::Unauthorized))
-                }
+            move |path_username: Username,
+                  auth_string: String,
+                  store: S,
+                  admin_auth_header: String| {
+                store
+                    .get_account_id_from_username(&path_username)
```
emschwartz: I just realized that this will make a call to the store even for requests that don't match the path. It shouldn't do that. The first filter should just extract the username from the path and pass it on. Once we've matched the full path, we can do the auth check and look up the details in the store. That way, we also don't need to do this kind of strange nested `Result`

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,57 +95,116 @@ where
...
+        match account {
+            Ok(account) => Ok(account),
+            Err(error) => Err(error.into()),
+        }
+    };
+
     // POST /accounts
     let btp_clone = btp.clone();
     let outgoing_handler_clone = outgoing_handler.clone();
     let post_accounts = warp::post2()
```
emschwartz: I know I said to split the changes into two PRs but now the `node_settings` routes have the path first and then the method. You might as well switch the order on all of them if we're going to do it for some of them

crates/interledger-api/tests/accounts.rs
emschwartz: Ok fair enough

crates/interledger-api/src/routes/accounts.rs
dora-gt: I see... I have to learn Futures more.

crates/interledger-api/src/routes/accounts.rs
dora-gt: I've tried that approach and I could not make it at the point but... I just realized that maybe there is a different way. Let me try.

I really feel so sorry that I'm taking time so long... and thank you so much for being patient. I'm improving my Rust skills...

crates/interledger-api/src/routes/accounts.rs
dora-gt: Fixed.

crates/interledger-api/src/routes/accounts.rs
dora-gt: I may be doing something wrong but it seemed not compatible in this case:

It says:
```
error[E0277]: the trait bound `std::result::Result<<A as interledger_service::Account>::AccountId, warp::reject::Rejection>: futures::future::Future` is not satisfied
   --> crates/interledger-api/src/routes/accounts.rs:122:64
    |
122 |             store.get_account_id_from_username(&path_username).then(
    |                                                                ^^^^ the trait `futures::future::Future` is not implemented for `std::result::Result<<A as interledger_service::Account>::AccountId, warp::reject::Rejection>`
    |
    = note: required because of the requirements on the impl of `futures::future::Future` for `futures::future::either::Either<std::result::Result<<A as interledger_service::Account>::AccountId, warp::reject::Rejection>, _>`
    = note: required because of the requirements on the impl of `futures::future::IntoFuture` for `futures::future::either::Either<std::result::Result<<A as interledger_service::Account>::AccountId, warp::reject::Rejection>, _>`
```

https://github.com/interledger-rs/interledger-rs/commit/11ba1b351fe6f04bf1f117d7f5a0a0fb9c8bc793

crates/interledger-api/src/routes/accounts.rs
dora-gt: Finally... I could. Fixed.

crates/interledger-api/src/routes/accounts.rs
dora-gt: Sorry just forgot, fixed.

crates/interledger-api/src/error.rs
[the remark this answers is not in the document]
dora-gt: Fixed anyway.

crates/interledger-api/src/routes/accounts.rs
dora-gt: Maybe... because `Either` doesn't accept `IntoFuture`?
I'm not sure but I leave as-is for now.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,57 +95,99 @@ where
...
                     error!("Error getting account id from username: {}", username);
-                    warp::reject::custom(ApiError::AccountNotFound)
+                    ApiError::account_not_found().into()
                 })
         })
         .boxed();
-    let valid_account_authorization = warp::header::<AuthToken>("authorization")
+
+    let prepare_for_admin_or_authorized_user_only = account_username
+        .and(warp::header::<String>("authorization"))
```
emschwartz: I don't think any of the filters aside from `account_username` need to come before the rest of the path segments. Couldn't we have it where you have:

1. account username
2. other path filters
3. path::end
4. admin_or_authorized_user_only

And admin_or_authorized_user would be:
```
warp::header::<String>("authorization")
.and(with_store.clone())
.and(with_admin_auth_header.clone())
.and_then(|path_username: Username, auth_string: String, store: S, admin_auth_header: String| { ... });
```

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,57 +95,99 @@ where
...
+                                } else {
+                                    err(ApiError::unauthorized().into())
+                                }
+                            }),
+                    )
+                },
+            )
+        };
+
+    let prepare_for_authorized_user_only = account_username
```
emschwartz: I don't think we need this filter, because the authorization and with_store can just be included in `authorized_user_only`

crates/interledger-api/src/routes/accounts.rs
dora-gt: I think your opinion is right but very difficult to implement because `and_then` of your code can not take `path_username: Username` argument. I think it is because the arguments are set by `warp::header` and `with_store` and `with_admin_auth_header`. It doesn't know where `Username` comes from.

There might be some way of setting very difficult type annotations, I could not come up with. 😞 Could you suggest some?

emschwartz: Hmm let me see

dora-gt: Ah, we might be able to in another way, I'll try.

dora-gt: What about this one?

emschwartz: It looks like [`ext`](https://docs.rs/warp/0.1.20/warp/filters/ext/index.html) is meant for this purpose but I'm having trouble getting it to work

dora-gt: I didn't notice that function. It worked. Thank you.

Taking a very glance at [Route](https://github.com/seanmonstar/warp/blob/master/src/route.rs), it seemed safe to use `ext` for this use because it is reset by a request.

crates/interledger-api/src/routes/accounts.rs
dora-gt: Fixed using `ext`.

dora-gt:
If `ext` looks good, I'll merge this.
Anything else? @gakonst

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,48 +95,103 @@ where
...
-                    error!(
-                        "Invalid authorization provided for user: {}",
-                        auth.username()
-                    );
-                    warp::reject::custom(ApiError::Unauthorized)
-                })
+
+    // This is required when using `admin_or_authorized_user_only` or `authorized_user_only` filter.
+    // Sets Username from path into ext for context.
+    let set_account_username_to_ext = account_username
```
emschwartz: nit: should we just replace `account_username` with this? It doesn't seem like it would cause any problems if we set the field and then don't always use it (though I think it is used in most cases for `account_username`)

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,48 +95,103 @@ where
...
+                        let auth = match AuthToken::from_str(&auth_string) {
+                            Ok(auth) => auth,
+                            Err(_) => return Either::A(err(ApiError::account_not_found().into())),
+                        };
+                        Either::B(
+                            store
+                                .get_account_from_http_auth(auth.username(), auth.password())
+                                .then(move |authorized_account: Result<A, _>| {
+                                    if authorized_account.is_err() {
+                                        return err(ApiError::account_not_found().into());
```
emschwartz:
```suggestion
                                        return err(ApiError::unauthorized().into());
```

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -87,48 +95,103 @@ where
...
+                Err(_) => {
+                    return Either::A(err::<A, Rejection>(ApiError::account_not_found().into()))
+                }
+            };
+            Either::B(
+                store
+                    .get_account_from_http_auth(auth.username(), auth.password())
+                    .then(move |authorized_account: Result<A, _>| {
+                        if authorized_account.is_err() {
+                            return err::<A, Rejection>(ApiError::account_not_found().into());
```
emschwartz:
```suggestion
                            return err::<A, Rejection>(ApiError::unauthorized().into());
```

crates/interledger-api/src/routes/accounts.rs
dora-gt: Do you mean like this?

```
    let account_username = accounts
        .and(warp::path::param2::<Username>())
        .and_then(|username: Username| {
            warp::filters::ext::set(username);
            Ok(username)
        })
        .boxed();
```

It would cause errors like this:
```
error[E0593]: closure is expected to take 3 arguments, but it takes 2 arguments
   --> crates/interledger-api/src/routes/accounts.rs:261:10
    |
261 |         .and_then(|id: A::AccountId, store: S| {
    |          ^^^^^^^^ ---------------------------- takes 2 arguments
    |          |
    |          expected closure that takes 3 arguments
    |
    = note: required because of the requirements on the impl of `warp::generic::Func<(interledger_service::auth::username::Username, <A as interledger_service::Account>::AccountId, S)>` for `[closure@crates/interledger-api/src/routes/accounts.rs:261:19: 267:10]`
```

because the `username` parameter is not consumed. If we don't place `Ok(username)`,  we would have to change `account_username_to_id` to take `username` from `ext`, that looks a bit abnormal, I think.

dora-gt: Or ... `account_username_to_id` could use `warp::path::param2::<Username>()` for its self.

dora-gt: Fixed.

dora-gt:
Moved `error` module into `interledger-http` because we needed to return appropriate errors also in ILP over HTTP API.

Also it seemed that the HTTP APIs depend on `interledger-http` because `interledger-api` crate **uses** `interledger-http` crate. So I thought moving the error module into `interledger-http` is reasonable.

In addition, I made it `pub` because @gakonst suggested that he wanted to use the functions and structs from settlement engines.

dora-gt:
#415 would be resolved too.

gakonst:
One thing that could be nice would also be to add `Deserialize` support to the `ApiError` type. Due to using the `&'static str` references it might be hard to implement a Deserializer though.

crates/interledger-http/src/error/error_types.rs
```diff
@@ -0,0 +1,75 @@
...
+};
+pub const DEFAULT_IDEMPOTENT_CONFLICT_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Provided idempotency key is tied to other input",
+    status: http::StatusCode::CONFLICT,
+};
+
+// ILP over HTTP specific errors
+pub const INVALID_ILP_PACKET_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("ilp-over-http/invalid-packet"),
```
gakonst: This assumes that at some poin we will put all our errors in `https://errors.interledger.org/http-api`, which is fine, but we might want to add a TODO somewhere that it is currently not live

crates/interledger-http/src/error/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+            None => "Invalid string was given as an account ID".to_owned(),
+        });
+        ApiError::from_api_error_type(&INVALID_ACCOUNT_ID_TYPE).detail(detail)
+    }
+
+    pub fn invalid_ilp_packet() -> Self {
+        ApiError::from_api_error_type(&INVALID_ILP_PACKET_TYPE)
+    }
+
+    pub fn detail<T>(mut self, detail: Option<T>) -> Self
```
gakonst: Is there any reason for this to take an `Option`? I'd expect that if we're calling the `detail` method we already want to store something. Doubt we'd want to call `detail(None)` in order to reset the details of an error. I'd suggest making this `.detail(mut self, detail: T)`

crates/interledger-http/src/error/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+    where
+        T: Into<String>,
+    {
+        self.detail = detail.map(|detail| detail.into());
+        self
+    }
+
+    #[allow(dead_code)]
+    pub fn instance<T>(mut self, instance: Option<T>) -> Self
```
gakonst: Same comment as with `detail`, when doing this pattern I do not think the params should be `Option`

crates/interledger-http/src/error/error_types.rs
dora-gt: I assume that it is going to be published. I'll put a comment on that.

crates/interledger-http/src/error/mod.rs
dora-gt: The reason was that I thought setting always String was not essential because the struct field is `Option<String>`. If it is very common practice in Rust, I'm OK with changing it so that it accepts just `String`.

crates/interledger-http/src/error/error_types.rs
dora-gt: Added.

crates/interledger-http/src/error/mod.rs
dora-gt: A person in Japanese Rust slack suggested that there is a similar use-case:
http://gtk-rs.org/docs/src/gtk/auto/window.rs.html#259-325
http://gtk-rs.org/docs/src/gtk/auto/window.rs.html#597
This seem to accept pure types and wraps it with `Some`. Also I assume that we might not want to reset it with `None`, I'll change this.

dora-gt: Fixed.

crates/interledger-http/src/error/mod.rs
dora-gt: Fixed.

dora-gt:
What would you want to use Deserializer for?
If it is not required now, it could be when it comes necessary. I'd be happy to change then.

gakonst:
Alright this looks great to me. @emschwartz if you're good with the latest changes I think we should go ahead and merge it.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -66,69 +65,129 @@ where
...
-                Err(warp::reject::custom(ApiError::Unauthorized))
+                Err(ApiError::unauthorized().into())
             }
         })
         // This call makes it so we do not pass on a () value on
         // success to the next filter, it just gets rid of it
         .untuple_one()
         .boxed();
     let with_store = warp::any().map(move || store.clone()).boxed();
+    let admin_auth_header = format!("Bearer {}", admin_api_token);
```
emschwartz: This can be done in a separate PR but we should make this a `secrecy::SecretString`

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -194,22 +254,25 @@ where
...
-        .and(admin_or_authorized_account.clone())
+        .and(account_username.clone())
         .and(warp::path::end())
+        .and(admin_or_authorized_user_only.clone())
         .and(with_store.clone())
         .and_then(|id: A::AccountId, store: S| {
             store
                 .get_accounts(vec![id])
-                .map_err(|_| warp::reject::not_found())
+                // TODO return username
```
emschwartz: Why? The caller knows it because it was included in their request

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -234,47 +297,50 @@ where
...
                     error!("Error deleting account {}", id);
-                    warp::reject::custom(ApiError::InternalServerError)
+                    ApiError::internal_server_error().into()
                 })
                 .and_then(|account| Ok(warp::reply::json(&account)))
         })
         .boxed();
 
     // PUT /accounts/:username/settings
     let put_account_settings = warp::put2()
```
emschwartz: Some of the methods come before the path and some come after. Can you make them consistent? If you think it's not related to this PR, it can be in another one but it seems odd that it's inconsistent

crates/interledger-http/src/error/error_types.rs
```diff
@@ -0,0 +1,75 @@
...
+    status: http::StatusCode::BAD_REQUEST,
+};
+pub const JSON_DATA_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("json-data"),
+    title: "JSON Data Error",
+    status: http::StatusCode::BAD_REQUEST,
+};
+pub const UNKNOWN_JSON_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::InterledgerHttpApi("json-unknown"),
+    title: "Unknown JSON Error",
```
emschwartz: It doesn't seem like this is produced anywhere

Also, when I read the name first I was confused about whether "unknown" meant the type of the JSON error was unknown or the type of the JSON data. If it's not being used we should take this out. If there is a reason to keep it I would rename it to something like unexpected or invalid JSON

crates/interledger-http/src/lib.rs
```diff
@@ -30,3 +37,19 @@ pub trait HttpStore: Clone + Send + Sync + 'static {
         token: &str,
     ) -> Box<dyn Future<Item = Self::Account, Error = ()> + Send>;
 }
+
+pub fn deserialize_json<T: DeserializeOwned + Send>(
```
emschwartz: It seems kind of crazy that warp or serde don't have something like this built in

crates/interledger-http/src/error/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+
+/// API error type prefix of problems.
+/// This URL prefix is currently not published but we assume that in the future.
+const ERROR_TYPE_PREFIX: &str = "https://errors.interledger.org/http-api";
+
+/// This struct represents the fields defined in [RFC7807](https://tools.ietf.org/html/rfc7807).
+/// The meaning of each field could be found at [Members of a Problem Details Object](https://tools.ietf.org/html/rfc7807#section-3.1) section.
+/// ApiError implements Reply so that it could be used for responses.
+#[derive(Clone, Debug, Serialize)]
+pub struct ApiError {
```
emschwartz: I stumbled across https://github.com/chridou/http-api-problem while looking into why `http::StatusCode` doesn't implement `Deserialize` (https://github.com/hyperium/http/issues/273). Did you see this before? I could be convinced it's not worth adding another dependency but it seems it might be worth just using this crate...

crates/interledger-api/Cargo.toml
```diff
@@ -26,11 +26,23 @@ interledger-btp = { path = "../interledger-btp", version = "^0.2.2-alpha.1", def
...
+[dev-dependencies]
+ilp-node = { path = "../ilp-node", version = "^0.4.1-beta.2"}
+interledger = { path = "../interledger", version = "^0.4.1-alpha.1" }
+env_logger = "0.6.1"
+net2 = "0.2.33"
+rand = "0.6.5"
+redis = "0.13.0"
+tokio = "0.1.20"
+base64 = "0.10.1"
+approx = "0.3.2"
```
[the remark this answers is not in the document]
dora-gt: It might be too much but it is used to compare float values.
https://github.com/interledger-rs/interledger-rs/pull/324/files#diff-6e43ab15fc08afa75c9e9c6bef277b51R118

crates/interledger-api/src/lib.rs
```diff
@@ -289,38 +262,17 @@ where
...
+        routes::accounts_api(
+            self.server_secret,
+            self.admin_api_token.clone(),
+            self.default_spsp_account,
+            self.incoming_handler,
+            self.outgoing_handler,
+            self.btp,
+            self.store.clone(),
+        )
+        .or(routes::node_settings_api(self.admin_api_token, self.store))
```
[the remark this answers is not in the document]
dora-gt: Because it is handled at more near the root of filters.
https://github.com/interledger-rs/interledger-rs/pull/324/files#diff-a5480f9daf5ea636e9af7e5b0c00a617R308

I just added a rough test to check if it is working.
https://github.com/interledger-rs/interledger-rs/pull/324/files#diff-6e43ab15fc08afa75c9e9c6bef277b51R325
It rejects with RFC7807.

crates/interledger-api/src/routes/accounts.rs
dora-gt: Why don't we create a project (or an issue) that investigates security matters?
There might be other vulnerabilities.

crates/interledger-api/src/lib.rs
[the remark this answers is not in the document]
dora-gt: I just confirmed again just in case that ILP over HTTP returns RFC7807.

```
$ echo test | curl \
>         -X POST \
>         -H "Authorization: Bearer dora:dora" \
>         -H "Content-Type: application/json" \
>         -d @- \
>         http://localhost:7770/ilp 2>/dev/null | jq
{
  "type": "https://errors.interledger.org/http-api/ilp-over-http/invalid-packet",
  "title": "Invalid Packet",
  "status": 400,
  "datetime": "2019-10-19T16:04:26.511487+09:00"
}
```

crates/interledger-api/src/routes/accounts.rs
dora-gt: ... yes, right... I'm not sure why I thought so😨

Maybe misunderstood like the method is querying user by `id`.

dora-gt: Removed.

crates/interledger-api/src/routes/accounts.rs
dora-gt: Which API(s) are you referring to?
I think I've fixed all of the APIs. (method -> path)

crates/interledger-http/src/error/error_types.rs
dora-gt: It is used here:
https://github.com/interledger-rs/interledger-rs/pull/324/files#diff-e4f8daa4352bdb662146ead816fd6635R289

It could be renamed to something related to IO. I'll fix.
https://github.com/serde-rs/json/blob/master/src/error.rs#L112-L130

crates/interledger-http/src/lib.rs
dora-gt: I think they should even support RFC7807.

crates/interledger-http/src/error/mod.rs
dora-gt: Honestly, I didn't notice that implementation. If you really want to move to this crate, I'll change.

Hmmm but the RFC would not likely be changed so often, we could just leave it as-is, which means that using the crate is not so meaningful in the future. Idk, that's just my opinion. I respect your decision and sorry that I didn't notice the implementation.

dora-gt: I'm not sure they support `warp` for that.

crates/interledger-http/src/error/error_types.rs
dora-gt: Fixed.

crates/interledger-http/src/error/mod.rs
dora-gt: So if you want to move to this crate, we could consider:

1. Merge this PR to improve DX at this point.
1. Make a PR to `http-api-problem` to support `warp`.
1. Then refactor this to use `http-api-problem`

though I'm not sure this is worth or not.

gakonst: We could also @dora-gt's implementation to a separate crate and publish it (with the Warp features as a `feature`). More flexible as we'd control it and can customize things as we like.

emschwartz: The main downsides of that crate that I see are:
- Doesn't support converting to an `http::Response` (`warp` implements `Reply` for that [here](https://github.com/seanmonstar/warp/blob/master/src/reply.rs#L387))
- Doesn't support the nice JSON parsing errors implemented here

I would be supportive of merging this now and potentially refactoring it out into a separate crate as another PR if we want to use it in the `settlement-engines` repo. We can either create a new repo for it under `interledger-rs` or we could see if the author of `http-api-problem` would accept a PR for the additional functionality we want from it.

emschwartz: Sounds good. 

> Ping the author if he/she is interested in support warp (or some implementations which use http::Response).

Just to be really clear on the ask, I'm pretty sure that they would not need to support `warp`, but just converting to `http::Response`, which is much more widely used.

dora-gt:
I'm merging this.

Issue #313, opened 2019-09-20
Use Secret types in the API structs


emschwartz:
Such as `AccountSettings` and `AccountDetails`.

Note that we should also separate the `btp_outgoing_token` from the `btp_uri` field in `AccountDetails`.

Issue #286, opened 2019-09-06
Should we make the API field names more consistent with the node configuration?


emschwartz:
We are making a number of breaking changes in https://github.com/interledger-rs/interledger-rs/pull/236 to make the naming of config parameters more consistent. Is it worth making a bunch of breaking changes to the API to make the field names more consistent as well? For example, `http_endpoint` could be `http_server_url` and `btp_uri` could be `btp_server_url`

gakonst:
I'm fine with that. Let's also use that PR as a chance to split out the btp_uri into `btp_outgoing_token` and `btp_server_url` (to avoid having to prefix the URI with credentials)

bstrie:
I'd say it's now or never. 👍

PR #328, opened 2019-09-24
Separate secrets from URLs and make names consistent, use Secrets for structs

dora-gt:
Fixes: #286
Fixes: #313

**This is WIP** to confirm the basic concept. I haven't checked whether it works or not.

Now that we are separating BTP outgoing tokens from its URLs, it seemed that separating HTTP tokens from its URLs was natural.

Also, giving these explicit parameters from CLI looked easier to understand.
https://interledger.slack.com/archives/CHC51E54J/p1567729828054600
An ambiguous format may confuse people.

- Separate secrets from URLs and make names consistent
    - `btp_uri ` -> `btp_server_url` + `btp_outgoing_token`
    - `http_endpoint` -> `http_server_url` + (`http_outgoing_bearer_token` or (`http_outgoing_basic_auth_username ` + `http_outgoing_basic_auth_password`))
    - `http_incoming_token ` -> `http_incoming_bearer_token` or (`http_incoming_basic_auth_username ` + `http_incoming_basic_auth_password`)


crates/interledger-store-redis/src/account.rs
```diff
@@ -147,12 +127,12 @@ impl Account {
             asset_scale: details.asset_scale,
             max_packet_amount: details.max_packet_amount,
             min_balance: details.min_balance,
-            http_endpoint,
-            http_incoming_token,
-            http_outgoing_token,
-            btp_uri,
-            btp_incoming_token,
-            btp_outgoing_token,
+            http_endpoint: http_server_url,
```
emschwartz: It might be a little confusing that this is supposed to point to the `/ilp` endpoint rather than just the base URL of the http server

crates/interledger/src/main.rs
```diff
@@ -446,3 +456,67 @@ struct NodeAccountsAddOpt {
     packets_per_minute_limit: Option<u32>,
     amount_per_minute_limit: Option<u64>,
 }
+
+impl NodeAccountsAddOpt {
```
emschwartz: What is this needed for?

gakonst:
I think we should break this PR to split the `btp_uri` in `btp_endpoint` and `btp_outgoing_token`, and then convert all the tokens to `Secret` types in a separate PR.

I believe @dora-gt used `SecretBytes` instead of `SecretString` because `SecretString` does not implement any of the needed serde methods. @emschwarz

crates/interledger-api/src/lib.rs
```diff
@@ -106,12 +107,16 @@ pub struct AccountDetails {
...
     pub min_balance: Option<i64>,
-    pub http_endpoint: Option<String>,
-    // TODO use Secrets here
-    pub http_incoming_token: Option<String>,
-    pub http_outgoing_token: Option<String>,
-    pub btp_uri: Option<String>,
-    pub btp_incoming_token: Option<String>,
+    pub http_server_url: Option<String>,
+    #[serde(serialize_with = "optional_secret_bytes_to_utf8")]
+    pub http_incoming_token: Option<SecretBytes>,
```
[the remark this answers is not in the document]
dora-gt: It's just because:
- [Account uses `SecretBytes`](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-store-redis/src/account.rs#L46-L54)
- [Easier to use here](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-store-redis/src/account.rs#L142-L162)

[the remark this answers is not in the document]
dora-gt: I think it could be `SecretString` if I write some conversion code.

crates/interledger-api/src/lib.rs
```diff
@@ -121,6 +126,16 @@ pub struct AccountDetails {
...
 }
 
+fn optional_secret_bytes_to_utf8<S>(
+    _bytes: &Option<SecretBytes>,
+    serializer: S,
+) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+{
+    serializer.serialize_str("SECRET")
```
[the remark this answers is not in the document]
dora-gt: Oops... I thought there might be some reasons looking at `Account`. Thank you.

emschwartz:
I think it should work automatically if you use Option<SecretString>
because that implements Deserialize

>

crates/interledger-store-redis/src/account.rs
dora-gt: Do you mean we should specify these?
Sorry if I'm misunderstanding.

- `http_server_url` e.g. `http://some-host:7770`
- `ilp_over_http_path` e.g. `/ilp`
- `btp_path` e.g. `/ilp/btp`

This looks also a bit weird because the schemes don't match.
- `http`
- `btp+ws`

crates/interledger/src/main.rs
dora-gt: To encapsulate the logic of generating `http_incoming_token` and `http_outgoing_token`.
I think these should be encapsulated in the struct even if these are not reused somewhere. It could be.

dora-gt:
Changes on `Secret`izing variables could be another PR but I think CLI changes should be done here.

If we split `btp_uri` into `btp_server_url` and `btp_outgoing_token` then HTTP URLs should be as well. Then I think not only HTTP APIs but also CLI arguments should be modified.

crates/interledger-api/src/lib.rs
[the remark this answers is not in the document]
dora-gt: Fixed to use `SecretString`.

crates/interledger-api/src/lib.rs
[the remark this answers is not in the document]
dora-gt: Removed.

dora-gt:
Fixed some points though this PR contains multiple fixes.

- Separate secrets from URLs
- Make parameter names consistent
- Use `Secrets` for structs
- Make CLI arguments clear (`Bearer` token or `Basic` token)

crates/interledger/src/main.rs
gakonst: I'm fine with this approach

gakonst:
This looks good to me, is it still WIP?

crates/interledger-store-redis/src/account.rs
emschwartz: What do you think about `ilp_over_http_url`, `ilp_over_btp_url` (and then it would be `ilp_over_http_outgoing_token`, etc)?

"server_url" sounds more like it would be the base URL for something rather than a host and specific path to an endpoint

crates/interledger/Cargo.toml
```diff
@@ -63,6 +63,7 @@ tokio = "0.1.20"
 url = "2.1.0"
 libc = "0.2.62"
 warp = { version = "0.1.20", optional = true }
+secrecy = { version = "0.4.0", features = ["serde", "bytes"] }
```
emschwartz: Is bytes being used?

crates/interledger-store-redis/src/account.rs
dora-gt: I think it is nicer. Give me time to fix.

dora-gt:
Btw: I think we should avoid using `json!` macro because this is fragile against refactoring.

emschwartz:
> Btw: I think we should avoid using json! macro because this is fragile against refactoring.

What do you mean by that? Where are we using it that you think we shouldn't?

crates/interledger/Cargo.toml
dora-gt: No, I forgot to remove. Sorry. Fixed.

dora-gt:
Some tests use it like: ./crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs

crates/interledger-store-redis/src/account.rs
```diff
@@ -96,38 +96,18 @@ impl Account {
                 })?,
         };
 
-        let http_endpoint = if let Some(ref url) = details.http_endpoint {
+        let http_server_url = if let Some(ref url) = details.ilp_over_http_url {
```
dora-gt: oops... this looks a bit strange. I'll fix.

crates/interledger/src/main.rs
```diff
@@ -144,22 +145,46 @@ pub fn main() {
...
+                                .help("Basic auth username this account will use to authenticate HTTP requests sent to this server"),
+                            Arg::with_name("ilp_over_http_incoming_basic_auth_password")
+                                .long("ilp_over_http_incoming_basic_auth_password")
+                                .takes_value(true)
+                                .help("Basic auth password this account will use to authenticate HTTP requests sent to this server"),
+                            Arg::with_name("ilp_over_http_outgoing_bearer_token")
+                                .long("ilp_over_http_outgoing_bearer_token")
+                                .takes_value(true)
+                                .help("Bearer token this account will use to authenticate HTTP requests sent to their server"),
+                            Arg::with_name("ilp_over_http_outgoing_basic_auth_username")
```
emschwartz: nit: these could be shortened to just `ilp_over_http_outgoing_username` (username implies basic auth). Similarly the token one can just be `ilp_over_http_outgoing_token`

crates/interledger-store-redis/src/account.rs
```diff
@@ -147,12 +128,20 @@ impl Account {
             asset_scale: details.asset_scale,
             max_packet_amount: details.max_packet_amount,
             min_balance: details.min_balance,
-            http_endpoint,
-            http_incoming_token,
-            http_outgoing_token,
-            btp_uri,
-            btp_incoming_token,
-            btp_outgoing_token,
+            http_endpoint: ilp_over_http_url,
```
emschwartz: These field names should be updated to match the `Accountdetails`. It's clearer if the same field name is used throughout the code

crates/interledger/src/main.rs
dora-gt: I understand but I would like the current format because it is explicit. I think we should avoid ambiguous names.

My opinion:  **short  hard-to-understand < long understandable**

crates/interledger-store-redis/src/account.rs
```diff
@@ -147,12 +128,20 @@ impl Account {
             asset_scale: details.asset_scale,
             max_packet_amount: details.max_packet_amount,
             min_balance: details.min_balance,
-            http_endpoint,
-            http_incoming_token,
-            http_outgoing_token,
-            btp_uri,
-            btp_incoming_token,
-            btp_outgoing_token,
+            http_endpoint: http_server_url,
```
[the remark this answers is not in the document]
emschwartz: We can change what's stored in Redis too

crates/interledger/src/main.rs
emschwartz: I generally agree that longer, more explicit names are better. 7 words is a bit much for my taste though...

dora-gt: Hmmm it might be... Ok let's change.

crates/interledger-store-redis/src/account.rs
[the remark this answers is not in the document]
dora-gt: Fixed.

crates/interledger/src/main.rs
dora-gt: Fixed.

crates/interledger-store-redis/src/account.rs
dora-gt: Fixed.

dora-gt:
Changed across the board.
I think it got far easy-to-understand.

crates/interledger/src/main.rs
```diff
@@ -144,22 +145,46 @@ pub fn main() {
...
-                                .long("http_incoming_token")
+                            Arg::with_name("ilp_over_http_incoming_bearer_token")
+                                .long("ilp_over_http_incoming_bearer_token")
                                 .takes_value(true)
                                 .help("Bearer token this account will use to authenticate HTTP requests sent to this server"),
+                            Arg::with_name("ilp_over_http_incoming_username")
+                                .long("ilp_over_http_incoming_username")
+                                .takes_value(true)
+                                .help("Basic auth username this account will use to authenticate HTTP requests sent to this server"),
+                            Arg::with_name("ilp_over_http_incoming_token")
```
emschwartz: We should add an `ArgGroup` to make sure that the username and token are supplied together. Also, what do you think about calling this `_password`? Right now it's a little hard to tell the difference between this and the `bearer_token`

dora-gt: I think this is what we need:
https://github.com/interledger-rs/interledger-rs/pull/328/commits/fc9085aee81fc546452bf438a7ac3b20115b75b0

- `username` requires `password`
- `bearer_token` cannot be specified with basic auth (`username`)

dora-gt: Sorry that was not perfect. I fixed like: https://github.com/interledger-rs/interledger-rs/pull/328/commits/b08d7a8f308b362319e8fb96a0e8d971285132f5

This is better.

PR #329, opened 2019-09-25
Make ilp_address optional in the node configuration

emschwartz:
As discussed in https://github.com/interledger-rs/interledger-rs/pull/276#discussion_r327212567


crates/interledger/src/node.rs
```diff
@@ -90,8 +98,9 @@ where
 #[derive(Deserialize, Clone)]
 pub struct InterledgerNode {
     /// ILP address of the node
-    #[serde(deserialize_with = "deserialize_string_to_address")]
-    pub ilp_address: Address,
+    #[serde(deserialize_with = "deserialize_optional_address")]
+    #[serde(default)]
```
emschwartz: @gakonst this is the way to ensure serde doesn't throw a `missing field` error. Took me a little while to remember this attribute

PR #334, opened 2019-09-25
feat: ilp-cli tool for interacting with nodes

bstrie:
This PR adds a new command-line tool, `ilp-cli`. It's a beginner-friendly interface for interacting with a running interledger node. The intent is for `ilp-cli` to be for `interledger` what `redis-cli` is for `redis-server`. Right now it only includes as much functionality as is necessary (creating new accounts, posting payments, and getting account balances) to port the simple example from `curl` to `ilp-cli`. Supporting the remainder of the HTTP API will arrive in future PRs.


crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+        .version(crate_version!())
+        // TODO remove this line once this issue is solved:
+        // https://github.com/clap-rs/clap/issues/1536
+        .after_help("")
+        .args(&[
+            Arg::with_name("authorization_key")
+                .long("auth")
+                .env("ILP_CLI_AUTH")
+                .required(true)
+                .help("An authorization key granting access to the designated operation"),
```
emschwartz: This should specify whether this must be for bearer token-based auth

Cargo.toml
```diff
@@ -16,6 +16,7 @@ members = [
   "./crates/interledger-store-redis",
   "./crates/interledger-stream",
   "./crates/interledger-settlement-engines",
+  "./crates/ilp-cli",
```
emschwartz: Not sure about this being called `ilp-cli` rather than `interledger-cli`

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+            Arg::with_name("node_address")
```
emschwartz: In other places we call the URL a client is going to connect to `_url` and the address + port a service is going to bind to `_bind_address`
```suggestion
            Arg::with_name("node_url")
```

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                .long("node")
+                .env("ILP_CLI_NODE")
+                .default_value("localhost:7770")
+                .help("The URL of the node to which to connect"),
```
emschwartz:
```suggestion
                .help("The base URL of the node to connect to"),
```

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+            Arg::with_name("print_response")
```
emschwartz: As opposed to what?

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                .short("p")
+                .long("print-response")
```
emschwartz: I think we have a somewhat inconsistent use of underscores and dashes in our CLIs. Any reason not to use underscores? Mostly thinking about consistency with field names, though I could be convinced otherwise

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                .help("Upon a successful HTTP response, response body will be printed to stdout"),
+        ])
+        .subcommands(vec![
+            // Example: ilp-cli add-account alice ABC 9
+            SubCommand::with_name("add-account")
```
emschwartz: I think we might want to make this more like `ilp-cli accounts add...` because there may be other account-related operations we want to do. What do you think?

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
+use clap::{crate_version, App, Arg, ArgMatches, SubCommand};
+use reqwest;
+use std::{collections::HashMap, process::exit};
+
+pub fn main() {
+    // Define the arguments to the CLI
+    let mut app = App::new("ilp-cli")
```
emschwartz: Wonder if this should just be `ilp` when you call it

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                        .takes_value(true)
+                        .required(true)
+                        .help("The code of the asset associated with this account"),
+                    Arg::with_name("asset_scale")
+                        .index(3)
+                        .takes_value(true)
+                        .required(true)
+                        .help("The scale of the asset associated with this account"),
+                    // Optional, named arguments
+                    Arg::with_name("ilp_address").long("ilp-address").takes_value(true),
```
emschwartz: Each of these should have descriptions about what they are

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+            // Example: ilp-cli get-balance alice
+            SubCommand::with_name("get-balance")
+                .about("Returns the balance of an account")
+                .arg(Arg::with_name("account_username")
+                    .index(1)
+                    .takes_value(true)
+                    .required(true)
+                    .help("The username of the account whose balance to return")),
+            // Example: ilp-cli post-payment alice 500 "http://localhost:8770/accounts/bob/spsp"
+            SubCommand::with_name("post-payment")
```
emschwartz: nit:
```suggestion
            SubCommand::with_name("send-payment")
```

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                .about("Issue a payment from an account on this node")
```
emschwartz:
```suggestion
                .about("Send a payment from an account on this node")
```

crates/ilp-cli/src/main.rs
emschwartz: This won't actually work with the admin authorization right now. We should either make it so you can send a payment from a user's account using admin auth or make a note that it will only work for the authenticated user

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                .args(&[
+                    Arg::with_name("sender_username")
+                        .index(1)
```
emschwartz: I think positional arguments are slightly worse for something like this because it's hard to remember the order. I think it might be better to stick with named args and just provide the long and short versions of them

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                        .index(2)
+                        .takes_value(true)
+                        .required(true)
+                        .help("The amount to transfer from the sender to the receiver, denominated in units of the sender's assets"),
+                    Arg::with_name("receiver")
+                        .index(3)
+                        .takes_value(true)
+                        .required(true)
+                        // TODO: better way of describing this parameter
+                        .help("The SPSP address of the account receiving the payment"),
```
emschwartz:
```suggestion
                        .help("The Payment Pointer or SPSP address of the account receiving the payment (either of the form $bob.wallet.example or http://localhost:7770/accounts/just_an_example/spsp"),
```

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+    match matches.subcommand() {
+        // Execute the specified subcommand
+        (subcommand_name, Some(subcommand_matches)) => {
+            // Send HTTP request
+            let client = reqwest::Client::new();
+            let response = match subcommand_name {
+                "add-account" => {
+                    let args = extract_args(subcommand_matches);
+                    client
+                        // TODO: tacking on the protocol like this doesn't feel ideal,
```
emschwartz: By protocol do you mean `http://` or `/accounts`? The node URL, if specified should include either `http://` or `https://` so we don't have to guess which one it is

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                }
+                Ok(mut res) => match res.text() {
+                    Err(e) => {
+                        eprintln!("ILP CLI error: failed to parse response: {}", e);
+                        exit(1);
+                    }
+                    // Final output
+                    Ok(val) => {
+                        if res.status().is_success() {
+                            if matches.is_present("print_response") {
```
emschwartz: Instead of `print_response`, I think it's more common to have an optional `quiet` mode that's off by default

crates/interledger-api/src/lib.rs
```diff
@@ -21,6 +22,48 @@ use secrecy::SecretString;
...
+        NumOrStr::Num(n) => Ok(Some(n)),
+        NumOrStr::Str(s) => T::from_str(&s)
+            .map_err(de::Error::custom)
+            .and_then(|n| Ok(Some(n))),
+    }
+}
+
+// Ordinarily providing a serde default for Option types is unnecessary;
+// this is only for cases where we also set the `deserialize_with` attribute.
+pub fn optional_default<T>() -> Option<T> {
```
emschwartz: I believe just using `#[serde(default)]` will have the same effect

examples/simple/README.md
```diff
@@ -214,13 +214,13 @@ Let's create accounts on both nodes. The following script sets up accounts for t
...
 <!--!
-printf "Creating accounts:\n\n"
+printf "\nCreating accounts...\n\n"
 
 if [ "$USE_DOCKER" -eq 1 ]; then
     printf "Alice's account:\n"
     curl \
         -H "Content-Type: application/json" \
-        -H "Authorization: Bearer admin-a" \
+        -H "Authorization: Bearer admin-example" \
```
emschwartz: Why is this being changed? This is meant to show it using the admin credentials for different nodes

examples/simple/README.md
```diff
@@ -285,109 +285,73 @@ else
 ```bash
 # Insert accounts on Node A
 # One account represents Alice and the other represents Node B's account with Node A
+export ILP_CLI_AUTH=admin-example
```
emschwartz: I think we can either re-export this before specific calls or pass this in as a CLI arg in each call so that we don't need to make all the nodes use the same admin auth token

examples/simple/README.md
```diff
@@ -285,109 +285,73 @@ else
...
-    "asset_scale": 9,
-    "max_packet_amount": 100,
-    "ilp_over_http_incoming_token": "node_b-password",
-    "ilp_over_http_outgoing_token": "node_a:node_a-password",
-    "ilp_over_http_url": "http://localhost:8770/ilp",
-    "min_balance": -100000,
-    "routing_relation": "Peer"}' \
-    http://localhost:7770/accounts
+printf "Creating Alice's account on Node A...\n"
+cargo run --quiet --bin ilp-cli -- \
```
emschwartz: Ugh this command does not look amazing (because of the `cargo run...` part in front). Would it be terrible to do something like aliasing `ilp-cli` to that whole string? If you were really going to install the CLI on your system you'd do `cargo install ilp-cli` so that it would be available in your path without needing `cargo run...`

examples/simple/README.md
```diff
@@ -536,10 +501,10 @@ Check out the [other examples](../README.md) for more complex demos that show ot
 # For integration tests
 function hook_before_kill() {
     if [ $TEST_MODE -eq 1 ]; then
-        test_equals_or_exit '{"balance":"-500"}' test_http_response_body -H "Authorization: Bearer admin-a" http://localhost:7770/accounts/alice/balance
-        test_equals_or_exit '{"balance":"500"}' test_http_response_body -H "Authorization: Bearer admin-a" http://localhost:7770/accounts/node_b/balance
-        test_equals_or_exit '{"balance":"-500"}' test_http_response_body -H "Authorization: Bearer admin-b" http://localhost:8770/accounts/node_a/balance
-        test_equals_or_exit '{"balance":"500"}' test_http_response_body -H "Authorization: Bearer admin-b" http://localhost:8770/accounts/bob/balance
+        test_equals_or_exit '{"balance":"-500"}' test_http_response_body -H "Authorization: Bearer admin-example" http://localhost:7770/accounts/alice/balance
```
emschwartz: Should the tests also use the CLI? @bstrie @dora-gt

crates/ilp-cli/src/main.rs
gakonst: I agree. 

`interledger-cli accounts <cmd>` where `cmd` matches the [corresponding endpoints](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-api/src/routes/accounts.rs#L388-L400)

```rust
    get_spsp
        .or(get_spsp_well_known)
        .or(post_accounts)
        .or(get_accounts)
        .or(put_account)
        .or(get_account)
        .or(get_account_balance)
        .or(delete_account)
        .or(put_account_settings)
        .or(incoming_payment_notifications)
        .or(post_payments)
        .boxed()
```

crates/ilp-cli/src/main.rs
gakonst: That's probably due to how it's in the API https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-api/src/routes/accounts.rs#L398

crates/ilp-cli/src/main.rs
gakonst: Yeah I prefer 1 positional argument when there's some subcommand, and then use named args for the args of that subcommand

examples/simple/README.md
gakonst: I think we can just add a step in the top level of the examples that they should do `cargo install --release X` and replace all occurrences of `cargo run` with the binary name (or with `./target/{debug,release}/X`

examples/simple/README.md
gakonst: Yeah I'd prefer that

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                .long("print-response")
+                .help("Upon a successful HTTP response, response body will be printed to stdout"),
+        ])
+        .subcommands(vec![
+            // Example: ilp-cli add-account alice ABC 9
+            SubCommand::with_name("add-account")
+                .about("Creates a new account on this node")
+                .args(&[
+                    // Required, positional arguments
+                    Arg::with_name("username")
```
gakonst: I'd prefer if we used named arguments here, what do you think? Combined with the above comment, this would be

`interledger-cli accounts add --username alice --asset-code ABC --asset-scale 9

Also, let's use `-` everywhere in our CLIs instead of underscores?

examples/simple/README.md
```diff
@@ -136,12 +136,12 @@ When you want to watch logs, use the `tail` command. You can use the command lik
...
 
 <!--!
-printf "\nStarting nodes...\n"
+printf "\n\nStarting Interledger nodes...\n"
 if [ "$USE_DOCKER" -eq 1 ]; then
     $CMD_DOCKER run \
         -e ILP_ADDRESS=example.node_a \
         -e ILP_SECRET_SEED=8852500887504328225458511465394229327394647958135038836332350604 \
-        -e ILP_ADMIN_AUTH_TOKEN=admin-a \
+        -e ILP_ADMIN_AUTH_TOKEN=admin-example \
```
gakonst: we had admin-a and admin-b to show that each node can use a different auth token, think it'd be better to leave that as is

examples/simple/README.md
dora-gt: hmmmm difficult question.

This simply checks, finally it worked well or not. So using `cli` here means mixing adulterants in. That said, actually now this is working as an integration test, using `cli` makes it possible to test `cli` itself. It might be better.

Cargo.toml
bstrie: For the sake of being developer-friendly, I must strenuously object to the idea of requiring users to type that every time they want to issue a command. :P Consider the longest names that you've ever seen from a command-line tool; even "ilp-cli" is stretching the limits of what people will type without grousing. We dearly need a shorter synonym for the project, and if pressed I'd rather rename the `interledger` command to `ilp` or `ilp-node` than rename this to match that one. (Having the crates themselves named extremely long things like `interledger-settlement-ethereum` is less of an issue, because presumably we expect people to `use` them once and never worry again.)

crates/ilp-cli/src/main.rs
bstrie: It seems to be general convention in CLIs to use dashes rather than underscores (consider `git cherry-pick`); most CLIs sidestep the issue by never making flags more than a single word. In contrast, Rust identifiers can obviously only contain the latter. In the case here we seek to obviate the need for any manual intervention when converting from CLI arguments into JSON by *exclusively* naming our CLI arguments identically to the JSON fields that the node expects to receive.

I did seek to otherwise impose consistency here: as implemented, the command-line argument names *always* use dashes, and their corresponding entries in the Arguments hashmap will *always* use underscores. The only place that this is ever visible to the user is in the `help` output.

Once thing that I actually have not yet tested is whether Serde is smart enough to automatically deserialize a field named `foo-bar` into `foo_bar`; if that were the case then I would happily use dashes everywhere.

crates/ilp-cli/src/main.rs
bstrie: Absolutely, I'm all for being consistent with the HTTP API here (and not requiring me to think up original names for every subcommand :P ).

crates/ilp-cli/src/main.rs
bstrie: Indeed, rather than having me document what each of these mean only here in the CLI API, there should be a central document that we could refer to. I also don't actually feel qualified to authoritatively claim what many of these arguments even do.

crates/ilp-cli/src/main.rs
bstrie: If we go with the above suggestion we could have it be `accounts alice payments` instead.

crates/ilp-cli/src/main.rs
bstrie: In this case I was using positional arguments for all required arguments and named arguments for all optional arguments, but I'm happy to reconsider.

examples/simple/README.md
bstrie: I'm under the impression that we consider these examples unrepresentative because we don't actually expect anyone to host two logical nodes on the same machine. Furthermore, for any extensive use of this CLI I would encourage people to use the environment variable to avoid the need to type their password repeatedly. In that sense using different passwords for each logical node seems irrelevant to the example, and requiring that the example include additional command-line flags only to accommodate an atypical use case would seem to be a distraction from what we expect the user to take away from this example.

examples/simple/README.md
bstrie: Let's centralize this discussion at https://github.com/interledger-rs/interledger-rs/pull/334#discussion_r328406029.

examples/simple/README.md
bstrie: > I think we can just add a step in the top level of the examples that they should do cargo install --release X and replace all occurrences of cargo run with the binary name (or with ./target/{debug,release}/X

I'll explore the former, but with regard to the latter we might have trouble using a relative path because we can't guarantee from which directory the user will be running this example; we could require that they be in `examples/simple` if that seems reasonable.

examples/simple/README.md
bstrie: I do like the idea of using this to test the CLI as well; since this is a monorepo I think it's reasonable to make the CI test the examples and the CLI in the same test.

crates/ilp-cli/src/main.rs
bstrie: I'll look at implementing the suggestion ( https://github.com/interledger-rs/interledger-rs/pull/334#discussion_r328403702 ), although for consistency with the HTTP API I would probably switch to named arguments for every argument *except* username here, so `interledger-cli accounts add alice --asset-code ABC --asset-scale 9`.

bstrie: > Also, let's use - everywhere in our CLIs instead of underscores?

Let's centralize dicussion of this in https://github.com/interledger-rs/interledger-rs/pull/334#discussion_r328403524 .

examples/simple/README.md
bstrie: I believe that's evident from the `interledger node` command itself. Let's centralize discussion in https://github.com/interledger-rs/interledger-rs/pull/334#discussion_r328406029 .

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+    let mut app = App::new("ilp-cli")
+        .about("Interledger.rs Command-Line Interface")
+        .version(crate_version!())
+        // TODO remove this line once this issue is solved:
+        // https://github.com/clap-rs/clap/issues/1536
+        .after_help("")
+        .args(&[
+            Arg::with_name("authorization_key")
+                .long("auth")
+                .env("ILP_CLI_AUTH")
```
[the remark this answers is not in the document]
bstrie: > I think it is rather API's auth than CLI's auth, so I feel it a bit strange.
Or... ILP_CLI_NODE_AUTH also makes sense to me.

I would not mind considering other names for this.

> Also, we might want to give the token from STDIN.
> It could be another PR, maybe.

This is implemented already, and indicated by the output of `ilp-cli --help`.

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                    Arg::with_name("asset_scale")
+                        .index(3)
+                        .takes_value(true)
+                        .required(true)
+                        .help("The scale of the asset associated with this account"),
+                    // Optional, named arguments
+                    Arg::with_name("ilp_address").long("ilp-address").takes_value(true),
+                    Arg::with_name("max_packet_amount").long("max-packet-amount").takes_value(true),
+                    Arg::with_name("min_balance").long("min-balance").takes_value(true).allow_hyphen_values(true),
+                    Arg::with_name("ilp_over_http_url").long("ilp-over-http-url").takes_value(true),
```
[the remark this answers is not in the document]
bstrie: I was just about to submit this PR to your branch when git told me that your branch no longer existed; turns out you had just merged it into master. :P

bstrie:
Also note that this initial PR doesn't yet update any of the Docker side of things; wasn't sure if that ought to be left to a future PR.

crates/ilp-cli/src/main.rs
bstrie: I've checked and Serde doesn't accept a JSON field named `foo-bar` when looking to populate a Rust field named `foo_bar` (which is entirely reasonable), so if we're bothered by having dashes for one and underscores for the other we could change all the flags to be underscores instead. However, for now I'm defaulting to leaving it as-is, since it's a low priority style change that can be addressed in the future.

crates/ilp-cli/src/main.rs
bstrie: We've discussed this and we're going to look at having two binaries, `ilp-cli` and `ilp-node`.

crates/ilp-cli/src/main.rs
bstrie: Since this is a general issue with our docs I'm leaning towards leaving this as-is for now; I'll put a TODO so that it doesn't just get dropped on the floor.

crates/ilp-cli/src/main.rs
bstrie: Since it seems out of scope to make such a change to the HTTP API right now, I'll leave a TODO to note this as a future improvement.

examples/simple/README.md
bstrie: In the interest of expedience I'm changing these invocations to just `../../target/debug/ilp-cli`, which could be nicer with an alias or symbolic link, but is a good compromise for the moment and is the simplest and most transparent thing that works.

bstrie:
> If we use cli in examples, we have to amend the other examples too.
> I'd be happy to take it on if you want.

@dora-gt, I'm actually using the examples as my to-do list for which APIs to implement first, so no need. :) However, I don't intend on updating the docker side of the examples right now, and I'd be happy to have you work on that (but don't worry about doing that until after I have the examples updated).

Cargo.toml
emschwartz: Changed my mind on this. I think it makes sense to call the binaries `ilp-` and the libraries `interledger-`

examples/simple/README.md
emschwartz: Maybe it's not so bad. Do we need to run it with `--quiet`?

examples/simple/README.md
bstrie: Separate passwords reinstated; now that I've also begin using a symbolic link to get of the `cargo run` invocation, that leaves plenty of room for the occasional `--auth admin-b` argument that the one node will need, and manages to nicely show off the features of the CLI at the same time.

examples/simple/README.md
bstrie: We do want that flag because otherwise you get a few lines of bright green cargo output after every single invocation. I've actually gone even further and begun using a symbolic link here, which ends up looking very nice.

bstrie:
I've just force-pushed a rebased and squashed version with all changes since the last review.

Things not yet done:

* Most of the CLI subcommands still need proper argument parsing and HTTP request logic. However, if people can confirm that the general direction of the PR looks good, this would be a very straightforward addition given how the code is currently structured.
* This commit also updates the XRP settlement example, but the remaining three examples are not yet updated; I wanted to make sure that the general direction here is correct before proceeding further on that task.

Cargo.toml
```diff
@@ -17,4 +17,5 @@ members = [
...
```
emschwartz: nit: put this in alphabetical order

crates/ilp-cli/src/interpreter.rs
```diff
@@ -0,0 +1,223 @@
...
+
+use clap::{App, ArgMatches};
+use reqwest::{Client, Response};
+use std::process::exit;
+
+pub fn run<'a, 'b>(mut app: App<'a, 'b>) {
+    // Parse the CLI input using the defined arguments
+    let matches = app.clone().get_matches();
+
+    let node = Node {
```
emschwartz: How about `NodeClient` or something like that? The thing we're interacting with is the "node"

crates/ilp-cli/src/parser.rs
```diff
@@ -0,0 +1,227 @@
...
+            Arg::with_name("account_username")
+                .index(1)
+                .takes_value(true)
+                .required(true)
+                .help("The username of the account whose balance to return"),
+        )
+}
+
+fn accounts_create<'a, 'b>() -> App<'a, 'b> {
+    // Example: ilp-cli add-account alice ABC 9
```
emschwartz: This example should be updated to use non-positional arguments

crates/ilp-cli/src/parser.rs
```diff
@@ -0,0 +1,227 @@
...
+    SubCommand::with_name("pay")
+                .about("Send a payment from an account on this node")
+                .args(&[
+                    Arg::with_name("sender_username")
+                        .index(1)
+                        .takes_value(true)
+                        .required(true)
+                        .help("The username of the account on this node issuing the payment"),
+                    Arg::with_name("source_amount")
+                        .long("source-amount")
```
emschwartz: What do you think about adding short parameters for this and receiver? (maybe `-a` and `-r`?)

crates/ilp-cli/src/parser.rs
```diff
@@ -0,0 +1,227 @@
...
+                        .takes_value(true)
+                        .required(true)
+                        .help("The amount to transfer from the sender to the receiver, denominated in units of the sender's assets"),
+                    Arg::with_name("receiver")
```
emschwartz: Would it be clearer what you need to put here if the parameter were called "payment pointer" instead of "receiver"?

examples/simple/README.md
```diff
@@ -201,7 +204,7 @@ printf "\nWaiting for Interledger.rs nodes to start up"
 wait_to_serve "http://localhost:7770" 10 || error_and_exit "\nFailed to spin up nodes. Check out your configuration and log files."
 wait_to_serve "http://localhost:8770" 10 || error_and_exit "\nFailed to spin up nodes. Check out your configuration and log files."
 
-printf "done\nThe Interledger.rs nodes are up and running!\n\n"
+printf " done\nThe Interledger.rs nodes are up and running!\n\n"
```
```suggestion
printf "done\nThe Interledger.rs nodes are up and running!\n\n"
```

examples/simple/README.md
```diff
@@ -283,111 +288,72 @@ else
 -->
 
 ```bash
+# We merely use this symbolic link here to pretend that ilp-cli is installed
```
```suggestion
# We merely use this symbolic link here to pretend that ilp-cli is installed
# You would not need this line if you installed `ilp-cli` by running `cargo install ilp-cli`
```

examples/simple/README.md
```diff
@@ -283,111 +288,72 @@ else
...
+ln -s ../../target/debug/ilp-cli ilp-cli
+
+# For authenticating to nodes, we can set credentials as follows
```
```suggestion
# For authenticating to nodes, we can set credentials as an environment variable or a CLI argument
```

examples/xrp-settlement/README.md
```diff
@@ -385,77 +389,60 @@ else
 -->
 
 ```bash
+ln -s ../../target/debug/ilp-cli ilp-cli
```
emschwartz: We shouldn't do this if it doesn't work when run from a different directory

examples/xrp-settlement/README.md
```diff
@@ -385,77 +389,60 @@ else
...
+export ILP_CLI_API_AUTH=hi_alice
```
emschwartz: If this is only used in once, I think it would be better to pass it in as a CLI arg

examples/xrp-settlement/README.md
```diff
@@ -385,77 +389,60 @@ else
...
+    --asset-scale 6 \
+    --max-packet-amount 100 \
+    --settlement-engine-url http://localhost:3001 \
+    --ilp-over-http-incoming-token alice_password \
+    --ilp-over-http-outgoing-token bob:bob_password \
+    --ilp-over-http-url http://localhost:7770/ilp \
+    --settle-threshold 500 \
+    --min-balance -1000 \
+    --settle-to 0 \
+    --routing-relation Peer > logs/account-bob-alice.log &
```
emschwartz: Did you mean to get rid of `2>/dev/null`?

crates/ilp-cli/src/parser.rs
gakonst: As a user, I do not want to know what "payment pointer" means, or have to look at the [RFC](https://interledger.org/rfcs/0026-payment-pointers/). I think "receiver" should be fine here.

crates/ilp-cli/src/interpreter.rs
```diff
@@ -0,0 +1,223 @@
...
+                                command
+                            ),
+                        }
+                    }
+                    _ => {
+                        app.print_help().unwrap();
+                        exit(1);
+                    }
+                },
+                "status" => node.get_root(ilp_cli_matches),
```
gakonst: There's a few commands currently which do not require any authorization, and as such the `--auth` flag should not be required. These are: `/status`, GET `/rates`, GET `/routes`. @emschwartz not sure if you'd want to add auth on the rates and routes endpoints, probably not.

crates/ilp-cli/src/interpreter.rs
```diff
@@ -0,0 +1,223 @@
...
+                        "update" => {
+                            if accounts_matches.is_present("is_admin") {
+                                node.put_account(accounts_matches)
+                            } else {
+                                node.put_account_settings(accounts_matches)
+                            }
+                        }
+                        command => panic!("Unhandled `ilp-cli accounts` subcommand: {}", command),
+                    },
+                    // TODO: subcommand-specific help output
```
gakonst: Yeah that'd be helpful. Also, when i run `./ilp-cli`, I'd prefer if I got the full output of `./ilp-cli --help` (as below) instead of being prompted to add the `--help` flag for more info

```
./target/debug/ilp-cli --help
ilp-cli 0.0.1
Interledger.rs Command-Line Interface

USAGE:
    ilp-cli [FLAGS] [OPTIONS] --auth <authorization_key> [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -q, --quiet      Disable printing the bodies of successful HTTP responses upon receipt
    -V, --version    Prints version information

OPTIONS:
        --auth <authorization_key>    An HTTP bearer authorization token permitting access to the designated operation
                                      [env: ILP_CLI_API_AUTH=]
        --node <node_url>             The base URL of the node to connect to [env: ILP_CLI_NODE_URL=]  [default:
                                      http://localhost:7770]

SUBCOMMANDS:
    accounts              Operations for interacting with accounts
    help                  Prints this message or the help of the given subcommand(s)
    pay                   Send a payment from an account on this node
    rates
    routes
    settlement-engines
    status
```

crates/ilp-cli/src/interpreter.rs
```diff
@@ -0,0 +1,223 @@
...
+                    _ => {
+                        app.print_help().unwrap();
+                        exit(1);
+                    }
+                },
+                "pay" => node.post_account_payments(ilp_cli_matches),
```
gakonst: How about making the sender provided as `--from` and rename all `--receiver` instances to `--to` and `--source-amount` to `--amount` or `--value`? Think this is easier to type. I understand that `source-amount` is more accurate since the exactly the same amount will not get sent to the receiver, but once we get quoting working we could perhaps make the functionality match the looser language. 

```
./target/debug/ilp-cli pay --help
ilp-cli-pay
Send a payment from an account on this node

USAGE:
    ilp-cli --auth <authorization_key> pay <sender_username> --receiver <receiver> --source-amount <source_amount>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --receiver <receiver>              The Payment Pointer or SPSP address of the account receiving the payment
        --source-amount <source_amount>    The amount to transfer from the sender to the receiver, denominated in units
                                           of the sender's assets

ARGS:
```

crates/ilp-cli/src/parser.rs
bstrie: I've done this in my branch, part of my original reluctance is that I don't want to end up with the same short flags having different meanings in different subcommands (even though technically this is no problem to implement), so I wanted to ask about which commands to prioritize implementing short flags for (for example, I was considering using `-a` for the `--auth` top-level flag).

examples/xrp-settlement/README.md
bstrie: I'm not sure what the preferred alternative is; this is replacing the hideous `cargo run --quiet --bin ilp-cli --` that otherwise preceded every command and served to obfuscate what the example was trying to show (and is annoyingly slower in practice, since every individual `cargo run` has to check to see if a recompilation should occur). We could `cargo install` perhaps, but that seems like a more heavyweight solution and I fear that we're going to have users accidentally managing to run old versions of the CLI and running into maddening errors.

examples/xrp-settlement/README.md
bstrie: The env variable gets used in six of the thirteen `ilp-cli` invocations in this example (and could be used in an additional one if admin auth could be used to send payments, as you had mentioned was something we were looking at implementing). In lieu of having an alternative way to authenticate (e.g. a `sudo`-like setup, or a file containing credentials, or an interactive session requiring the user to type their password every time) I'd suggest that the environment variable should be the recommended way of providing credentials, with the stdin approach providing a fallback.

examples/xrp-settlement/README.md
bstrie: Sort of, I wasn't sure why we were silencing stderr output and figured that I would err on the side of being too verbose rather than accidentally quelching an error that we'd rather notice.

crates/ilp-cli/src/interpreter.rs
bstrie: My default stance was to name the command line arguments identically to the field names that we expect in the JSON payload (in case somebody wanted to go from using this tool to using the HTTP API directly), but I'm happy to make this more ergonomic, and in fact I can even implement aliases for these arguments so that people can type either `--to` or `--receiver` at their discretion.

examples/simple/README.md
bstrie: This space is deliberate because the preceding output is the line of repeated `.` that we print while waiting for the nodes to spin up, so this makes the difference between the output looking like `Waiting for Interledger.rs nodes to start up.....done` and `Waiting for Interledger.rs nodes to start up..... done`.

examples/simple/README.md
bstrie: My only concern with suggesting that people use `cargo install` is that I worry about a situation where someone has a stale version of the CLI installed which might end up causing errors that are frustrating for users to diagnose down the road. Let me know if you think the risk here isn't worth worrying about.

crates/ilp-cli/src/interpreter.rs
bstrie: If this is the case then the most straightforward solution seems to be to move the `--auth` flag to exist only underneath those subcommands that require it, since this "uppermost" level is intended for arguments that apply to all subcommands.

examples/xrp-settlement/README.md
emschwartz: Replacing the longer command sounds good as long as it works when called from different directories. I don't actually know if that link is relative to the file it's being called in or from the working directory. If it's the working directory, is there a way to force it to be from the perspective of the file?

examples/simple/README.md
emschwartz: Not worth it in this case. If the CLI isn't stable right now, it should be very soon

crates/ilp-cli/src/parser.rs
emschwartz: That's fair. Maybe not

crates/ilp-cli/src/interpreter.rs
```diff
@@ -21,203 +21,252 @@ pub fn run<'a, 'b>(mut app: App<'a, 'b>) {
...
+        self.client
+            .get(&format!("{}/rates", self.url))
+            .send()
+            .map_err(Error::ClientErr)
     }
 
-    fn put_rates(&self, matches: &ArgMatches) -> reqwest::Result<Response> {
-        unimplemented!()
+    fn put_rates(&self, matches: &ArgMatches) -> Result<Response, Error> {
+        let auth = matches.value_of("authorization_key").unwrap();
```
gakonst: If this fails will it show that the authorization key was not found? Maybe replace the unwrap with a more verbose expect message?

crates/ilp-cli/src/parser.rs
```diff
@@ -0,0 +1,365 @@
...
+fn routes<'a, 'b>() -> App<'a, 'b> {
+    SubCommand::with_name("routes").about("Operations for interacting with the routing table")
+}
+
+fn routes_list<'a, 'b>() -> App<'a, 'b> {
+    SubCommand::with_name("list").about("View this node's routing table")
+}
+
+fn routes_set<'a, 'b>() -> App<'a, 'b> {
+    SubCommand::with_name("set")
```
gakonst:
```suggestion
    SubCommand::with_name("set").about("Set the routes for this node")
```

crates/ilp-cli/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+                    );
+                    exit(1);
+                }
+            }
+        },
+    }
+}
+
+#[cfg(test)]
+// Note that this module contains interface tests, not integration tests.
```
gakonst: Great idea.

crates/ilp-cli/src/parser.rs
```diff
@@ -0,0 +1,365 @@
...
+                .long("quiet")
+                .help("Disable printing the bodies of successful HTTP responses upon receipt"),
+        ])
+}
+
+fn accounts<'a, 'b>() -> App<'a, 'b> {
+    SubCommand::with_name("accounts")
+        .about("Operations for interacting with accounts")
+        .arg(
+            Arg::with_name("is_admin")
```
emschwartz: If this is only used for the update settings method, I think we should just have two different method names for those operations. In general, the node will figure out whether you're doing stuff as the admin or not

crates/ilp-cli/src/parser.rs
```diff
@@ -0,0 +1,365 @@
...
+                .long("admin")
+                .help("Attempts to perform the specified operation as an administrator"),
+        )
+}
+
+fn accounts_balance<'a, 'b>() -> App<'a, 'b> {
+    SubCommand::with_name("balance")
+        .about("Returns the balance of an account")
+        .args(&[
+            Arg::with_name("authorization_key")
```
emschwartz: nit: this is repeated a lot, what about putting it in a lazy_static or separate function?

crates/ilp-cli/src/parser.rs
```diff
@@ -0,0 +1,365 @@
...
+                .index(1)
+                .takes_value(true)
+                .required(true)
+                .help("The username of the account to update"),
+            Arg::with_name("authorization_key")
+                .long("auth")
+                .env("ILP_CLI_API_AUTH")
+                .required(true)
+                .help("An HTTP bearer authorization token permitting access to this operation"),
+            Arg::with_name("asset_code")
```
emschwartz: Since these fields are repeated in two methods, let's put them in a separate function so we only have to update the fields in one place

examples/simple/README.md
```diff
@@ -283,111 +288,74 @@ else
...
-    "asset_scale": 9,
-    "max_packet_amount": 100,
-    "ilp_over_http_incoming_token": "node_b-password",
-    "ilp_over_http_outgoing_token": "node_a:node_a-password",
-    "ilp_over_http_url": "http://localhost:8770/ilp",
-    "min_balance": -100000,
-    "routing_relation": "Peer"}' \
-    http://localhost:7770/accounts
+printf "Creating Alice's account on Node A...\n"
+./ilp-cli --quiet accounts create alice \
```
emschwartz: Beautiful

examples/xrp-settlement/README.md
```diff
@@ -385,77 +389,62 @@ else
...
+    --asset-scale 6 \
+    --max-packet-amount 100 \
+    --settlement-engine-url http://localhost:3000 \
+    --ilp-over-http-incoming-token bob_password \
+    --ilp-over-http-outgoing-token alice:alice_password \
+    --ilp-over-http-url http://localhost:8770/ilp \
+    --settle-threshold 500 \
+    --min-balance -1000 \
+    --settle-to 0 \
+    --routing-relation Peer > logs/account-alice-bob.log &
```
emschwartz: We should make sure it doesn't care about the capitalization of this

crates/ilp-cli/src/interpreter.rs
bstrie: I should sprinkle some more comments around here, basically the gist of this file is that anywhere you see an `unwrap` it's "impossible" to actually trigger a panic because if the field were omitted then Clap would have already errored out during the `get_matches` stage (I say "impossible" in quotes because there could always be a bug in Clap or something).

bstrie: To be clear here's what the message looks like today if this field is omitted:
```
⌁ ./ilp-cli rates set-all -r a 1.0
error: The following required arguments were not provided:
    --auth <authorization_key>

USAGE:
    ilp-cli rates set-all --auth <authorization_key> --rate <rate> <rate>

For more information try --help
```

crates/ilp-cli/src/parser.rs
bstrie: I was thinking about that and it got me thinking about whether or not these two sets of arguments ought to be slightly different; it might make sense to make `asset_code` and `asset_scale` optional for updating an account, while keeping it mandatory for creating an account. But in lieu of that I'll break this out into its own thing for now.

examples/xrp-settlement/README.md
bstrie: Currently `ilp-cli` deliberately does as little inspection and processing of the arguments as possible, so if the user passes "peer" or "Peer" or "pEeR" it will just forward it as-is to the server; the general mindset here is that we could focus on improving error messages on the server side rather than duplicating validation on both client and server. If we did want validation we could certainly use https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.possible_values combined with https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.case_insensitive to make it work, but I'll call that low-priority for now (and if we're going to make any TODOs here, it should be to hash out a policy regarding client-side vs. server-side validation).

PR #336, opened 2019-09-26
Move node to its own crate

emschwartz:
The `interledger` package is now only a library that assembles the rest of the core functionality.
`ilp-node` is the binary for running the node.
The `ilp-` prefix will be used to differentiate binaries from libraries.


examples/eth-xrp-three-nodes/README.md
```diff
@@ -411,7 +411,7 @@ ILP_ADMIN_AUTH_TOKEN=hi_alice \
 ILP_REDIS_URL=redis://127.0.0.1:6379/ \
 ILP_HTTP_BIND_ADDRESS=127.0.0.1:7770 \
 ILP_SETTLEMENT_API_BIND_ADDRESS=127.0.0.1:7771 \
-cargo run --all-features --bin interledger -- node &> logs/node-alice.log &
+cargo run --bin ilp-node &> logs/node-alice.log &
```
dora-gt: If I'm correct, this `--all-features` was maybe required. I'll check later.

dora-gt: I've checked, it is required.
If we don't specify this, it causes another compilation (because the bins are compiled with [`--all-features`](https://github.com/interledger-rs/interledger-rs/pull/336/files/904d5f1642ab8751af4f79856951d7c1a3ec0b67#diff-3f3dd35f7904c604ebf9e19a33ed2062R143)), which is not good.

We cannot omit this because the settlement engine requires this.

Cargo.toml
```diff
@@ -17,6 +18,3 @@ members = [
   "./crates/interledger-stream",
   "./crates/interledger-settlement-engines",
```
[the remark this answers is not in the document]
emschwartz: Yup, that should be changed later. I think we might want to move it into its own repo, but we can deal with that at another point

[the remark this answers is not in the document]
emschwartz: Opened https://github.com/interledger-rs/interledger-rs/issues/343 to track this discussion

crates/ilp-node/README.md
```diff
@@ -0,0 +1,6 @@
+# Interledger.rs
+
+This crate provides both a command-line tool and a library API for the Rust
```
emschwartz: This description is wrong and should be updated to reflect the fact that it's only the node binary now

crates/ilp-node/src/node.rs
```diff
@@ -334,42 +335,30 @@ impl InterledgerNode {
         tokio_run(self.serve());
     }
 
+    #[doc(hidden)]
+    #[allow(dead_code)]
     pub fn insert_account(
```
gakonst: I think we should eventually get rid of this method and just use HTTP requests. Some of the tests already use HTTP requests to create accounts.

emschwartz: Agreed. I had actually taken it out and then realized a bunch of tests used it and didn't feel like updating those right now

Issue #337, opened 2019-09-26
Add a mapping from asset code to settlement engine URL in the node


emschwartz:
Configuring the settlement engine URL on each account is annoying. We should add a global mapping from `asset_code` to settlement engine URL so that the SE URLs can be configured once and used by all the accounts.

PR #339, opened 2019-09-26
Node-wide settlement engine url map + various fixes

### LINKED ISSUE #178: Add identifier for each settlement engine instance
If a settlement engine instance is ever migrated to a different port or IP address, every account linked to that settlement engine must have the URL associated with it updated. 

If there was a separate mapping of settlement engine identifiers to settlement engine URLs, each account could be linked to the immutable settlement engine identifier, and the URL would only have to be updated in a single place.

emschwartz:
Resolves https://github.com/interledger-rs/interledger-rs/issues/337
Resolves #178 

This also includes various fixes related to BTP, CCP, and dynamically setting the ILP address based on the ILDCP response from the parent.

### DIFF
--- a/crates/interledger-api/src/routes/accounts.rs
+++ b/crates/interledger-api/src/routes/accounts.rs
@@ -23,11 +23,7 @@ use log::{debug, error, trace};
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use std::convert::TryFrom;
-use std::time::Duration;
 use warp::{self, Filter};
-
-const MAX_RETRIES: usize = 10;
-const DEFAULT_HTTP_TIMEOUT: Duration = Duration::from_millis(5000);
 
 #[derive(Deserialize, Debug)]
 struct SpspPayRequest {
@@ -534,7 +530,7 @@ where
             let settlement_engine_url = account.settlement_engine_details().map(|details| details.url).or(default_settlement_engine);
             if let Some(se_url) = settlement_engine_url {
                 let id = account.id();
-                let http_client = Client::new(DEFAULT_HTTP_TIMEOUT, MAX_RETRIES);
+                let http_client = Client::default();
                 trace!(
                     "Sending account {} creation request to settlement engine: {:?}",
                     id,
--- a/crates/interledger-api/src/routes/accounts.rs
+++ b/crates/interledger-api/src/routes/accounts.rs
@@ -402,15 +402,15 @@ where
         .boxed()
 }
 
-fn get_address_from_parent_and_update_routes<S, A, T>(
-    mut service: S,
+fn get_address_from_parent_and_update_routes<O, A, S>(
+    mut service: O,
     parent: A,
-    store: T,
+    store: S,
 ) -> impl Future<Item = (), Error = ()>
 where
-    S: OutgoingService<A> + Clone + Send + Sync + 'static,
+    O: OutgoingService<A> + Clone + Send + Sync + 'static,
     A: CcpRoutingAccount + Clone + Send + Sync + 'static,
-    T: AddressStore + Clone + Send + Sync + 'static,
+    S: AddressStore + Clone + Send + Sync + 'static,
 {
     let prepare = IldcpRequest {}.to_prepare();
     service
@@ -477,16 +477,16 @@ where
 // 2b. Perform a RouteControl Request to make them send us any new routes
 // 3. If they have a settlement engine endpoitn configured: Make a POST to the
 //    engine's account creation endpoint with the account's id
-fn connect_to_external_services<S, A, T, B>(
-    service: S,
+fn connect_to_external_services<O, A, S, B>(
+    service: O,
     account: A,
-    store: T,
+    store: S,
     btp: BtpOutgoingService<B, A>,
 ) -> impl Future<Item = A, Error = warp::reject::Rejection>
 where
-    S: OutgoingService<A> + Clone + Send + Sync + 'static,
+    O: OutgoingService<A> + Clone + Send + Sync + 'static,
     A: CcpRoutingAccount + BtpAccount + SettlementAccount + Clone + Send + Sync + 'static,
-    T: AddressStore + Clone + Send + Sync + 'static,
+    S: NodeStore<Account = A> + AddressStore + Clone + Send + Sync + 'static,
     B: OutgoingService<A> + Clone + 'static,
 {
     // Try to connect to the account's BTP socket if they have
@@ -501,48 +501,57 @@ where
     };
 
     btp_connect_fut.and_then(move |_| {
-    // If we added a parent, get the address assigned to us by
-    // them and update all of our routes
-    let get_ilp_address_fut = if account.routing_relation() == RoutingRelation::Parent {
-        Either::A(
-            get_address_from_parent_and_update_routes(service, account.clone(), store)
-            .map_err(|_| {
-                warp::reject::custom(ApiError::InternalServerError)
-            })
-        )
-    } else {
-        Either::B(ok(()))
-    };
+        // If we added a parent, get the address assigned to us by
+        // them and update all of our routes
+        let get_ilp_address_fut = if account.routing_relation() == RoutingRelation::Parent {
+            Either::A(
+                get_address_from_parent_and_update_routes(service, account.clone(), store.clone())
+                .map_err(|_| {
+                    warp::reject::custom(ApiError::InternalServerError)
+                })
+            )
+        } else {
+            Either::B(ok(()))
+        };
 
-    // Register the account with the settlement engine
-    // if a settlement_engine_url was configured on the account
-    get_ilp_address_fut.and_then(move |_|
-    if let Some(se_details) = account.settlement_engine_details() {
-        let se_url = se_details.url;
-        let id = account.id();
-        let http_client = Client::new(DEFAULT_HTTP_TIMEOUT, MAX_RETRIES);
-        trace!(
-            "Sending account {} creation request to settlement engine: {:?}",
-            id,
-            se_url.clone()
-        );
-        Either::A(
-            http_client.create_engine_account(se_url, id)
+        let default_settlement_engine_fut = store.get_asset_settlement_engine(account.asset_code())
             .map_err(|_| {
                 warp::reject::custom(ApiError::InternalServerError)
-            })
-            .and_then(move |status_code| {
-                if status_code.is_success() {
-                    trace!("Account {} created on the SE", id);
-                } else {
-                    error!("Error creating account. Settlement engine responded with HTTP code: {}", status_code);
-                }
-                Ok(())
-            })
-            .and_then(move |_| {
-                Ok(account)
-            }))
-    } else {
-        Either::B(ok(account))
-    })})
+            });
+
+        // Register the account with the settlement engine
+        // if a settlement_engine_url was configured on the account
+        // or if there is a settlement engine configured for that
+        // account's asset_code
+        default_settlement_engine_fut.join(get_ilp_address_fut).and_then(move |(default_settlement_engine, _)| {
+            let settlement_engine_url = account.settlement_engine_details().map(|details| details.url).or(default_settlement_engine);
+            if let Some(se_url) = settlement_engine_url {
+                let id = account.id();
+                let http_client = Client::new(DEFAULT_HTTP_TIMEOUT, MAX_RETRIES);
+                trace!(
+                    "Sending account {} creation request to settlement engine: {:?}",
+                    id,
+                    se_url.clone()
+                );
+                Either::A(
+                    http_client.create_engine_account(se_url, id)
+                    .map_err(|_| {
+                        warp::reject::custom(ApiError::InternalServerError)
+                    })
+                    .and_then(move |status_code| {
+                        if status_code.is_success() {
+                            trace!("Account {} created on the SE", id);
+                        } else {
+                            error!("Error creating account. Settlement engine responded with HTTP code: {}", status_code);
+                        }
+                        Ok(())
+                    })
+                    .and_then(move |_| {
+                        Ok(account)
+                    }))
+            } else {
+                Either::B(ok(account))
+            }
+        })
+    })
 }
--- a/crates/interledger-api/src/routes/accounts.rs
+++ b/crates/interledger-api/src/routes/accounts.rs
@@ -412,6 +412,11 @@ where
     A: CcpRoutingAccount + Clone + Send + Sync + 'static,
     S: AddressStore + Clone + Send + Sync + 'static,
 {
+    debug!(
+        "Getting ILP address from parent account: {} (id: {})",
+        parent.username(),
+        parent.id()
+    );
     let prepare = IldcpRequest {}.to_prepare();
     service
         .send_request(OutgoingRequest {
@@ -428,7 +433,7 @@ where
                     err
                 );
             });
-            debug!("Got ILDCP response: {:?}", response);
+            debug!("Got ILDCP response from parent: {:?}", response);
             let ilp_address = match response {
                 Ok(info) => info.ilp_address(),
                 Err(_) => return err(()),
@@ -436,6 +441,7 @@ where
             ok(ilp_address)
         })
         .and_then(move |ilp_address| {
+            debug!("ILP address is now: {}", ilp_address);
             // TODO we may want to make this trigger the CcpRouteManager to request
             let prepare = RouteControlRequest {
                 mode: Mode::Sync,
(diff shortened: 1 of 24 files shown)


crates/interledger-api/src/routes/node_settings.rs
```diff
@@ -166,11 +167,31 @@ where
         })
         .boxed();
 
+    // PUT /settlement/engines
+    let put_settlement_engines = warp::put2()
```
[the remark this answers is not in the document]
emschwartz: Yup, good catch

crates/interledger-store-redis/src/store.rs
```diff
@@ -1767,6 +1747,55 @@ fn update_routes(
...
+type LoadAccountsResult = (SharedConnection, Vec<AccountWithEncryptedTokens>);
+/// Load accounts from Redis
+/// Also, if an account does not have a settlement_engine_url configured
+/// but there is one configured globally for that account's asset_code,
+/// then set the account's settlement_engine_url to the global one
+fn load_accounts(
+    connection: SharedConnection,
+    account_ids: Vec<AccountId>,
+) -> impl Future<Item = (SharedConnection, Vec<AccountWithEncryptedTokens>), Error = ()> + Send {
+    // TODO can we make these two calls with one pipeline call
```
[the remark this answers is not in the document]
emschwartz: Agreed. I tried for 2-3 hours this evening to make the lua script to work but couldn't get it. I will give it another try tomorrow but if I can't figure it out quickly I'll just leave a TODO about it

crates/interledger-store-redis/src/store.rs
```diff
@@ -1767,6 +1747,55 @@ fn update_routes(
...
+            )| {
+                debug!(
+                    "settlement engines: {:?}, accounts: {:?}",
+                    settlement_engines, accounts
+                );
+                for account in accounts.iter_mut() {
+                    if account.account.settlement_engine_url.is_none() {
+                        debug!("account {} does not have se", account.account.username);
+                        if let Some(url) =
+                            settlement_engines.get(account.account.asset_code.as_str())
```
kincaidoneil: What if you're running multiple settlement engines for the same asset, e.g. XRP on-ledger and XRP paychan? (Common scenario on old live network). In that case, do you still have to the set URL for each account?

kincaidoneil:
Also resolves #178  👀

crates/interledger-store-redis/src/store.rs
emschwartz: Yes

crates/interledger-store-redis/src/store.rs
```diff
@@ -1767,6 +1747,55 @@ fn update_routes(
...
+            |((connection, settlement_engines), (_, mut accounts)): (
+                LoadSettlementEnginesResult,
+                LoadAccountsResult,
+            )| {
+                debug!(
+                    "settlement engines: {:?}, accounts: {:?}",
+                    settlement_engines, accounts
+                );
+                for account in accounts.iter_mut() {
+                    if account.account.settlement_engine_url.is_none() {
```
dora-gt: Just a question, does this work by just setting SE?
There isn't any initial process?

crates/interledger-api/src/routes/node_settings.rs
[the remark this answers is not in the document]
dora-gt: ^ this might not be true. If you assume that people add these *once* when they initially set up their nodes, it seems sufficient to provide `set` function.

crates/interledger-store-redis/src/store.rs
dora-gt: For example, https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-api/src/routes/accounts.rs#L151

crates/interledger-store-redis/src/store.rs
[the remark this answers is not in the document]
emschwartz: It's a bit janky but I finally got it. Left a TODO in case we figure out a way to clean up the script

crates/interledger-store-redis/src/store.rs
emschwartz: Ooh good catch! You're totally right

emschwartz: Implemented in https://github.com/interledger-rs/interledger-rs/pull/339/commits/92092c1384f87f2c8b947614959c2b065f1c24e6

emschwartz: Implemented in https://github.com/interledger-rs/interledger-rs/pull/339/commits/92092c1384f87f2c8b947614959c2b065f1c24e6

crates/interledger-store-redis/src/store.rs
kincaidoneil: Should we support that? (e.g. a custom identifier for each settlement engine instance, which could be the asset code, but is more extensible?)

emschwartz:
The original commit in this PR is currently causing the settlement tests to fail and I can't figure out why right now. I'll work on this later.

It seemed like the problem could be a lua script issue but the main problem also shows up without the commit that changes load accounts to use a lua script.

emschwartz:
The failing test made me realize there were actually a bunch of problems with BTP and CCP so I added some fixes and an additional test to fix the other issues

crates/interledger-store-redis/src/store.rs
emschwartz: I don't think so because then you need to configure each account with that identifier (that isn't the `asset_code`), so at that point you might as well just set the URL on the account

emschwartz:
Note the three_node test still seems somewhat flaky on CI (see failing build [here]()) but I can't reproduce the issue locally. It might be related to how long we wait for nodes to start up or accounts to be created.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -501,48 +508,57 @@ where
...
-        Either::B(ok(account))
-    })})
+            });
+
+        // Register the account with the settlement engine
+        // if a settlement_engine_url was configured on the account
+        // of if there is a settlement engine configured for that
+        // account's asset_code
+        default_settlement_engine_fut.join(get_ilp_address_fut).and_then(move |(default_settlement_engine, _)| {
+            let settlement_engine_url = account.settlement_engine_details().map(|details| details.url).or(default_settlement_engine);
```
dora-gt: If my recognition is correct, this doesn't look sufficient because... this works in the following case:

- Set global settlement engines -> Add accounts

but the following procedure would not work:

- Add accounts -> Set global settlement engines

dora-gt: That said, this might not happen so often.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -501,48 +508,57 @@ where
...
-                Ok(account)
-            }))
-    } else {
-        Either::B(ok(account))
-    })})
+            });
+
+        // Register the account with the settlement engine
+        // if a settlement_engine_url was configured on the account
+        // of if there is a settlement engine configured for that
```
gakonst:
```suggestion
        // or if there is a settlement engine configured for that
```

crates/interledger-ccp/src/server.rs
```diff
@@ -198,6 +202,9 @@ where
                 .map(|index| ilp_address.to_bytes().slice_to(index + 1))
                 .unwrap_or_else(|| ilp_address.to_bytes().clone());
             *self.ilp_address.write() = ilp_address;
+
+            // TODO should we reset our routes and request new ones
```
gakonst: Yes. The store's `set_ilp_address` method updates the stored routes to the current routing table, but I agree that we need to update our routes before that operation. I'll try to fix that.

crates/interledger-ccp/src/server.rs
```diff
@@ -1082,7 +1138,7 @@ mod handle_route_control_request {
         assert_eq!(update.from_epoch_index, 0);
         assert_eq!(update.to_epoch_index, 1);
         assert_eq!(update.current_epoch_index, 1);
-        assert_eq!(update.new_routes.len(), 2);
```
gakonst: This is because the connector's route is also sent in the first epoch right?

crates/interledger-ccp/src/server.rs
```diff
@@ -695,6 +735,22 @@ where
         // Merge the new routes and withdrawn routes from all of the given epochs
         let mut new_routes: Vec<Route> = Vec::with_capacity(epochs_to_take);
         let mut withdrawn_routes: Vec<Bytes> = Vec::new();
+
+        // Include our own prefix if its the first update
```
gakonst: Why only on the first update? What if they rejected it the first time, and then they added us as an account, hence they start accepting it?

crates/interledger-ccp/src/server.rs
```diff
@@ -566,36 +584,51 @@ where
...
-                                let old_route = forwarding_table.get_route(&prefix);
-                                if old_route.is_none() || old_route.unwrap().0.id() != account.id() {
-                                    route.path.insert(0, ilp_address.to_bytes());
-                                    // Each hop hashes the auth before forwarding
-                                    route.auth = hash(&route.auth);
-                                    forwarding_table.set_route(prefix.clone(), account.clone(), route.clone());
-                                    new_routes.push(route);
-                                }
+                        // or that advertise the whole global prefix
+                        let correct_global_prefix = route.prefix.starts_with(&global_prefix[..])
```
gakonst: We should definitely include some kind of diagram illustrating this process in the docs. The slides that were shared with me from the ILP Summit do not cover all these cases and it'd be nice to have all the scenarios we want to cover well documented. Also maybe expand the tests

crates/interledger-router/src/router.rs
```diff
@@ -121,7 +116,26 @@ where
                     }),
             )
         } else {
-            error!("No route found for request: {:?}", request);
+            error!(
+                "No route found for request{}: {:?}",
+                {
+                    // Log a warning if the global prefix does not match
+                    let mut segments = (&self.ilp_address as &str).split(|c| c == '.');
```
gakonst: nit: We have [Address.segments()](https://github.com/bstrie/interledger-rs/blob/27d7c60eb03c1d9f6c1f60e8f55dd251e389c7cb/crates/interledger-packet/src/address.rs#L151)

crates/interledger-store-redis/tests/accounts_test.rs
```diff
@@ -78,40 +78,49 @@ fn insert_accounts() {
...
+                });
+                let ilp_address = Address::from_str("test.parent.our_address").unwrap();
+                store
+                    .set_ilp_address(ilp_address.clone())
+                    .and_then(move |_| {
+                        let ret = store.get_ilp_address();
+                        assert_eq!(ilp_address, ret);
+                        store.get_all_accounts().and_then(move |accounts: Vec<_>| {
+                            let mut accounts = accounts.clone();
+                            accounts.sort_by(|a, b| {
```
gakonst: nit: You should be able to [derive Eq and PartialEq for Username](https://github.com/bstrie/interledger-rs/blob/51a0faa54f994e2b7dd28c1a65bf50671779386a/crates/interledger-service/src/auth/username.rs#L23) and do this comparison.

crates/interledger-stream/src/client.rs
```diff
@@ -37,21 +37,32 @@ where
...
-            congestion_controller: CongestionController::default(),
-            pending_requests: Cell::new(Vec::new()),
-            delivered_amount: 0,
-            should_send_source_account: true,
-            sequence: 1,
-            rejected_packets: 0,
-            error: None,
+        .and_then(move |account_details| {
+            let source_account = account_details.ilp_address();
+            let source_scheme = source_account.segments().next().expect("Addresses should have a scheme first");
```
gakonst: This is always guaranteed by usage of the `Address` data type and should never be hit. I guess it's fine to panic here

crates/interledger-api/src/routes/accounts.rs
gakonst: Each time there's a change in the global settlement engine, we want to iterate over all accounts which use the global and load the account with that address.

@emschwartz note how this is a similar situation with the Account's ILP Address being stored in the account struct instead of being able to dynamically configure it with some extra information which is contained in the store. We could alter the `Account::try_from` to take an optional `settlement_engine_url` which would be the default engine URL for that asset and use that if no url was configured in the account.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -501,48 +508,57 @@ where
...
-            .map_err(|_| {
-                warp::reject::custom(ApiError::InternalServerError)
-            })
-        )
-    } else {
-        Either::B(ok(()))
-    };
+        // If we added a parent, get the address assigned to us by
+        // them and update all of our routes
+        let get_ilp_address_fut = if account.routing_relation() == RoutingRelation::Parent {
```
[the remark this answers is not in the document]
gakonst: yup

crates/interledger-ccp/src/server.rs
emschwartz: Yes

crates/interledger-ccp/src/server.rs
emschwartz: In that case I think they will send a `RouteControlRequest` to request all the routes starting from epoch 0. If so, our address will be included

crates/interledger-api/src/routes/accounts.rs
emschwartz: Good point @dora-gt. Right now, the correct `settlement_engine_url` will be loaded because the lua script that loads the accounts will fill it in with the globally configured one. However, you're correct that it won't create the account on the settlement engine right now.

emschwartz: Implemented in https://github.com/interledger-rs/interledger-rs/pull/339/commits/37ff7db829d0a5fd79c65d4648983d87bf169dc7

crates/interledger-packet/src/address.rs
```diff
@@ -156,6 +156,15 @@ impl Address {
         }
     }
 
+    /// Returns the first segment of the address, which is the scheme.
+    /// See [`IL-RFC 15: ILP Addresses](https://github.com/interledger/rfcs/blob/master/0015-ilp-addresses/0015-ilp-addresses.md#allocation-schemes)
+    pub fn scheme(&self) -> &str {
```
gakonst: FYI we had added that method but decided to remove it in the relevant PR https://github.com/interledger-rs/interledger-rs/pull/75/commits/fde033c11670b0c232a7d77c870a0b402ea3eb76#diff-8c5852cf578f26f54fc6aa011fb65509R151

crates/interledger-packet/src/address.rs
```diff
@@ -343,6 +352,21 @@ mod test_address {
         );
     }
 
+    #[test]
+    fn test_scheme() {
```
gakonst: These should be moved to https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-packet/src/address.rs#L287

crates/interledger-api/src/routes/node_settings.rs
```diff
@@ -178,14 +179,56 @@ where
...
-                .and_then(move |_| Ok(reply))
+                .and_then(move |_| {
+                    // Create the accounts on the settlement engines for any
+                    // accounts that are using the default settlement engine URLs
+                    // (This is done in case we modify the globally configured settlement
+                    // engine URLs after accounts have already been added)
+
+                    // TODO we should come up with a better way of ensuring
+                    // the accounts are created that doesn't involve loading
+                    // all of the accounts from the database into memory
```
gakonst: We could page over the accounts instead of loading them all together

crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs
```diff
@@ -187,9 +187,13 @@ fn eth_xrp_interoperable() {
...
 
-    let bob_fut = set_node_settlement_engines(node2_http, bob_settlement_engines, "admin")
-        .join(create_account_on_node(node2_http, alice_on_bob, "admin"))
-        .join(create_account_on_node(node2_http, charlie_on_bob, "admin"));
+    let bob_fut = create_account_on_node(node2_http, alice_on_bob, "admin")
+        .join(create_account_on_node(node2_http, charlie_on_bob, "admin"))
+        // Setting the settlement engines after the accounts are created should
+        // still trigger the call to create their accounts on the settlement engines
+        .and_then(move |_| {
+            set_node_settlement_engines(node2_http, bob_settlement_engines, "admin")
```
gakonst: nice.

crates/interledger-packet/src/address.rs
emschwartz: True. My bad for suggesting removing it. I think it's a close call on whether it should be part of the address but in this case it seems like it makes sense to include it because otherwise you'd need `unwrap`s elsewhere while it's clear as part of the `Address` definition that that can't panic.

crates/interledger-api/src/routes/node_settings.rs
emschwartz: Yup, there's a TODO [here](https://github.com/interledger-rs/interledger-rs/blob/7428c26032672073af6c52843e53a3b4b294b929/crates/interledger-api/src/lib.rs#L92) for that

emschwartz:
I'm going to merge this now. If there are other changes we want to make to this behavior, we can do them in follow up PRs

PR #342, opened 2019-09-27
Handle leftover amounts on the connector side

gakonst:
Previously, when receiving a settlement request, the connector would scale it to the account's scale, save it to the store, and then upscale the amount back to the engine's scale and return it to the engine. The engine would then compare it with the amount that they sent, and if there was any precision loss, it'd save that as a leftover amount, which would get queued up in the next settlement call.

As discussed, modifying the state based on the connector's response is bad, so we instead change is so that the receiver of the request performs any operations necessary to ensure there's no precision loss in the long run. 

A follow up PR will adjust the engine's functionality.

Flaky until #365 is merged


crates/interledger-settlement/src/api.rs
```diff
@@ -149,24 +150,36 @@ impl_web! {
...
         fn do_receive_settlement(&self, account_id: String, body: Quantity, idempotency_key: Option<String>) -> Box<dyn Future<Item = (StatusCode, Bytes), Error = (StatusCode, String)> + Send> {
             let store = self.store.clone();
-            let amount = body.amount;
+            let store_clone = self.store.clone();
+            let engine_amount = body.amount;
             let engine_scale = body.scale;
-            Box::new(result(A::AccountId::from_str(&account_id)
+
+            // Convert to the desired data types
+            let account_id = match A::AccountId::from_str(&account_id) {
```
gakonst: note: We do futures-error-handling in a few places in the code, while instead we can use synchronous error handling like this. I prefer that and am starting to slowly change the code to use this pattern. do you agree?

crates/interledger-settlement/src/api.rs
```diff
@@ -178,42 +191,41 @@ impl_web! {
                 }
             })
             .and_then(move |account| {
-                result(BigUint::from_str(&amount))
-                .map_err(move |_| {
-                    let error_msg = format!("Could not convert amount: {:?}", amount);
+                let account_id = account.id();
+                let asset_scale = account.asset_scale();
+                // scale to account's scale from the engine's scale
+                let (scaled_engine_amount, precision_loss) = scale_with_precision_loss(engine_amount, asset_scale, engine_scale);
```
gakonst: If we're downscaling we might have some precision error which we must save as leftovers. Upscaling is OK since we're using biguint's.

crates/interledger-settlement/src/api.rs
```diff
@@ -178,42 +191,41 @@ impl_web! {
...
+
+                // load the leftovers and downscale them to the account's asset scale
+                store_clone.load_uncredited_settlement_amount(account_id, asset_scale)
```
gakonst: This will load any leftovers (which are saved in the highest so far received scale by the engine), will downscale them to the account's asset scale and return them. If there was any precision loss due to downscaling, it will also update the leftovers to the new leftovers value

crates/interledger-store-redis/src/store.rs
```diff
@@ -1734,6 +1742,153 @@ impl SettlementStore for RedisStore {
...
+    ) -> Box<dyn Future<Item = (Self::AssetType, u8), Error = ()> + Send> {
+        let mut pipe = redis::pipe();
+        pipe.lrange(uncredited_amount_key(account_id.to_string()), 0, -1);
+        Box::new(
+            pipe.query_async(self.connection.as_ref().clone())
+                .map_err(move |err| error!("Error getting uncredited_settlement_amount {:?}", err))
+                // todo: There must be some way to properly deserialize this when
+                // called with lrange instead of us having to manually parse it as
+                // a list of strings here. FromRedisValue did not seem to work very well.
+                .and_then(move |(_, values): (_, Vec<Vec<String>>)| {
```
gakonst: How can we save tuples in redis and load them as tuples with `lrange`? I inplemented `FromRedisValue` for `AmountWithScale`, but instead it loads a list of strings. Maybe I should implement it for `Vec<AmountWithScale>`?

crates/interledger-settlement/src/api.rs
```diff
@@ -149,24 +150,36 @@ impl_web! {
...
+                    let error_msg = format!("Unable to parse account id: {}", account_id);
+                    error!("{}", error_msg);
+                    return Box::new(err((StatusCode::from_u16(400).unwrap(), error_msg)))
+                }
+            };
+
+            let engine_amount = match BigUint::from_str(&engine_amount) {
+                Ok(a) => a,
+                Err(_) => {
+                    let error_msg = format!("Could not convert amount: {:?}", engine_amount);
```
emschwartz: nit: just put the string in the `error!` macro

crates/interledger-settlement/src/api.rs
```diff
@@ -178,42 +191,41 @@ impl_web! {
...
             .and_then(move |account| {
-                result(BigUint::from_str(&amount))
-                .map_err(move |_| {
-                    let error_msg = format!("Could not convert amount: {:?}", amount);
+                let account_id = account.id();
+                let asset_scale = account.asset_scale();
+                // scale to account's scale from the engine's scale
+                let (scaled_engine_amount, precision_loss) = scale_with_precision_loss(engine_amount, asset_scale, engine_scale);
+
+                // load the leftovers and downscale them to the account's asset scale
```
emschwartz: Just to clarify, you aren't making assumptions about which has the higher scale, right?

crates/interledger-settlement/src/api.rs
emschwartz: Comments like this should go in the code :)

crates/interledger-settlement/src/api.rs
```diff
@@ -303,7 +315,37 @@ impl_web! {
...
+    amount: BigUint,
+    local_scale: u8,
+    remote_scale: u8,
+) -> (BigUint, BigUint) {
+    let scaled = amount
+        .normalize_scale(ConvertDetails {
+            from: remote_scale,
+            to: local_scale,
+        })
+        .unwrap();
```
emschwartz: Why is this being unwrapped? We should make sure this code can't panic

crates/interledger-settlement/src/api.rs
```diff
@@ -303,7 +315,37 @@ impl_web! {
...
+
+    if local_scale < remote_scale {
+        // Upscale it again, and return any precision loss
+        let upscaled = scaled
+            .normalize_scale(ConvertDetails {
```
emschwartz: I don't understand why we would need to call this twice

crates/interledger-settlement/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![recursion_limit = "128"]
+#![recursion_limit = "256"]
```
emschwartz: Did you actually get an error with the lower limit? I wonder what's causing that :thinking:

crates/interledger-store-redis/src/account.rs
```diff
@@ -420,7 +420,7 @@ impl FromRedisValue for AccountWithEncryptedTokens {
     }
 }
 
-fn get_value<V>(key: &str, map: &HashMap<String, Value>) -> Result<V, RedisError>
+pub fn get_value<V>(key: &str, map: &HashMap<String, Value>) -> Result<V, RedisError>
```
emschwartz: Why does this need to be public?

crates/interledger-store-redis/src/store.rs
```diff
@@ -179,6 +183,10 @@ static ROUTES_KEY: &str = "routes:current";
 static STATIC_ROUTES_KEY: &str = "routes:static";
 static STREAM_NOTIFICATIONS_PREFIX: &str = "stream_notifications:";
 
+fn uncredited_amount_key(account_id: impl ToString) -> String {
+    format!("uncredited-amount:{}", account_id.to_string())
```
emschwartz: nit: I think we have inconsistent uses of dashes and underscores in our redis schemas

crates/interledger-store-redis/src/store.rs
emschwartz: Implementing `FromRedisValue` for `AmountWithScale` should be enough because `FromRedisValue` is implemented for [`Vec<T: FromRedisValue>`](https://docs.rs/redis/0.12.0/redis/trait.FromRedisValue.html#impl-FromRedisValue-for-Vec%3CT%3E). What was the problem you ran into? At the very least, you should be able to make it `Vec<(String, String)>` so you don't need to iterate over the inner `Vec`

crates/interledger-store-redis/src/store.rs
```diff
@@ -1734,6 +1742,153 @@ impl SettlementStore for RedisStore {
...
+                .map_err(move |err| error!("Error getting uncredited_settlement_amount {:?}", err))
+                // todo: There must be some way to properly deserialize this when
+                // called with lrange instead of us having to manually parse it as
+                // a list of strings here. FromRedisValue did not seem to work very well.
+                .and_then(move |(_, values): (_, Vec<Vec<String>>)| {
+                    let len = values[0].len();
+                    let mut values = values[0].iter();
+                    let mut amounts = Vec::new();
+                    let mut max_scale = 0;
+                    for _ in (0..len).step_by(2) {
```
emschwartz: This seems super janky. There has got to be a better way to do this parsing

crates/interledger-store-redis/src/store.rs
```diff
@@ -1734,6 +1742,153 @@ impl SettlementStore for RedisStore {
...
+        let connection = self.connection.clone();
+        trace!("Loading uncredited_settlement_amount {:?}", account_id);
+        Box::new(
+            self.get_uncredited_settlement_amount(account_id)
+                .and_then(move |amount| {
+                    // scale the amount from the max scale to the local scale, and then
+                    // save any potential leftovers to the store
+                    let (scaled_amount, precision_loss) =
+                        scale_with_precision_loss(amount.0, local_scale, amount.1);
+                    let mut pipe = redis::pipe();
```
```suggestion
                    let mut pipe = redis::pipe();
                    pipe.atomic();
```

crates/interledger-settlement/src/api.rs
gakonst: It's BigUint's and their multiplications [can never fail](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-settlement/src/lib.rs#L154-L165) . The return value of `normalize_scale` is a Result, because ti can also be implemented for u64/f64 where overflows can happen.

crates/interledger-settlement/src/api.rs
gakonst: We go frmo remote scale to local scale, and then take the result and upscale it again to check if there was any precision loss. Note how `from` and `to` are reversed in the 2 calls.

crates/interledger-settlement/src/api.rs
gakonst: I return it in the error response right below, hence why i assign it to a variable, since it's used in both places

crates/interledger-settlement/src/api.rs
gakonst: I am not, the test ensures that it works for downscaling and upscaling

crates/interledger-store-redis/src/store.rs
gakonst: Refactored it slightly, this is an issue because of using `lrange`

crates/interledger-settlement/src/api.rs
emschwartz: One of the reasons I didn't do this elsewhere is that the error type might not match everywhere so you end up needing a ton of `Either::A`'s everywhere to avoid the "match arms have different types" error.

crates/interledger-settlement/src/api.rs
```diff
@@ -303,7 +322,42 @@ impl_web! {
...
     }
+}
 
+pub fn scale_with_precision_loss(
+    amount: BigUint,
+    local_scale: u8,
+    remote_scale: u8,
+) -> (BigUint, BigUint) {
+    // It's safe to unwrap here since BigUint's normalize_scale
+    // [cannot fail](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-settlement/src/lib.rs#L154-L165)
```
emschwartz: Including a link to the master branch is brittle because the line numbers may change

crates/interledger-settlement/src/lib.rs
```diff
@@ -98,6 +98,35 @@ pub trait IdempotentStore {
...
+    /// If any precision loss occurs during the scaling, it should be saved as
+    /// the new leftover value.
+    fn load_uncredited_settlement_amount(
+        &self,
+        account_id: Self::AccountId,
+        local_scale: u8,
+    ) -> Box<dyn Future<Item = Self::AssetType, Error = ()> + Send>;
+
+    // Gets the current amount of leftovers in the store
+    fn get_uncredited_settlement_amount(
```
emschwartz: What's the point of including this? This method is kind of a foot-gun so I'd prefer to only have `load_...`. If it's for tests, we can just issue the commands to the DB to check what's stored

crates/interledger-settlement/src/api.rs
gakonst: I think I prefer Either's to multiple levels of nested closures

crates/interledger-settlement/src/lib.rs
gakonst: This is in preparation for allowing pluggable backends. A backend would not implement the `load` method. A backend implementer would only have to implement the `save` and `get` methods, and the consuming store would do the following in the `load` method: 

```
backend.get_uncredited_settlement_amount
<scale it here>
backend.save_uncredited_settlement_amount
```

emschwartz: That is unsafe because two threads could call get and save at the same time. It is crucial to do the operation in one atomic step

gakonst:
@emschwartz the race condition should no longer be present as of https://github.com/interledger-rs/interledger-rs/pull/342/commits/02a971824a1bf4b9ff005ba675a650750cc78585, this is good for one more roudn of review

PR #348, opened 2019-09-27
Fix readme headers

pgarg-ripple:
Minor modifications to README headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ cargo run
 
 ```bash #
 # Passing from STDIN in JSON, TOML, YAML format.
-some_command | cargo run -- node
+some_command | cargo run
 ```
 
 #### Configuration files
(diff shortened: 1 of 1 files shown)


dora-gt:
I just mean

- https://github.com/interledger-rs/interledger-rs/pull/348/files#diff-04c6e90faac2675aa89e2176d2eec7d8R84
- https://github.com/interledger-rs/interledger-rs/pull/348/files#diff-04c6e90faac2675aa89e2176d2eec7d8R94

I think these would cause errors like "the config file `node` was not found."

pgarg-ripple:
@dora-gt, just confirming:

cargo run -- node config.yml -> cargo run -- config.yml 
(Is the -- still required?)

some_command | cargo run -- node -> some_command | cargo run

dora-gt:
> cargo run -- node config.yml -> cargo run -- config.yml

Yes.
I think both works, but specifying `--` might be more clear.

> some_command | cargo run -- node -> some_command | cargo run

Yes.

PR #351, opened 2019-09-29
Make services fetch ILP Address from the store

gakonst:
Instead of giving the address once in the constructor.


gakonst:
opened the pr twice for some reason

PR #352, opened 2019-09-29
Fetch ILP Address from the store

gakonst:
Previously we were constructing the value once, and did not account for any changes to the node's ilp address due to adding a parent.


emschwartz:
@gakonst can you rebase this and maybe squash the commits into one? It seems strange to have one commit that changes how something works and then have a bunch of other commits after "fixing" the fact that they don't work with the new change yet

PR #355, opened 2019-09-30
Make Three Nodes test work reliably

gakonst:
This test was flaky, we went through a few rounds of ideas, ended up modifying the test to avoid race conditions altogether.

### DIFF
--- a/crates/interledger-store-redis/src/store.rs
+++ b/crates/interledger-store-redis/src/store.rs
@@ -1383,9 +1383,24 @@ impl AddressStore for RedisStore {
                             pipe.hdel(ROUTES_KEY, account.ilp_address.as_bytes())
                                 .ignore();
 
-                            let new_ilp_address = ilp_address_clone
-                                .with_suffix(account.username().as_bytes())
-                                .unwrap();
+                            // if the username of the account ends with the
+                            // node's address, we're already configured so no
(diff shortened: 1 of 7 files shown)


crates/interledger-store-redis/src/store.rs
```diff
@@ -1383,9 +1383,24 @@ impl AddressStore for RedisStore {
...
-                                .with_suffix(account.username().as_bytes())
-                                .unwrap();
+                            // if the username of the account ends with the
+                            // node's address, we're already configured so no
+                            // need to append anything.
+                            let ilp_address_clone2 = ilp_address_clone.clone();
+                            // Note: We are assuming that if the node's address
+                            // ends with the account's username, then this
+                            // account represents the node's non routing
+                            // account. Is this a reasonable assumption to make?
```
emschwartz: I'm not sure it is :thinking:

gakonst:
OK so the race condition in `three_nodes` is fixed and the bug is no longer encountered. We additionally wait some time after inserting the accounts before making any payments, to ensure that our node receives a few broadcasts from Bob and is still able to route properly (ie nothing unexpected happens after receiving routes when our ILP Address has changed)

emschwartz:
Merging this in

Issue #353, opened 2019-09-29
Ethereum settlement engine making a crazy number of blockchain calls


emschwartz:
We're currently using Infura for the ETH settlement engine and it's making a crazy number of requests without much activity.

![image](https://user-images.githubusercontent.com/3262610/65835777-491ce100-e2b8-11e9-808c-245b3a47f4c6.png)

Since it looks like this is only making calls to scan blocks for incoming transactions, we should see if we can switch to using the Websocket API instead: https://infura.io/docs/ethereum/wss/introduction.md

gakonst:
Since when is that? IIRC we had fixed the bug where too many requests were made. I can take a look at the WS API although I haven't tried using it with rust-web3 yet.

emschwartz:
Our 3 testnet nodes made 400k requests yesterday.

The original issue was that it would request every single block ever on startup. This is something different but I'm not totally sure what's causing it

gakonst:
What did they set the CLI param `--poll_frequency` to? 
It defaults to 5000ms, I've been running it for an hour so far and I haven't encountered any problem.

The only issue which MAY arise, is if they initially ran the engine on a network with a smaller block number than the one they're currently using, e.g. a ganache network with 10 blocks, then they shutdown the engine and point it to mainnet. Due to how it's done currently, the `recently_observed_block` will be set to whatever the small network's block number was, e.g. 10, and when they have the engine watch mainnet, it will see the current block number 8648962 and will try to fetch all. https://github.com/interledger-rs/interledger-rs/blob/500e731e3e095717be9e8a6c7e8efb035bd217da/crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs#L331-L346

We should add a line in the docs of the engines: "When switching networks, make sure to clear the `recently_observerd_block` value in your store". Alternatively, we should add a key which is specific to the network ID (not the chain id!) so that this never happens.

PR #362, opened 2019-09-30
Namespace block numbers

gakonst:
Otherwise we may spam our ethereum node, and we do not want that. Should fix https://github.com/interledger-rs/interledger-rs/issues/353


crates/interledger-settlement-engines/src/engines/ethereum_ledger/eth_engine.rs
```diff
@@ -100,6 +100,7 @@ pub struct EthereumLedgerSettlementEngine<S, Si, A> {
     poll_frequency: Duration,
     connector_url: Url,
     asset_scale: u8,
+    net_version: String,
```
emschwartz: Does this need to be manually configured by the user?

Also, it would be good to include a comment explaining what this does

gakonst: It's not configured by the user, the builder should be used so that it fetches it automatically

Issue #332, opened 2019-09-25
Default route


emschwartz:
Nodes should be able to configure a "default route" so that even if they do not have a specific route to a destination, they will try forwarding the packet to the default route's next hop. For child nodes, this should automatically be set to the parent. For others, we may want to allow this to be configured separately.

gakonst:
note: A node cannot know that it's a child currently

emschwartz:
It would just check if it has a parent

Issue #193, opened 2019-08-05
Reconnect database if connection drops


emschwartz:

emschwartz:
This was partially done by https://github.com/interledger-rs/interledger-rs/pull/365.

The only thing that's missing is to allow `redis::Script#invoke_async` to be called on a generic `redis::aio::ConnectionLike` instead of `SharedConnection` so we can pass the wrapped `SharedConnection` to it.

emschwartz:
Waiting for https://github.com/mitsuhiko/redis-rs/pull/242 to be merged before this can be finished

gakonst:
@emschwartz https://github.com/mitsuhiko/redis-rs/releases/tag/0.13.0 release is in

PR #365, opened 2019-09-30
Reliability improvements

emschwartz:
Resolves https://github.com/interledger-rs/interledger-rs/issues/332
Partially implements https://github.com/interledger-rs/interledger-rs/issues/193 (blocked on an upstream change that's needed to implement it fully)

### DIFF
--- a/crates/interledger-ccp/src/server.rs
+++ b/crates/interledger-ccp/src/server.rs
@@ -634,32 +634,30 @@ where
 
     /// Send RouteUpdateRequests to all peers that we send routing messages to
     fn send_route_updates(&self) -> impl Future<Item = (), Error = ()> {
-        let mut outgoing = self.outgoing.clone();
-        let to_epoch_index = self.forwarding_table.read().epoch();
-
-        let from_epoch_index: u32 = {
-            let mut lock = self.last_epoch_updates_sent_for.lock();
-            let epoch = *lock;
-            *lock = to_epoch_index;
-            epoch
-        };
-
-        let route_update_request = self.create_route_update(from_epoch_index, to_epoch_index);
-        debug!(
-            "Sending route udpates for epochs {} - {}: {:?}",
-            from_epoch_index, to_epoch_index, route_update_request,
-        );
-
-        let prepare = route_update_request.to_prepare();
+        let self_clone = self.clone();
         self.store
             .get_accounts_to_send_routes_to()
             .and_then(move |mut accounts| {
+                let mut outgoing = self_clone.outgoing.clone();
+                let to_epoch_index = self_clone.forwarding_table.read().epoch();
+
+                let from_epoch_index: u32 = {
+                    let mut lock = self_clone.last_epoch_updates_sent_for.lock();
+                    let epoch = *lock;
+                    *lock = to_epoch_index;
+                    epoch
+                };
+
+                let route_update_request = self_clone.create_route_update(from_epoch_index, to_epoch_index);
+
+                let prepare = route_update_request.to_prepare();
                 accounts.sort_unstable_by_key(|a| a.id().to_string());
                 accounts.dedup_by_key(|a| a.id());
 
                 let broadcasting = !accounts.is_empty();
                 if broadcasting {
-                    trace!("Sending route updates to accounts: {}", {
+                    trace!("Sending route update for epochs {} - {} to accounts: {:?} {}",
+                        from_epoch_index, to_epoch_index, route_update_request, {
                         let account_list: Vec<String> = accounts
                             .iter()
                             .map(|a| {
--- a/crates/interledger-store-redis/src/reconnect.rs
+++ b/crates/interledger-store-redis/src/reconnect.rs
@@ -1,0 +1,105 @@
+use futures::{
+    future::{err, result, Either},
+    Future,
+};
+use log::{debug, error};
+use parking_lot::RwLock;
+use redis::{
+    aio::{ConnectionLike, SharedConnection},
+    Client, ConnectionInfo, RedisError, Value,
+};
+use std::sync::Arc;
+
+/// Wrapper around a Redis SharedConnection that automatically
+/// attempts to reconnect to the DB if the connection is dropped
+#[derive(Clone)]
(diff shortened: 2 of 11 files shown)


emschwartz:
I guess naming your PR "reliability improvements" is just guaranteed to jinx it

emschwartz:
Note the build is currently failing because the ubuntu keyserver is offline

crates/interledger-ccp/src/server.rs
```diff
@@ -634,63 +669,109 @@ where
...
+                                    } else {
+                                        // Skip sending to this account next time
+                                        unavailable_accounts.insert(account.id(), BackoffParams {
+                                            max: 1,
+                                            skip_intervals: 1,
+                                        });
+                                    }
+
+                                    if account.routing_relation() == RoutingRelation::Child {
+                                        trace!("Error sending route update to Child account {} (id: {}), increased backoff to {}: {:?}",
```
gakonst: Backoff gets increased when we're broadcasting to a `Peer` node also. Shouldn't that case be logged too?

crates/interledger-store-redis/src/reconnect.rs
```diff
@@ -0,0 +1,105 @@
...
+        self,
+        cmd: Vec<u8>,
+    ) -> Box<dyn Future<Item = (Self, Value), Error = RedisError> + Send> {
+        let clone = self.clone();
+        Box::new(
+            self.get_shared_connection()
+                .req_packed_command(cmd)
+                .or_else(move |error| {
+                    if error.is_connection_dropped() {
+                        // TODO should the request be retried automatically?
```
gakonst: I think any retry logic should be wrapped around the function call, rather than the retry happening inside this function (ie this function looks good as is)

crates/interledger-store-redis/src/store.rs
```diff
@@ -1287,13 +1287,52 @@ impl NodeStore for RedisStore {
...
         )
     }
 
+    fn set_default_route(
+        &self,
+        account_id: AccountId,
+    ) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        let routing_table = self.routes.clone();
+        Box::new(
+            cmd("EXISTS")
```
gakonst: Could do this in a lua script to save on a db query, but I guess it's fine since this shouldn't happen too often

crates/interledger-store-redis/src/store.rs
```diff
@@ -1848,23 +1888,37 @@ type RouteVec = Vec<(String, AccountId)>;
...
+                    "Loaded routes from redis. Static routes: {:?}, default route: {:?}, other routes: {:?}",
                     static_routes,
+                    default_route,
                     routes
                 );
+                // If there is a default route set in the db,
+                // set the entry for "" in the routing table to route to that account
+                let default_route_iter = iter::once(default_route)
+                    .filter_map(|r| r)
+                    // TODO should the default route prefix be the global scheme instead of the empty string?
```
gakonst: I think empty is fine? We use an empty route as the catch-all address here: https://github.com/interledger-rs/interledger-rs/blob/9a9a54d7e80da405d9acf27f68a2963f09582b5e/crates/interledger-router/src/router.rs#L73

crates/interledger-store-redis/src/store.rs
```diff
@@ -1848,23 +1888,37 @@ type RouteVec = Vec<(String, AccountId)>;
...
+                    .map(|account_id| (String::new(), account_id));
                 let routes = HashMap::from_iter(
                     routes
                         .into_iter()
+                        // Include the default route if there is one
+                        .chain(default_route_iter)
```
gakonst: I think that the `default_route_iter` must be before the `routes` iterator? Otherwise, if there was a route in the `routes` iterator it will get overwritten by any route in the `default_route_iter`

crates/interledger-ccp/src/server.rs
```diff
@@ -248,6 +263,26 @@ where
...
         if control.mode == Mode::Sync {
+            // Don't skip them in the route update broadcasts anymore since this
+            // tells us that they are online
+            // TODO what happens if they can send to us but we can't send to them?
+            if self
+                .unavailable_accounts
+                .read()
+                .contains_key(&request.from.id())
+                // Note we take the read lock first because we want to avoid
+                // taking the write lock if it's not necessary
```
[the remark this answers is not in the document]
emschwartz: I believe so. I originally had nested if statements but clippy yelled at me

crates/interledger-ccp/src/server.rs
```diff
@@ -634,63 +669,109 @@ where
...
+            })
+            .collect();
         self.store
-            .get_accounts_to_send_routes_to()
+            .get_accounts_to_send_routes_to(unavailable)
             .and_then(move |mut accounts| {
+                let mut outgoing = self_clone.outgoing.clone();
+                let to_epoch_index = self_clone.forwarding_table.read().epoch();
+
+                let from_epoch_index: u32 = {
```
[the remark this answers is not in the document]
emschwartz: Agreed

@sentientwaffle made a good point a while ago that it may be safer to have a single RwLock on the whole struct instead of many different ones that may interact in strange ways

crates/interledger-store-redis/src/store.rs
emschwartz: `routes` shouldn't be able to overwrite the default account (we should never accept a route broadcast for the empty prefix)

crates/interledger-ccp/src/server.rs
emschwartz: Changing it to only do the backoff for Child accounts

crates/interledger-ccp/src/server.rs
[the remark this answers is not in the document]
emschwartz: Confirmed https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6c966220b6700c3ded1bf5f31a29c3a2

emschwartz:
I pulled out the CCP changes because those were the ones causing problems. I'll submit a separate PR for that

PR #367, opened 2019-10-01
Don't broadcast routes to unavailable accounts

emschwartz:
Right now the testnet node wastes a lot of resources trying to broadcast routes to all of the child accounts, many of whom are offline (because they use localtunnel to connect). This changes the behavior such that it will back off from sending route updates to child accounts when the requests fail. This does not affect the behavior for Peer accounts.

### DIFF
--- a/crates/interledger-ccp/src/server.rs
+++ b/crates/interledger-ccp/src/server.rs
@@ -108,6 +108,7 @@ where
             last_epoch_updates_sent_for: Arc::new(Mutex::new(0)),
             local_table: Arc::new(RwLock::new(RoutingTable::default())),
             incoming_tables: Arc::new(RwLock::new(HashMap::new())),
+            unavailable_accounts: Arc::new(Mutex::new(HashMap::new())),
         };
 
         #[cfg(not(test))]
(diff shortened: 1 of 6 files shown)


emschwartz:
I'm going to merge this now so it gets in before the next release. If there are any issues, we can address them in a follow-up PR

crates/interledger-ccp/src/server.rs
```diff
@@ -635,60 +667,104 @@ where
...
+                                })
+                                .then(move |res| Ok((account, res)))
+                        }))
+                        .and_then(move |results: Vec<(A, Result<Fulfill, Reject>)>| {
+                            // Handle the results of the route broadcast attempts
+                            trace!("Updating unavailable accounts");
+                            let mut unavailable_accounts = unavailable_accounts.lock();
+                            for (account, result) in results.into_iter() {
+                                match (account.routing_relation(), result) {
+                                    (RoutingRelation::Child, Err(err)) => {
```
dora-gt: Just a question, why did you select `Child`? Shouldn't we also do that for `Peer`?

gakonst: We assume that our `Peer` nodes are online most of the time, and it's critical that they always have the latest routes, so we prefer not to do any backoff for them

crates/interledger-ccp/src/server.rs
```diff
@@ -635,60 +667,104 @@ where
...
+                                        if let Some(backoff) = unavailable_accounts.get_mut(&account.id()) {
+                                            // Increase the number of intervals we'll skip
+                                            // (but don't overflow the value it's stored in)
+                                            backoff.max = backoff.max.saturating_add(1);
+                                            backoff.skip_intervals = backoff.max;
```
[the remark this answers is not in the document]
emschwartz: Right, I guess we could have made it exponential. If you submit a PR to change it, I'd be fine with that

PR #368, opened 2019-10-02
feat(ilp-cli): incoming-payments subcommand

bstrie:
I've been awake for 40 hours and have ascended beyond earthly concerns such as code quality

### DIFF
--- a/crates/ilp-cli/Cargo.toml
+++ b/crates/ilp-cli/Cargo.toml
@@ -12,3 +12,5 @@ http = "0.1.18"
 reqwest = { version = "0.9.20", default-features = false, features = ["default-tls"] }
 serde = "1.0.101"
 serde_json = "1.0.39"
+tungstenite = "0.9.1"
+url = "2.1.0"
--- a/crates/ilp-cli/src/interpreter.rs
+++ b/crates/ilp-cli/src/interpreter.rs
@@ -2,7 +2,9 @@ use clap::ArgMatches;
 use http;
 use reqwest::{self, Client, Response};
 use serde_json::value::*;
-use std::collections::HashMap;
+use std::{borrow::Cow, collections::HashMap};
+use tungstenite::{connect, handshake::client::Request};
+use url::Url;
 
 #[derive(Debug)]
 pub enum Error {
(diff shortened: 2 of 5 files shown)


crates/ilp-cli/src/interpreter.rs
```diff
@@ -127,8 +132,32 @@ impl NodeClient<'_> {
             .map_err(Error::ClientErr)
     }
 
-    fn ws_account_payments_incoming(&self, _matches: &ArgMatches) -> Result<Response, Error> {
-        unimplemented!()
+    fn ws_account_payments_incoming(&self, matches: &ArgMatches) -> Result<Response, Error> {
+        let (auth, args) = extract_args(matches);
+        let mut headers = Headers::new();
+        headers.set(Authorization(Bearer {
```
emschwartz: Instead of manually setting the Authorization header, you can just include the auth details in the URL you're connecting to `ws://username:password@domain.example` or `ws://:bearer-token@domain.example`

crates/ilp-cli/Cargo.toml
```diff
@@ -9,4 +9,5 @@ repository = "https://github.com/interledger-rs/interledger-rs"
 [dependencies]
 clap = { version = "2.33.0", default-features = false }
 reqwest = { version = "0.9.20", default-features = false, features = ["default-tls"] }
-serde = "1.0.101"
\ No newline at end of file
+serde = "1.0.101"
+websocket = "0.23.0"
```
emschwartz: We already have so many dependencies, I would much prefer to use tungstenite or tokio-tungstenite because we already depend on them

crates/ilp-cli/src/interpreter.rs
```diff
@@ -129,10 +135,33 @@ impl NodeClient<'_> {
...
-        unimplemented!()
+    // WebSocket /accounts/:username/payments/incoming
+    fn ws_account_payments_incoming(&self, matches: &ArgMatches) -> Result<Response, Error> {
+        let (auth, args) = extract_args(matches);
+        let mut url = Url::parse(&format!(
+            "{}/accounts/{}/payments/incoming",
+            self.url, args["username"]
+        ))
+        .expect("Could not parse URL");
+        // TODO: wss if node url is https?
```
emschwartz: Yes

PR #371, opened 2019-10-02
Testnet docker bundle

emschwartz:


crates/ilp-cli/src/parser.rs
```diff
@@ -395,13 +395,16 @@ fn testnet<'a, 'b>() -> App<'a, 'b> {
...
+        .args(&[
             Arg::with_name("asset")
                 .index(1)
                 .required(true)
                 .takes_value(true)
                 .possible_values(&["xrp", "eth"])
                 .case_insensitive(true)
                 .help("The asset that will be tied to the new testnet account"),
-        )
+            Arg::with_name("return_testnet_credential")
```
dora-gt: _ and - look inconsistent, is that OK?

docker/run-testnet-bundle.js
```diff
@@ -0,0 +1,194 @@
...
+        const config = JSON.parse(fs.readFileSync(configFilePath))
+        config.firstRun = false
+        return config
+    } else {
+        const nodeName = process.env.NAME || `ilp_node_${randomBytes(10).toString('hex')}`
+        const adminAuthToken = process.env.ADMIN_AUTH_TOKEN || `admin-token-${randomBytes(20).toString('hex')}`
+        const secretSeed = randomBytes(32).toString('hex')
+        const currency = (process.env.CURRENCY || 'XRP').toUpperCase()
+        const ethKey = process.env.ETH_SECRET_KEY || '380EB0F3D505F087E438ECA80BC4DF9A7FAA24F868E69FC0440261A0FC0567DC'
+        const ethUrl = process.env.ETH_URL || 'https://rinkeby.infura.io/v3/2bd7fa2c387e4f07a5599bc64d0a9c33'
```
dora-gt: Where is this pointed? just asking. Xpring one?

emschwartz: Yes, for now

crates/ilp-cli/src/parser.rs
emschwartz: All of the clap args are defined with underscores for the internal names and dashes for the parameters the user will set

PR #373, opened 2019-10-02
fix(interpreter.rs): send numbers in JSON

dora-gt:
We must fix this in the future, just for the hackathon.

### DIFF
--- a/crates/ilp-cli/src/interpreter.rs
+++ b/crates/ilp-cli/src/interpreter.rs
@@ -2,6 +2,7 @@ use clap::ArgMatches;
 use http;
 use reqwest::{self, Client, Response};
 use std::collections::HashMap;
+use serde_json::value::*;
 
 #[derive(Debug)]
 pub enum Error {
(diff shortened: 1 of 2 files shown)


crates/ilp-cli/src/interpreter.rs
```diff
@@ -88,6 +88,7 @@ struct NodeClient<'a> {
     url: &'a str,
 }
 
+use serde_json::value::*;
```
gakonst: this should be the top of the file

crates/ilp-cli/src/interpreter.rs
```diff
@@ -153,10 +154,21 @@ impl NodeClient<'_> {
     fn put_account_settings(&self, matches: &ArgMatches) -> Result<Response, Error> {
         let (auth, mut args) = extract_args(matches);
         let user = args.remove("username").unwrap();
+
+        // FIXME
```
emschwartz:
```suggestion
        // This is done to ensure that the values that need to be numbers
        // are sent as numbers, because otherwise the node rejects
        // the request with a (strange) 401: Unauthorized error
        // TODO: find a less hacky way to send the request
```

emschwartz:
Failing test is unrelated, merging anyway

PR #375, opened 2019-10-02
Testnet bundle improvements

emschwartz:


docker/run-testnet-bundle.js
```diff
@@ -158,7 +165,8 @@ function runEthSettlementEngine({ ethKey, ethUrl }) {
         '--settlement_api_bind_address=127.0.0.1:3002',
         `--ethereum_url=${ethUrl}`,
         `--private_key=${ethKey}`,
-        `--poll_frequency=15000`,
+        '--poll_frequency=15000',
+        '--confirmations=0',
```
kincaidoneil: Should this be 1?

emschwartz: Why?

gakonst: Sorry @kincaidoneil this is not 0-conf, this is just 0-blocks built on top of the fetched block

PR #379, opened 2019-10-02
Make minor edits to the testnet doc

pgarg-ripple:


docs/testnet.md
```diff
@@ -8,19 +8,19 @@
...
-    - Running a node and peering
+- Manual Configuration
+    - Requirements
+    - Running a Node and Peering
 
 
 ## Get Your Own Credentials
 
-In order to connect to the testnet, you need to generate
+To connect to the testnet, you need to generate
```
dora-gt: @pgarg-ripple May I just ask? What is the difference between `in order to` and `to`?
Because I'm not a native speaker, I can't understand why we should write `to` in this case.

dora-gt: Sounds like too formal?

emschwartz: They both sound pretty similar to me actually

pgarg-ripple: @dora-gt ,
Yes, they are very similar, but in a technical documentation we must avoid verbosity. For example:

- In order to -> To
- Due to the fact that -> Because
- At this point in time -> Now; Currently
- For the purpose of -> To

PR #382, opened 2019-10-03
Proper fix for allowing ilp-cli to send strings to the node for /rates

bstrie:


gakonst:
Could you change this line to use a String as well, to make sure we test for this in our e2e tests? https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-settlement-engines/tests/eth_xrp_interoperable.rs#L126

bstrie:
Can anyone reproduce the failure? All works on my end.

PR #385, opened 2019-10-03
feat(api): include crate version in status response

emschwartz:
note this will return the version of the interledger-api crate, rather than the ilp-node crate

### DIFF
--- a/crates/interledger-api/src/routes/node_settings.rs
+++ b/crates/interledger-api/src/routes/node_settings.rs
@@ -48,15 +48,24 @@ where
         .boxed();
     let with_store = warp::any().map(move || store.clone()).boxed();
 
+    let crate_version = format!(
+        "{}.{}.{}{}",
+        env!("CARGO_PKG_VERSION_MAJOR"),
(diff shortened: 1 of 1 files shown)


gakonst:
@emschwartz what do you think of @dora-gt comments?

emschwartz:
I'd be fine adding a config parameter later along the lines of what @dora-gt, if that's the recommended security practice. In general though, I'm somewhat skeptical of how much security it really adds not to publish the version. If there is a vulnerable version, people should update ASAP because this is software that deals with money. Whether or not you publicize the version, people will try exploiting all public endpoints they can find

tarcieri:
I'd say it's exceedingly common for severs to report the versions of their server software, which is useful for things like network telemetry "for good".

In the event it's not reported, an attacker can always blindly attempt the attack. Perhaps the one disadvantage to the attacker by not reporting it is if they do this, the attack is less stealthy than if they just target specific versions.

dora-gt:
I meant that the attackers could "easily" find the targets.
I agree that attackers could attack even if we hide the information.

PR #392, opened 2019-10-07
Interface test for `ilp-cli testnet`

bstrie:


bstrie:
Test failure looks spurious. I'd like to consider either no longer gating our CI on that test, or no longer gating PR merges on CI, or on granting the other repo contributors the ability to bypass the CI requirement for merging, to keep this from randomly halting momentum in the future.

PR #394, opened 2019-10-08
chore: move engines to own repo

### LINKED ISSUE #343: Move the interledger-settlement-engines to its own repo
Should we:
- Put each settlement engine in a separate crate (to avoid having one binary and using conditional compilation to avoid pulling in dependencies when they aren't needed)
- Move the settlement engines to a separate repo under `interledger-rs` (but keep multiple engines in the same repo so they can easily share the core code)
- Move the settlement engines to a repo or organization where engines written in other languages will also live

cc @gakonst

gakonst:
as title. Closes #343


.circleci/config.yml
```diff
@@ -66,7 +48,7 @@ jobs:
             cargo clippy --all-targets --all-features -- -D warnings
       - run:
           name: Audit Dependencies
-          command: LD_PRELOAD=/usr/lib/libcurl.so.3 cargo audit
```
bstrie: Are the libcurl3 changes related to this PR, or is this just regular maintenance?

gakonst: The libcurl change was initially introduced because `libcurl4` would be installed when we'd try to install `curl` to download `nvm`, which replaced `libcurl3`. Now that we de not have any `node` components, that is no longer needed

Issue #341, opened 2019-09-27
Investigate whether using the CircleCI cache speeds up or slows down builds


emschwartz:
Restoring and saving the cache on CircleCI seems to take >2 minutes. We should check whether removing the cache altogether speeds up the builds.

koivunej:
CircleCI is no longer used, closing this one.

PR #398, opened 2019-10-08
Speed up CircleCI builds

emschwartz:
- Don't use cargo cache (creating and loading the cache is slower than building from scratch)
- Remove unnecessary libcurl3 dep
- Split style checking from main build and test task to run them in parallel

Resolves https://github.com/interledger-rs/interledger-rs/issues/341

Note: I tried switching to a `large` box and removing the command to reduce codegen units but it only shaved 40 seconds off the build time so it doesn't seem worth it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,11 +1,10 @@
 version: 2
 jobs:
-  build:
+  style-and-audit:
     docker:
       - image: circleci/rust
         environment:
           CARGO_HOME: /home/circleci/.cargo
-    resource_class: medium+
     steps:
       - checkout
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,34 +1,38 @@
...
       - run:
           name: Install Cargo Extensions
           command: |
-            # cargo-audit started requiring libcurl3
-            echo "deb http://security.ubuntu.com/ubuntu xenial-security main" | sudo tee -a /etc/apt/sources.list
-            sudo apt-key adv --recv-keys --keyserver keyserver.ubuntu.com 3B4FE6ACC0B21F32
-            sudo apt-get update
-            sudo apt-get install libcurl3 -y
             cargo install --quiet cargo-audit || true # cargo-kcov
             rustup component add rustfmt clippy || true
```
dora-gt: Though this is not related to your change, I'm wondering why we are using `|| true` here.
We always need this, don't we?

emschwartz: Those commands fail if the component is already installed so the `|| true` makes sure it doesn't fail the build step

emschwartz: @tarcieri posted on the rust-lang internals forum about a cargo feature that would make this unnecessary: https://internals.rust-lang.org/t/idea-cargo-install-update/11072

emschwartz:
> Doesn't seem like splitting the clippy job to a separate workflow provides any significant gains, given that it has to spend a lot of time building.

@gakonst part of the goal here was to remove the cache of cargo deps because we're stuck on old versions of those external tools. If we do that, installing clippy and such takes 2 min (see [here](https://circleci.com/gh/interledger-rs/interledger-rs/2288)), so I didn't want that to block building and testing the project. In the split version, running rustfmt and clippy takes [1:40](https://circleci.com/gh/interledger-rs/interledger-rs/2288) whereas in the combined version it takes [0:52](https://circleci.com/gh/interledger-rs/interledger-rs/2366). Since the style and build/test tasks take around the same amount of time, I don't think it's so much of a problem for the format part to take slightly longer.

We might actually get the best speedup if we could figure out a way to cache only the built version of clippy and the other external tools we use (not all the subdependencies), use cargo install-update to make sure we're on the latest version, and only update the cache if the versions of the tools we're using changed rather than saving it on every build.

gakonst:
No strong opinions this is better than the current situation and looks like we won't be going with github actions, approved

emschwartz:
I'm going to close this because I'm not sure it actually speeds things up

Issue #292, opened 2019-09-09
Try out Github actions


emschwartz:
To see how it compares with CircleCI

emschwartz:
Also see https://svartalf.info/posts/2019-09-16-github-actions-for-rust/

gakonst:
Closing this, given that we tried it in #399 and chose to stick with our current setup

PR #399, opened 2019-10-08
Switch to Github Actions

emschwartz:
This switches the project to use Github actions instead of CircleCI and is an alternative to https://github.com/interledger-rs/interledger-rs/pull/398.

Advantages of Github Actions:
- Very nice integration with Github to show problems without looking at terminal logs
- Ability to import actions created by others
- Free

Advantages of Circle:
- Build takes 4-5 min instead of 8-11(note Github Actions is [planning](https://github.com/interledger-rs/interledger-rs/pull/398) to support caching in November)
- Allows SSH access to the machines it's building on to help debug issues

Resolves https://github.com/interledger-rs/interledger-rs/issues/292

What do you think?


.github/workflows/audit.yml
```diff
@@ -0,0 +1,27 @@
...
+    paths-ignore:
+      - "docs/**/*"
+      - "scripts/**/*"
+      - "README.md"
+
+name: Audit Dependencies
+
+jobs:
+  audit:
+    runs-on: ubuntu-latest
```
bstrie: What are the options here? Is this the platform that we encourage people to use?

.github/workflows/build.yml
```diff
@@ -0,0 +1,22 @@
+name: Build and Test
```
bstrie: Is there a semantic reason why the name is at the top for this one?

crates/interledger-ccp/src/server.rs
```diff
@@ -933,25 +933,23 @@ fn get_best_route_for_prefix<A: CcpRoutingAccount>(
...
+                        .cmp(&account.routing_relation()),
+                    best_route.path.len().cmp(&route.path.len()),
+                    best_account.id().to_string().cmp(&account.id().to_string()),
+                ) {
+                    (Ordering::Greater, _, _)
+                    | (Ordering::Equal, Ordering::Less, _)
+                    | (Ordering::Equal, Ordering::Equal, Ordering::Less) => {
+                        (best_account, best_route)
+                    }
+                    _ => (account, route),
```
bstrie: Are the changes in this file from an unrelated commit?

emschwartz: https://github.com/interledger-rs/interledger-rs/pull/399/commits/440bba33dcdabcddabda8c7d43264876410838de

.github/workflows/audit.yml
emschwartz: Virtual environment | YAML workflow label
-- | --
Windows Server 2019 | windows-latest or windows-2019
Windows Server 2016 R2 | windows-2016
Ubuntu 18.04 | ubuntu-latest or ubuntu-18.04
Ubuntu 16.04 | ubuntu-16.04
macOS X Mojave 10.14 | macOS-latest or macOS-10.14

[Source](https://help.github.com/en/articles/virtual-environments-for-github-actions#supported-virtual-environments-and-hardware-resources)

Not necessarily, but it's what we were building on Circle already. If we want to support others, we can add steps to build it on other platforms too (as a separate PR)

.github/workflows/audit.yml
```diff
@@ -0,0 +1,27 @@
+on:
+  pull_request:
```
gakonst: Shouldn't `cargo audit` be run `on: push`, instead of `on: pull_request`?

.github/workflows/style.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          command: fmt
+          args: --all -- --check
+
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - id: component
+        uses: actions-rs/components-nightly@v1
```
gakonst: Should we be using nightly here?

gakonst:
image

Wonder why `Build and Test` appears twice, once with "build" and once with "clippy" next to it?

emschwartz:
Since this is quite a bit slower than Circle, I think we should hold off on this switch for now. We can revisit this later when they add caching

Issue #140, opened 2019-07-18
Consider review of dependencies


bstrie:
Unstructured musings:

* the parking_lot crate is potentially being integrated into libstd (though that will take a while and I don't want to guess at an ETA), though if we're not certain that we need parking_lot for speed then we could benchmark it and see if it's necessary for our use case
* our remaining uses of tower-web should be switched out for warp
* in a recent release some functionality of the failure crate was upstreamed into libstd, so we may be able to reduce the need for that dependency
* several of our crates are depending on chrono, which may be possible to jettison depending on what functionality specifically we're using
* regarding the rand crate, I'm curious what functionality of rand that we require and whether we're using random numbers in a context where the quality of the PRNG matters
* I want to investigate what the regex crate is being used for and see whether a non-regex solution might be preferable for those
* I want to comb through some of the packages I've never heard of before and make sure they're still maintained
* likewise if we have packages pulling in different versions of the same transitive dependency then it might be useful to submit PRs to projects to see if we can get them all to upgrade to the same version
* quick-error in interledger-packet can be removed by refactoring the Error type

emschwartz:
> the futures crate might often be able to be replaced by the recently stabilized std::future module, depending on how much functionality is used

Definitely worth investigating. Want to add that to your queue?

> the parking_lot crate is potentially being integrated into libstd (though that will take a while and I don't want to guess at an ETA), though if we're not certain that we need parking_lot for speed then we could benchmark it and see if it's necessary for our use case

The other advantage is that it doesn't return a `Result` so you don't need to have `.unwrap` all over the place

> we've got a lot of different crates doing http like reqwest and hyper and tower-web, so depending on our needs we may want to consolidate here

`reqwest` and `tower-web` both use `hyper` so we should just make sure we have consistent versions of the first two and that they're pulling in the same version of `hyper`

> in a recent release some functionality of the failure crate was upstreamed into libstd, so we may be able to reduce the need for that dependency

:+1: 

> several of our crates are depending on chrono, which may be possible to jettison depending on what functionality specifically we're using

Worth looking into. Right now I believe no crate exposes `chrono` types in its API but it is useful for parsing timestamps from ILP packets (not sure how much code it would be to replicate the specific behavior we need)

> regarding the rand crate, I'm curious what functionality of rand that we require and whether we're using random numbers in a context where the quality of the PRNG matters

[...]

PR #400, opened 2019-10-08
Prune dependencies

### LINKED ISSUE #387: Consolidate transitive dependencies in our crate graph
Currently a full compile will pull in several different versions of the same dependency, which should be unnecessary. Subset of #140.

bstrie:
This series of commits fixes #387 and addresses the lowest-hanging of #140. In practical terms this causes `cargo build` to now build 291 packages, down from 329 (down to 358 from 410 for `cargo test`), saving ten seconds overall for clean builds and shaving 3MB off of the `ilp-node` binary.


bstrie:
Can't reproduce the test failure locally, does that look spurious to you?

crates/ilp-node/src/main.rs
```diff
@@ -151,8 +151,8 @@ fn merge_std_in(config: &mut Config) {
         if let Ok(buf_str) = String::from_utf8(buf) {
             let config_hash = FileFormat::Json
                 .parse(None, &buf_str)
-                .or_else(|_| FileFormat::Yaml.parse(None, &buf_str))
-                .or_else(|_| FileFormat::Toml.parse(None, &buf_str))
+                //.or_else(|_| FileFormat::Yaml.parse(None, &buf_str))
```
emschwartz: The features for `config` just need to be turned on to support this

crates/ilp-node/src/node.rs
```diff
@@ -343,28 +343,6 @@ impl InterledgerNode {
     pub fn run(&self) {
         tokio_run(self.serve());
     }
-
-    #[doc(hidden)]
-    #[allow(dead_code)]
-    pub fn insert_account(
```
emschwartz: This is still used by the tests

crates/ilp-node/tests/exchange_rates.rs
```diff
@@ -13,7 +13,8 @@ use tokio::runtime::Builder as RuntimeBuilder;
 mod redis_helpers;
 use redis_helpers::*;
 
-#[test]
+// TODO: reenable this test once it's no longer flaky
```
emschwartz: What's the problem you've been having with this? It works fine for me

crates/interledger-packet/Cargo.toml
```diff
@@ -8,22 +8,21 @@ edition = "2018"
...
-hex = "0.3.2"
-quick-error = "1.2.2"
-serde = { version = "1.0.99", features = ["derive"], optional = true }
-regex = "1.1.6"
-lazy_static = "1.3"
+byteorder = { version = "1.3.2", default-features = false }
+bytes = { version = "0.4.12", default-features = false, features = ["serde"] }
+chrono = { version = "0.4.9", default-features = false }
+hex = { version = "0.4.0", default-features = false }
+quick-error = { version = "1.2.2", default-features = false }
```
emschwartz: We can also get rid of this dependency by refactoring the Error type (that can be done in a separate PR)

crates/interledger/Cargo.toml
```diff
@@ -17,7 +18,6 @@ default = [
     "router",
     "service-util",
     "settlement",
-    "spsp",
```
emschwartz: This should be needed

crates/ilp-node/src/main.rs
bstrie: Indeed, the reasons these are commented out is just to give you the explicit option of considering whether we want to support all of JSON, YAML, and TOML rather than just one (where JSON is the natural choice here, IMO). I can add these back if you value the flexibility that offering more configuration formats provides.

crates/ilp-node/src/node.rs
bstrie: I ran `cargo test` in every individual crate's subdirectory, as well as `cargo test --all --all-features`, and ran into no failures as a result of this function's deletion. What would I need to do to exercise it?

crates/interledger-packet/Cargo.toml
bstrie: Indeed, there are plenty of dependencies that are candidates for removal (tower-web foremost among them), I'll make note in #140 of ones like this as I come across them.

crates/interledger/Cargo.toml
bstrie: It did seem curious, but `ilp-node` doesn't appear to actually be using anything that would require this. All tests pass without it, including the simple example (I didn't test the non-simple examples because they were broken by the extraction of the settlement engines).

crates/ilp-node/src/main.rs
emschwartz: Yes, please add them back

crates/ilp-node/src/node.rs
emschwartz: It looks like the tests that still use it were moved to https://github.com/interledger-rs/settlement-engines/tree/master/tests

crates/ilp-node/tests/exchange_rates.rs
bstrie: Amusingly it was failing for me 100% of the time yesterday, but now it works just fine.

crates/interledger/Cargo.toml
emschwartz: I think it's because the API already includes the SPSP functionality it's using

crates/ilp-node/tests/exchange_rates.rs
emschwartz: This TODO should be removed

Issue #404, opened 2019-10-09
Fix the examples so they can be run from any directory


emschwartz:
You should be able to run `./scripts/run-md.sh examples/simple` from the root of the project as well as `../../scripts/run-md.sh README.md` from the directory of one of the examples.

Instead of using a symlink with a relative path, we should investigate using bash aliases to make the command look nicer than calling `cargo run -p ilp-cli ...` everywhere.

PR #405, opened 2019-10-09
fix: make simple example work when run from any directory

bstrie:
Resolves https://github.com/interledger-rs/interledger-rs/issues/404

### DIFF
--- a/examples/eth-settlement/README.md
+++ b/examples/eth-settlement/README.md
@@ -100,8 +100,11 @@ else
     done
 fi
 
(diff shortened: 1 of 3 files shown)


bstrie:
Absolutely, just wanted to make sure you were fine with this direction. Although note that I won't be able to successfully test the remaining examples until they're updated for the settlement-engines move, I don't think this change poses much risk.

bstrie:
eth and xrp examples updated, the last one I'm saving for a separate commit because I'm only just now realizing I never updated it to use ilp-cli.

PR #406, opened 2019-10-10
fix(example): fix docker mode of simple example

dora-gt:
Docker mode of simple example is not working because of multiple problems.

1. `node.dockerfile` is wrong, it should not specify `node` for arguments.
1. Checking balance commands are not working, `ilp-cli` command is not found on Docker mode.
    - Specifying `ilp-cli` in Docker mode looked inefficient because we have to compile.
    - Using `curl` might be another option but using Docker image of `ilp-cli` looked suitable also.

Because `ilp-cli` needs `--rm` option, this might be not so efficient. Needs to be fixed in the future.

### DIFF
--- a/docker/node.dockerfile
+++ b/docker/node.dockerfile
@@ -34,4 +34,3 @@ WORKDIR /opt/app
 ENV RUST_LOG=ilp,interledger=debug
 
 ENTRYPOINT [ "/usr/local/bin/ilp-node" ]
-CMD [ "node" ]
--- a/examples/simple/README.md
+++ b/examples/simple/README.md
@@ -337,18 +337,32 @@ fi
 <!--!
 # check balances before payment
 printf "\n\nChecking balances prior to payment...\n"
-printf "\nAlice's balance: "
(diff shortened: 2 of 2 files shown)


dora-gt:
I think the container cannot be reused because we have specified some arguments when `run` and cannot override the arguments when `start` it. Am I wrong? I tried a few times but I could not make it work.

emschwartz:
What's the exact problem? When I run different commands with `docker run -it interledgerrs/ilp-cli` it seems to work fine

dora-gt:
Docker `run` always starts with a NEW container and it still remains after the run. So if we `run` many times, the garbages would be piled up. Then, if we use `start`, the arguments cannot be overridden. That's the problem.

emschwartz:
Ah you're right. I was only looking at the running containers

dora-gt:
OK, I'm merging this.

emschwartz:
Sorry about that, I changed the CI check to the split checks added in one of my PRs. Changed the branch protection rule back for now

PR #411, opened 2019-10-11
feat(testnet docker bundle): accepts http_bind_address from env and defaults to `0.0.0.0:7770`

dora-gt:
Now that the node is bound to `0.0.0.0`, we can bind localhost's 7770 to the container's 7770.

Also includes changes for the settlement engine repo change.


dora-gt:
Merged and I've updated the image.

PR #412, opened 2019-10-11
fix(btp, cargo.toml): enable TLS

dora-gt:
TLS is not enabled for BTP as reported [here](https://interledger.slack.com/archives/CHC51E54J/p1570774735020000?thread_ts=1570729525.012400&cid=CHC51E54J), this PR fixes it.


dora-gt:
Updated Docker image.

Issue #407, opened 2019-10-10
Any plan to update `ring` dependency to latest version?


calvinaco:
The latest version of `ring` is [0.16.9](https://docs.rs/ring/0.16.9/ring/) but this repository and the `settlement-engines` repository is still using `0.14.6`.

This causes a problem when I try to implement a settlement engine by including my own project as a crate in `Cargo.toml`.

```
error: failed to select a version for `ring`.
    ... required by package `rustls v0.16.0`
    ... which is depended on by `{my-project}`
    ... which is depended on by `interledger-settlement-engines v0.1.1-beta.2`
versions that meet the requirements `^0.16.5` are: 0.16.9, 0.16.7, 0.16.6, 0.16.5

the package `ring` links to the native library `ring-asm`, but it conflicts with a previous package which links to `ring-asm` as well:
package `ring v0.14.6`
    ... which is depended on by `ilp-node v0.4.1-beta.2`
    ... which is depended on by `interledger-settlement-engines v0.1.1-beta.2`
    ...
```

I understand this update is optional to this repository but I would like to know if there is any plan on doing so? Thank you.

emschwartz:
For sure! The main issue is that there was a breaking change to the API so it's a bit of a pain to update. (PRs welcome if you want to expedite this! :wink:)

PR #413, opened 2019-10-11
Upgrade ring to 0.16.9

calvinlauco:
As explained in https://github.com/interledger-rs/interledger-rs/issues/407, the `ring` dependency latest version is `0.16.9`, which involves breaking changes in method signatures with the `0.14.6` version `interledger-rs` is using.

This PR
- upgrades the `ring` dependency to `0.16.9` and
- migrates existing code to adopt the new method signature.

I am quite new to Rust so please let me know if anything can be improved. I am more than happy to try to change it to keep up with the standard here.

### DIFF
--- a/crates/interledger-store-redis/src/account.rs
+++ b/crates/interledger-store-redis/src/account.rs
@@ -154,7 +154,7 @@ impl Account {
 
     pub fn encrypt_tokens(
         mut self,
-        encryption_key: &aead::SealingKey,
+        encryption_key: &aead::LessSafeKey,
     ) -> AccountWithEncryptedTokens {
         if let Some(ref token) = self.ilp_over_btp_outgoing_token {
             self.ilp_over_btp_outgoing_token = Some(SecretBytes::from(encrypt_token(
@@ -190,7 +190,7 @@ pub struct AccountWithEncryptedTokens {
 }
 
 impl AccountWithEncryptedTokens {
-    pub fn decrypt_tokens(mut self, decryption_key: &aead::OpeningKey) -> Account {
+    pub fn decrypt_tokens(mut self, decryption_key: &aead::LessSafeKey) -> Account {
         if let Some(ref encrypted) = self.account.ilp_over_btp_outgoing_token {
             self.account.ilp_over_btp_outgoing_token =
                 decrypt_token(decryption_key, &encrypted.expose_secret()).map(SecretBytes::from);
--- a/crates/interledger-store-redis/src/crypto.rs
+++ b/crates/interledger-store-redis/src/crypto.rs
@@ -1,7 +1,7 @@
 use bytes::Bytes;
 use log::error;
 use ring::{
-    aead, digest, hmac,
+    aead, hmac,
     rand::{SecureRandom, SystemRandom},
 };
 
@@ -14,13 +14,13 @@ use std::ptr;
 use zeroize::Zeroize;
 
 #[derive(Debug)]
-pub struct EncryptionKey(pub(crate) aead::SealingKey);
+pub struct EncryptionKey(pub(crate) aead::LessSafeKey);
 
 #[derive(Debug)]
-pub struct DecryptionKey(pub(crate) aead::OpeningKey);
+pub struct DecryptionKey(pub(crate) aead::LessSafeKey);
 
 #[derive(Debug)]
-pub struct GenerationKey(pub(crate) hmac::SigningKey);
+pub struct GenerationKey(pub(crate) hmac::Key);
 
 impl DebugSecret for EncryptionKey {}
 impl DebugSecret for DecryptionKey {}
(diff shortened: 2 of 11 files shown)


calvinlauco:
Yes, sure. Sorry I have to confess I forgot to take a look on the contribution guideline first.

I am away from my laptop now. I will follow the guide in the doc and update this PR tomorrow, thanks!

crates/interledger-store-redis/src/account.rs
```diff
@@ -154,7 +154,7 @@ impl Account {
 
     pub fn encrypt_tokens(
         mut self,
-        encryption_key: &aead::SealingKey,
+        encryption_key: &aead::LessSafeKey,
```
gakonst: Why change to `LessSafeKey`? The [docs](https://docs.rs/ring/0.16.9/ring/aead/struct.LessSafeKey.html) say:

> Prefer to use OpeningKey/SealingKey and NonceSequence when practical.

crates/interledger-store-redis/src/crypto.rs
```diff
@@ -95,51 +99,44 @@ fn atomic_fence() {
...
-        aead::Aad::from(&[]),
-        &mut token,
-        encryption_key.algorithm().tag_len(),
-    ) {
-        token.split_off(out_len);
-        token.append(&mut nonce_copy.as_ref().to_vec());
-        Bytes::from(token)
-    } else {
-        panic!("Unable to encrypt token");
+    match encryption_key.seal_in_place_append_tag(nonce, aead::Aad::from(&[]), &mut token) {
```
gakonst: Should we be panicking here? Maybe `encrypt_token` should return a `Result` instead, and return `Err` instead of panicking

Also, are we able to skip the `split_off` call because `seal_in_place_append_tag` does that internally now?

crates/interledger-store-redis/src/account.rs
calvinlauco: Thanks for the review! I am not a cryptography expert but these are my understandings of the situation and hence my decisions so far:

So my first intention is also to use OpeningKey/SealingKey, but these two key structures are now bound to a nonce sequence and have to be encrypted/decrypted using the same nonce sequence. So it is not exactly the current behaviour we are using: which is to use a random nonce and append the nonce after the ciphertext.

But as the docs mention it is preferred to use OpeningKey/SealingKey instead, to make this possible we will have to implement a nonce sequence and advancement algorithm (I think there should be some good advancement strategy out there? I need to do some research) and we should manage the same nonce sequence between each encrypt-er and decrypt-er pairs?

I think given the complexity and things have to be decided it is better to move this into separate tasks and keep this PR as simple as keeping the current behaviour. I am also willing to continue this follow-up task and we can discuss a bit more on the possibilities.

crates/interledger-store-redis/src/crypto.rs
calvinlauco: I agree, I think `panic` here seems to be a bit too severe. I haven't read through the whole repo yet but I think returning a Result and handle it upstream is generally a good practice. But again I would assume this deserve a new issue & PR?

calvinlauco: I agree, I think `panic` here seems to be a bit too severe. I haven't read through the whole repo yet but I think returning a Result and handle it upstream is generally a good practice. Any other concerns?

calvinlauco: Regarding `seal_in_place_append_tag `, yes right now the function
> Encrypts and signs (“seals”) data in place, appending the tag to the resulting ciphertext.

So you don’t have to allocate extra space in the bytes vector for appending the tag in place and `split_off`  those didn’t get overwritten.

Btw, since we are talking,about the `split_off`, One think I want to make sure is: isit just an extra safety check in the original code? Since it is using AES256 all the time, the tag length is always 16 bytes, so the returned length in the old version of `seal_in_place` will always be the original plaintext length + tag length in bytes, and `split_off` always split nothing off. The code will be useful only when the encryption algorithm is changed?

emschwartz: Re: panicking -- it looks like the only reason that function returns an error is if you pass in more data than the algorithm is meant to take.

> Since it is using AES256 all the time, the tag length is always 16 bytes, so the returned length in the old version of seal_in_place will always be the original plaintext length + tag length in bytes, and split_off always split nothing off. The code will be useful only when the encryption algorithm is changed?

That seems correct, though I don't remember the logic behind the original code (and this is why it's important to leave comments in important bits of code like this :man_facepalming:)

emschwartz:
@gakonst @tarcieri if you guys don't see any major issues with this PR, I can rebase and merge it (after fixing the conflicts) tomorrow

crates/interledger-store-redis/src/account.rs
bstrie: Agreed that this can be left for a separate PR, so that this one can proceed. I've filed https://github.com/interledger-rs/interledger-rs/issues/460 to track it.

tarcieri: I think `ring` is using a confusing name here trying to get across the point that manual nonce management is fraught with peril. That said, this usage is, at least, no worse than what you were doing before.

emschwartz:
Thanks @calvinlauco for this PR!

PR #414, opened 2019-10-12
docs: fix engine docs

gakonst:


README.md
```diff
@@ -48,7 +48,7 @@ docker run -it interledgerrs/node
 # This is a simple CLI for interacting with the node's HTTP API
 docker run -it interledgerrs/ilp-cli
```
dora-gt:
```suggestion
docker run -it --rm interledgerrs/ilp-cli
```

because the container remains after the execution and the container cannot be reused.
(though this is not related to your change)

PR #420, opened 2019-10-15
Switch Settlement API to Warp

gakonst:
As title. Depends on #324. Idempotency functions were extracted under `interledger_http::idempotency`, so that they can be shared with the engines. I'd also like to extract the `LeftoversStore` trait to a package that I can import in the engines for the same reason.

The diff is actually even smaller than what it seems to be, check commits https://github.com/interledger-rs/interledger-rs/pull/420/commits/1fbbf5e6648fed6b67c3848991a091d5528f13dd, https://github.com/interledger-rs/interledger-rs/pull/420/commits/1b2f57b02ab464711d52aeedcd68226317009bb0 and https://github.com/interledger-rs/interledger-rs/pull/420/commits/fb8681712ed5d4b6d945f006d547f46c3e2cc5ff.

I think I prefer Warp to tower_web, although it does take a while to figure out how to manage routes.


crates/interledger-service-util/Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository = "https://github.com/interledger-rs/interledger-rs"
 [dependencies]
 bytes = { version = "0.4.12", default-features = false }
 byteorder = { version = "1.3.2", default-features = false }
-chrono = { version = "0.4.9", default-features = false }
+chrono = { version = "0.4.9", features = ["clock"] }
```
emschwartz: You should still include `default-features = false`

crates/interledger-settlement/src/api.rs
```diff
@@ -17,166 +18,286 @@ use num_traits::Zero;
...
+    A: SettlementAccount + Send + Sync + 'static,
+{
+    let with_store = warp::any().map(move || store.clone()).boxed();
+    let idempotency = warp::header::optional::<String>("idempotency-key");
+    let account_id = warp::path("accounts").and(warp::path::param2::<String>()); // account_id
+
+    // POST /accounts/:account_id/settlements (optional idempotency-key header)
+    // Body is a Quantity object
+    let settlement_endpoint = account_id.and(warp::path("settlements"));
+    let settlements = warp::post2()
```
emschwartz: @dora-gt what was the conclusion on method / path order?

crates/interledger-settlement/src/api.rs
```diff
@@ -17,166 +18,286 @@ use num_traits::Zero;
...
+        .and(warp::body::json())
+        .and(with_store.clone())
+        .and_then(
+            move |id: String, idempotency_key: Option<String>, quantity: Quantity, store: S| {
+                let input = format!("{}{:?}", id, quantity);
+                let input_hash = get_hash_of(input.as_ref());
+
+                let idempotency_key_clone = idempotency_key.clone();
+                let store_clone = store.clone();
+                let f =
```
emschwartz: What is f?

crates/interledger-settlement/src/api.rs
```diff
@@ -17,166 +18,286 @@ use num_traits::Zero;
...
+                    move || do_receive_settlement(store_clone, id, quantity, idempotency_key_clone);
+                make_idempotent_call(store, f, input_hash, idempotency_key)
+                    // TODO Replace with responses
```
emschwartz: If we want to reuse stuff from the `interledger-api` crate, we _could_ make that a dependency but use conditional compilation to avoid pulling in everything except the error stuff

crates/interledger-settlement/src/api.rs
```diff
@@ -17,166 +18,286 @@ use num_traits::Zero;
...
+    store
+        .load_idempotent_data(idempotency_key.clone())
+        .map_err(move |_| {
+            let error_msg = "Couldn't load idempotent data".to_owned();
+            error!("{}", error_msg);
+            error_msg
+        })
+        .and_then(move |ret: Option<IdempotentData>| {
+            if let Some(ret) = ret {
+                if ret.2 == input_hash {
```
emschwartz: Please include comments describing what the things in this tuple are

crates/interledger-settlement/src/api.rs
dora-gt: It is "method -> path -> auth -> body", why we chose this was that placing method filters seemed more common in the "Warp" world, and it would not produce any downside effects.

crates/interledger-settlement/src/api.rs
```diff
@@ -17,166 +18,294 @@ use num_traits::Zero;
...
+    let account_id = warp::path("accounts").and(warp::path::param2::<String>()); // account_id
+
+    // POST /accounts/:account_id/settlements (optional idempotency-key header)
+    // Body is a Quantity object
+    let settlement_endpoint = account_id.and(warp::path("settlements"));
+    let settlements = warp::post2()
+        .and(settlement_endpoint)
+        .and(warp::path::end())
+        .and(idempotency)
+        .and(warp::body::json())
```
dora-gt: You could use `interledger-http::deserialize_json` if you want.
It returns errors neatly.

crates/interledger-settlement/src/api.rs
```diff
@@ -17,166 +18,294 @@ use num_traits::Zero;
...
+                let input = format!("{}{:?}", id, message);
+                let input_hash = get_hash_of(input.as_ref());
+
+                let store_clone = store.clone();
+                // Wrap do_send_outgoing_message in a closure to be invoked by
+                // the idempotency wrapper
+                let send_outgoing_message_fn =
+                    move || do_send_outgoing_message(store_clone, outgoing_handler, id, message);
+                make_idempotent_call(store, send_outgoing_message_fn, input_hash, idempotency_key)
+                    // TODO Replace with error case with response
```
dora-gt: If you want to use HTTP `error` module, you could just return `ApiError`s with `into()` here, which converts `ApiError` into `Rejection`, and use `default_rejection_handler` to `recover`.

crates/interledger-settlement/src/api.rs
```diff
@@ -17,166 +18,294 @@ use num_traits::Zero;
...
+    // Body is a Quantity object
+    let settlement_endpoint = account_id.and(warp::path("settlements"));
+    let settlements = warp::post2()
+        .and(settlement_endpoint)
+        .and(warp::path::end())
+        .and(idempotency)
+        .and(warp::body::json())
+        .and(with_store.clone())
+        .and_then(
+            move |id: String, idempotency_key: Option<String>, quantity: Quantity, store: S| {
```
dora-gt: I think `id` is confusing. It could be anything, I can't imagine what it is. I'd rather call it `account_id`. If it conflicts, rename `account_id` to `account_id_filter` or something.

crates/interledger-settlement/src/api.rs
```diff
@@ -17,166 +18,294 @@ use num_traits::Zero;
...
-                .load_idempotent_data(idempotency_key.clone())
-                .map_err(move |_| {
-                    let error_msg = "Couldn't load idempotent data".to_owned();
-                    error!("{}", error_msg);
-                    error_msg
+// make_idempotent_call takes a function instead of direct arguments so that we
+// can reuse it for both the messages and the settlements calls
+fn make_idempotent_call<S, A, F>(
+    store: S,
+    f: F,
```
dora-gt: I think `f` is not understandable.
There must be more understandable names, though I don't come up with a nice one.

- `not_idempotent_function`
- `function_to_be_idempotent`
- `raw_function`
- `original_function`

or something.

emschwartz: Agreed, didn't notice that the first time but generally :-1: on single-letter function names or variables

crates/interledger-settlement/src/api.rs
```diff
@@ -17,166 +18,294 @@ use num_traits::Zero;
...
+                            .and_then(
+                                move |ret: (StatusCode, Bytes)| {
+                                    store
+                                        .save_idempotent_data(
+                                            idempotency_key,
+                                            input_hash,
+                                            ret.0,
+                                            ret.1.clone(),
+                                        )
+                                        .map_err({
```
dora-gt: I rethought about this and personally I think mapping this error is **VERY** dangerous.

If the save fails, it means that **we can not take responsibility for idempotency**. They would retry when accidentally they lost the connection etc. Then some resources might be affected by the retry requests.

Shouldn't we return some special error responses for these? This means **the entire system isn't working properly**. At least it should be logged somewhere. It might cause serious situation later.

gakonst: What else we can do other than warning that an idempotent operation failed? Should we just log the error and reject the message? We'd have to rollback any state transitions we did in that function. 

I think logging the error should be fine, don't think it's that likely that there will be both an error connecting to a local store AND us taking too long to respond such that the sender tries to resend a message. 

(remember that the sender in this case is the engine, which is a component we control, so it's unlikely imo that it will try to spam us maliciously to take advantage of this)

dora-gt: I think ultimately we could consider even stopping the system, logging the incident because it is not working properly. If we could continue working without the idempotency protection, it makes me feel like "Why do we need this?".

From my experience, **something could happen just happens**. It's `potential * tries`. When I was working a game company that runs a title where tons of users connect and transact every second, a super minor condition bug that requires "nano" sec level condition occurred sometimes, which seemed almost impossible. We had very very hard time then to figure out what was the problem and how to fix the problem. I'm afraid of having that kind of problem in this project.

I'm not sure everyone agrees with my opinion but I think it is better to stop the entire system if we are going to (possibly) break data. It spawns more and more problems. Stopping system keeps the wound as small as possible.

And thank you for the log. As we did something we could at this point, I'll approve this.
It might be better to make it more robust in the future.

dora-gt: (Sorry about my rebase, if you have any problem please let me know)

crates/interledger-http/src/error/error_types.rs
```diff
@@ -1,80 +1,112 @@
...
+    r#type: &ProblemType::Default,
+    title: "Conversion error",
+    status: StatusCode::INTERNAL_SERVER_ERROR,
+};
+
+// Account without an engine error
+pub const NO_ENGINE_CONFIGURED_ERROR_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    status: StatusCode::NOT_FOUND,
+    title: "No settlement engine configured",
```
emschwartz: Seems like this should be defined where the settlement-related functionality is rather than in `interledger-http`

crates/interledger-http/src/error/error_types.rs
```diff
@@ -1,80 +1,112 @@
...
+
+// Idempotency errors
+pub const IDEMPOTENT_STORE_CALL_ERROR_TYPE: ApiErrorType = ApiErrorType {
+    r#type: &ProblemType::Default,
+    title: "Store idempotency error",
+    status: StatusCode::CONFLICT,
+};
+
+// Number coversion errors
+pub const CONVERSION_ERROR_TYPE: ApiErrorType = ApiErrorType {
```
emschwartz: Don't think this should be in this crate either

crates/interledger-http/src/error/mod.rs
```diff
@@ -1,5 +1,6 @@
 mod error_types;
 pub use error_types::*;
+pub use http::StatusCode;
```
emschwartz: Doesn't seem like this is used

crates/interledger-http/src/idempotency.rs
```diff
@@ -0,0 +1,143 @@
...
+use bytes::Bytes;
+use futures::executor::spawn;
+use futures::{
+    future::{err, ok, Either},
+    Future,
+};
+use http::StatusCode;
+use log::error;
+
+pub type IdempotentData = (StatusCode, Bytes, [u8; 32]);
```
emschwartz: Can you just make this a struct so it's clearer what's going on?

crates/interledger-http/src/lib.rs
```diff
@@ -15,6 +15,7 @@ mod server;
 
 // So that settlement engines can use errors
 pub mod error;
+pub mod idempotency;
```
emschwartz: Should this be behind a cargo feature?

crates/interledger-settlement/src/api.rs
emschwartz: This sounds useful

gakonst:
Comments addressed, thanks all!

PR #422, opened 2019-10-15
chore: update Cargo.lock with new diff-friendly lockfile format

bstrie:
Cargo has had a "version 2" of its lockfile format in the works for a bit now. This new format is designed to be diff-friendly, by omitting unnecessary data when dependency resolution is otherwise unambiguous and by moving some fields around to localize lockfile edits to contiguous chunks. Support for this new format was committed to Cargo in July, which means that all Rust versions since and including Rust 1.38 understand the new format. The new format is not yet enabled by default, which means that no released version of Cargo will freshly-generate this lockfile format on its own, but all of them will respect and preserve the new format if it is already found. The new lockfile in this patch was generated by forking Cargo to emit the new lockfile format by default and running `cargo update`. A subsequent `cargo update` using Cargo 1.38 caused no errors or changes to the lockfile, confirming that anyone using versions of Cargo that are at least this recent will have no trouble working with the new format.

The potential benefit of using this new format today rather than when it is eventually turned on by default (which hasn't yet been done even in nightly, so it won't be in stable for a few cycles at least) is that we often run into conflicts with the Cargo.lock when working in our different branches, so switching to the new format may reduce these.

The downside of making this change today is that if anyone ends up deleting and regenerating Cargo.lock in order to blithely resolve a merge conflict (as I myself have been known to do), their Cargo will default to using the "version 1" lockfile format, which would totally undo this change. Not sure whether the risk and annoyance of that makes this PR worth accepting (it doesn't even build yet, because I didn't bother spending much time on it without first bringing it up for discussion), but I did want to propose the option.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,2840 +4,2847 @@
 name = "adler32"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d2e7343e7fc9de883d1b0341e0b13970f764c14101234857d2ddafa1cb1cac2"
 
 [[package]]
 name = "aho-corasick"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "58fb5e95d83b38284460a5fda7d6470aa0b8844d283a0b614b8535e880800d2d"
 dependencies = [
- "memchr 2.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr",
 ]
 
 [[package]]
 name = "arrayvec"
 version = "0.4.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd9fd44efafa8690358b7408d253adf110036b88f55672a933f01d616ad9b1b9"
 dependencies = [
- "nodrop 0.1.14 (registry+https://github.com/rust-lang/crates.io-index)",
+ "nodrop",
 ]
 
 [[package]]
 name = "ascii"
 version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


dora-gt:
A few questions:

- What happens if I add a new dependency with older Cargo?
- How could we use new Cargo to generate v2 format?

also I think this should be documented somewhere if we dare to turn this on.

dora-gt:
Sorry my questions were unclear.
I would like to make it clear whether I can **add** new dependencies with normal 1.38 because I think it would happen relatively often.

tarcieri:
My understanding is the new format is in flux and they've mainly only been trying it out on `rust-lang/rust` itself to gather experience with it. Has it actually stabilized now?

bstrie:
> I would like to make it clear whether I can add new dependencies with normal 1.38 because I think it would happen relatively often.

@dora-gt Cargo 1.38 would be able to add dependencies, yes. Stable versions of Cargo are fully able to understand and produce the new format if it is found on the filesystem. The only thing that stable Cargo lacks is the ability to produce the new format if it is not already found.

> My understanding is the new format is in flux and they've mainly only been trying it out on rust-lang/rust itself to gather experience with it. Has it actually stabilized now?

@tarcieri It exists in stable Cargo 1.38, which is able to read and produce the new format when the new format is detected. I suppose they have the latitude to outright deprecate the new format and remove support for it if they desired, since there currently exists no stable way to *explicitly* produce it. However they are unable to alter this "version 2" format because it would mean eventual breakage for anyone trying to use that altered version with the stable 1.38 release; at this point they would need to introduce a brand-new "version 3" to do so, which seems unlikely.

bstrie:
Closing for now, let's revisit once we start enforcing a minimum Rust version (https://github.com/interledger-rs/interledger-rs/issues/449).

PR #305, opened 2019-09-15
Enables `run-md-test.sh` on CircleCI

dora-gt:
A part of #198

Resolves:
- #198 Make `.md`s testable
- #335 Make examples use CLI args vs env vars consistently
- Part of #230 (Docker images will be tagged)
- Part of #230 ([Binaries are built and attached to a release](https://github.com/interledger-rs/interledger-rs/pull/471))

---

- This PR enables `run-md-test.sh` to be run on CircleCI.
    - Also if the branch is `master` and all tests have passed, it updates docker images on DockerHub. Also, tagged commit would be pushed, being tagged with the git tag name.
- Because running `docker` command in containers requires `docker-in-docker`(`dind`), `docker/run-md-test.dockerfile` is based on `circleci/buildpack-deps:latest-dind`.
- Also, `dind` requires another job because it requires another image, so the job is defined as `test-md`.
- The `setup_remote_docker` command spins up a `docker` daemon in an **outer** machine. So we can't access it using `curl localhost:xxxx` which is used to test. In order to address this issue, `ncat` hack was used.
    - `ncat -l -k -c "docker exec -i interledger-rs-node_a nc 127.0.0.1 7770" -p 7770 &`
- `interledgerrs/settlement-engine` repo is deprecated, using `interledgerrs/settlement-engines` instead because the name looked wrong.

image

Questions:
- I'm wondering this is better or not. Using `Cargo.toml` doesn't seem to be efficient.
    - `cargo-cache-2-{{ checksum "Cargo.lock" }}-{{ .Branch }}` (changing `Cargo.toml` to `Cargo.lock`)
    - We have to remove `Cargo.lock` from `.gitignore` to do this.

PR #428, opened 2019-10-17
docs: fix readmes for eth engines now that they are in a separate repo

gakonst:
as title

### DIFF
--- a/examples/eth-settlement/README.md
+++ b/examples/eth-settlement/README.md
@@ -196,12 +196,9 @@ sleep 3
 -->
 
 ### 4. Launch Settlement Engines
-Because each node needs its own settlement engine, we need to launch both a
-settlement engine for Alice's node and another settlement engine for Bob's node.
+Because each node needs its own settlement engine, we need to launch both a settlement engine for Alice's node and another settlement engine for Bob's node.
 
-Note: The engines are part of a [separate
-repository](https://github.com/interledger-rs/settlement-engines) so you have to
-clone and install them according to [the instructions](https://github.com/interledger-rs/settlement-engines/blob/master/README.md)
+Note: The engines are part of a [separate repository](https://github.com/interledger-rs/settlement-engines) so you have to clone and install them according to [the instructions](https://github.com/interledger-rs/settlement-engines/blob/master/README.md)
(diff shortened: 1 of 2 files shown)


examples/eth-settlement/README.md
```diff
@@ -207,7 +212,7 @@ if [ "$USE_DOCKER" -eq 1 ]; then
         --network=interledger \
         --name=interledger-rs-se_a \
         -td \
-        interledgerrs/settlement-engine ethereum-ledger \
```
gakonst: note: Settlement-Engine was deprecated so we changed to the new one

examples/eth-settlement/README.md
```diff
@@ -196,7 +196,12 @@ sleep 3
 -->
 
 ### 4. Launch Settlement Engines
-Because each node needs its own settlement engine, we need to launch both a settlement engine for Alice's node and another settlement engine for Bob's node.
+Because each node needs its own settlement engine, we need to launch both a
+settlement engine for Alice's node and another settlement engine for Bob's node.
```
dora-gt: I think we should use soft wrap. Hard wrap changes lines which were not actually changed just like these lines, which is not good.

dora-gt:
Sorry I do really need this branch to be merged, I'll fix hard wrap problem and merge.

dora-gt:
It seems that cloning inside `interledger-rs` was a bad idea. It doesn't work.

```
error: current package believes it's in a workspace when it's not:
current:   /Users/dora_gt/Documents/git/interledger-rs/examples/eth-settlement/settlement-engines/Cargo.toml
workspace: /Users/dora_gt/Documents/git/interledger-rs/Cargo.toml

this may be fixable by adding `examples/eth-settlement/settlement-engines` to the `workspace.members` array of the manifest located at: /Users/dora_gt/Documents/git/interledger-rs/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

I'm fixing also this in #305 but examples are so broken now like even settlements don't happen and I'm struggling to fix this. Give me a bit more time.

PR #432, opened 2019-10-17
refactor(cli): now CLI doesn't include when returning the response

dora-gt:
Now CLI doesn't include `\n` when returning the response so that it more looks like CURL style.

Because of the `\n`, examples look a bit weird. If you guys don't have any preference, I'd like to introduce this.

### DIFF
--- a/crates/ilp-cli/src/main.rs
+++ b/crates/ilp-cli/src/main.rs
@@ -31,7 +31,7 @@ pub fn main() {
             Ok(body) => {
                 if response.status().is_success() {
                     if !matches.is_present("quiet") {
-                        println!("{}", body);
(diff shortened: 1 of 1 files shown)


bstrie:
Is the goal of this to make the docker examples produce the same output as the non-docker examples? Because ultimately my goal was to make the docker examples stop using curl and move to ilp-cli, which would make this unnecessary.

dora-gt:
If this command were a "normal" command, I think it is preferred to place `\n` but this is returning the output from the server. So I think `curl` is respecting its output. Also I thought it would be more flexible, we could choose how to deal with it.

That said, actually we can fix this in another way. I fixed the examples in such ways. So closing this.

PR #436, opened 2019-10-18
fix(three node example): fixes three node example

dora-gt:
This is originally a part of #305 but I feel like we should not leave this, people are struggling to try out `interledger-rs`.

Cloning settlement-engines inside interledger-rs is not good because it causes an error. In
addition, node setting of Charlie is wrong.

### DIFF
--- a/examples/eth-xrp-three-nodes/README.md
+++ b/examples/eth-xrp-three-nodes/README.md
@@ -314,7 +314,7 @@ else
 -->
 
 ```bash #
-# This should be done outside of the interledger-rs ditrectory, otherwise it will cause an error
+# This should be done outside of the interledger-rs directory, otherwise it will cause an error
 git clone https://github.com/interledger-rs/settlement-engines
 cd settlement-engines
 ```
--- a/examples/eth-xrp-three-nodes/README.md
+++ b/examples/eth-xrp-three-nodes/README.md
@@ -46,7 +46,7 @@ Interledger.rs and settlement engines written in other languages are fully inter
 
 Install the settlement engine as follows:
 
-```bash
+```bash #
 npm i -g ilp-settlement-xrp
 ```
 
@@ -147,7 +147,7 @@ else
 -->
 
 ```bash
-cargo build --all-features --bin ilp-node
+cargo build --bin ilp-node
 ```
 
 <!--!
@@ -247,7 +247,7 @@ In this example, we'll connect 3 Interledger nodes and each node needs its own s
 
 By default, the XRP settlement engine generates new testnet XRPL accounts prefunded with 1,000 testnet XRP (a new account is generated each run). Alternatively, you may supply an `XRP_SECRET` environment variable by generating your own testnet credentials from the [official faucet](https://xrpl.org/xrp-test-net-faucet.html).
 
-Note: The rust engines are part of a [separate repository](https://github.com/interledger-rs/settlement-engines) so you have to clone and install them according to [the instructions](https://github.com/interledger-rs/settlement-engines/blob/master/README.md)
+The engines are part of a [separate repository](https://github.com/interledger-rs/settlement-engines) so you have to clone and install them according to [the instructions](https://github.com/interledger-rs/settlement-engines/blob/master/README.md). In case you've never cloned `settlement-engine`, clone it first.
 
 <!--!
 printf "\nStarting settlement engines...\n"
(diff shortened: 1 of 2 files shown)


examples/eth-xrp-three-nodes/README.md
```diff
@@ -306,13 +306,27 @@ if [ "$USE_DOCKER" -eq 1 ]; then
...
     which ilp-settlement-xrp &> /dev/null || error_and_exit "You need to install \"ilp-settlement-xrp\"."
+    pushd "${SETTLEMENT_ENGINE_INSTALLL_DIR}" &>/dev/null
+    if [ ! -e "settlement-engines" ]; then
+        git clone https://github.com/interledger-rs/settlement-engines
+    fi
+    pushd settlement-engines &>/dev/null
 -->
 
+```bash #
+# Do this somewhere OUTER the interledger-rs directory otherwise you'll get an error.
```
emschwartz:
```suggestion
# This should be done outside of the interledger-rs ditrectory, otherwise it will cause an error
```

examples/eth-xrp-three-nodes/README.md
```diff
@@ -429,16 +444,15 @@ ILP_ADMIN_AUTH_TOKEN=hi_bob \
 ILP_REDIS_URL=redis://127.0.0.1:6381/ \
 ILP_HTTP_BIND_ADDRESS=127.0.0.1:8770 \
 ILP_SETTLEMENT_API_BIND_ADDRESS=127.0.0.1:8771 \
-cargo run --all-features --bin ilp-node &> logs/node-bob.log &
+cargo run --bin ilp-node &> logs/node-bob.log &
 
 # Start Charlie's node
-ILP_ADDRESS=example.bob.charlie \
```
dora-gt: This is kind of confusing. This should be documented or fixed in the future.

examples/eth-xrp-three-nodes/README.md
```diff
@@ -247,7 +247,7 @@ In this example, we'll connect 3 Interledger nodes and each node needs its own s
 
 By default, the XRP settlement engine generates new testnet XRPL accounts prefunded with 1,000 testnet XRP (a new account is generated each run). Alternatively, you may supply an `XRP_SECRET` environment variable by generating your own testnet credentials from the [official faucet](https://xrpl.org/xrp-test-net-faucet.html).
 
-Note: The rust engines are part of a [separate repository](https://github.com/interledger-rs/settlement-engines) so you have to clone and install them according to [the instructions](https://github.com/interledger-rs/settlement-engines/blob/master/README.md)
+The engines are part of a [separate repository](https://github.com/interledger-rs/settlement-engines) so you have to clone and install them according to [the instructions](https://github.com/interledger-rs/settlement-engines/blob/master/README.md). In case you've never cloned `settlement-engine`, clone it first.
```
sudheesh001:
```suggestion
The engines are part of a [separate repository](https://github.com/interledger-rs/settlement-engines) so you have to clone and install them according to [the instructions in settlement-engine](https://github.com/interledger-rs/settlement-engines/blob/master/README.md). In case you've never cloned `settlement-engine`, the first step would be to clone the repository.
```

examples/eth-xrp-three-nodes/README.md
```diff
@@ -338,7 +352,7 @@ cargo run --features "ethereum" -- ethereum-ledger \
 --settlement_api_bind_address 127.0.0.1:3001 \
 &> logs/node-bob-settlement-engine-eth.log &
 
-cd ..
+# Now go back to interledger-rs directory.
```
sudheesh001:
```suggestion
# Now go back to the `interledger-rs` directory.
```

examples/eth-xrp-three-nodes/README.md
sudheesh001: According to the diagram and the way these nodes are available, we could even call charlie's node address as `example.charlie` to be consistent with `alice` and `bob`

examples/eth-xrp-three-nodes/README.md
```diff
@@ -306,13 +306,27 @@ if [ "$USE_DOCKER" -eq 1 ]; then
...
     which ilp-settlement-xrp &> /dev/null || error_and_exit "You need to install \"ilp-settlement-xrp\"."
+    pushd "${SETTLEMENT_ENGINE_INSTALLL_DIR}" &>/dev/null
+    if [ ! -e "settlement-engines" ]; then
+        git clone https://github.com/interledger-rs/settlement-engines
+    fi
+    pushd settlement-engines &>/dev/null
 -->
 
+```bash #
+# This should be done outside of the interledger-rs ditrectory, otherwise it will cause an error
```
sudheesh001:
```suggestion
# This should be done outside of the interledger-rs directory, otherwise it will cause an error
```

examples/eth-xrp-three-nodes/README.md
dora-gt: I suppose the reason is that if we set `Charlie`'s node as a peer, it would take 20secs or so to completely establish the route from Alice to Charlie. That looks inappropriate in this example. So we have to set this `child/parent` relation

dora-gt:
Thank you for the quick responses, I'm merging this.

PR #444, opened 2019-10-22
Update `api.md`

dora-gt:
- Added missing API endpoints.
- Added an explanation of authorization spec (though this might need to be changed in the future)
- Changed the balance API to return a number

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -180,7 +180,7 @@ In the format of text message of WebSocket, the endpoint will send the following
 }
 ```
 
-Note that the sending account username is one on the receiving node, not one of terminal sender.
+Note that the `from_username` corresponds to the account that received the packet _on this node_, not the original sender.
 
 ### GET /accounts/:username/spsp
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -2,13 +2,53 @@
 
 For instructions on running the ILP Node, see the [Readme](../README.md).
 
-## Authentication
+## Authorization
 
-The ILP Node uses HTTP Bearer Token authorization. Most requests must either be authenticated with the admin token configured on the node or the token configured for a particular account.
+The ILP Node uses HTTP Bearer Token authorization. Most requests must either be authorized with the admin token configured on the node or the token configured for a particular account.
+
+Your HTTP requests would look like:
+
+```
+GET /accounts HTTP/1.1
+Authorization: Bearer BEARER-TOKEN-HERE
+```
+
+The format of the `Authorization` header is either of the following.
+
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -93,6 +162,26 @@ Account-holder only.
...
+{
+    "to_username": "Receiving account username",
+    "from_username": "Sending account username",
+    "destination": "Destination ILP address",
+    "amount": 1000,
+    "timestamp": "Receiving time in RFC3339 format"
+}
+```
+
+Note that the sending account username is one on the receiving node, not one of terminal sender.
```
dora-gt: As far as I read the code, I thought so. Is this correct?

emschwartz: Yeah that's correct

```suggestion
Note that the `from_username` corresponds to the account that received the packet _on this node_, not the original sender.
```

docs/api.md
```diff
@@ -145,7 +234,9 @@ Health check.
 
 ```json
 {
-    "status": "Ready"
+    "status": "Ready",
+    "ilp_address": "example.node",
+    "version": "0.1.1beta.3"
```
gakonst: (nit) I think this should be:
```suggestion
    "version": "0.1.1-beta.3"
```

docs/api.md
bstrie: @emschwartz That suggested change is still worded strangely to me, the `from_username` is the account on the previous connector that just sent us this packet. It's not guaranteed to be the originator that initiated this payment transfer, but it's not necessarily an account on this node, unless I'm grossly mistaken.

bstrie: Could be worth considering whether we ought to rename these fields like this in order to be clearer:

`to_username` -> `destination_user`

`destination` -> `destination_address`

`from_username` to `forwarding_user`

Not sure how willing we'd be to make a breaking change like this, but it could reduce confusion in the future.

emschwartz: The `from_username` is guaranteed to be the account _on this node_ that received the packet (which may represent the previous node -- or which could be the sender in the unusual case where the sender was on the same node)

emschwartz:
> Should I add Prometheus?

@dora-gt Let's do that in another PR. I would put it in a separate doc because it's an API that runs on a different port and needs to be enabled separately anyway

docs/api.md
dora-gt: I think it looks better but currently nodes return the version in this way.
I'll post an issue.

dora-gt: https://github.com/interledger-rs/interledger-rs/issues/452

Issue #439, opened 2019-10-21
The settlement of `eth-xrp-three-nodes` example **sometimes** fails


dora-gt:
The settlement of `eth-xrp-three-nodes` example **sometimes** fails. The logs are attached.

Potential causes are (though I'm not so sure):

`interledger-rs-node_b.log`
```
[ERROR interledger_store_redis::store] Error getting account from HTTP auth: An error was signalled by the server: Error running script (call to f_4f309de3c6e874e5de6f59384504d1a2d3f0c564): @user_script:5: user_script:5: attempt to concatenate local 'id' (a boolean value) 
```

`interledger-rs-node_b.log`
```
[ERROR interledger_settlement::api] Error sending message to peer settlement engine. Packet rejected with code: F00, message: Settlement engine rejected request with error code: 404 Not Found
```

- Failure case
    - [eth-xrp-three-nodes.zip](https://github.com/interledger-rs/interledger-rs/files/3749820/eth-xrp-three-nodes.zip)
- Success case
    - [eth-xrp-three-nodes.zip](https://github.com/interledger-rs/interledger-rs/files/3749888/eth-xrp-three-nodes.zip)

bstrie:
The first error is an error from within a Lua script, likely at the following line: https://github.com/interledger-rs/interledger-rs/blob/0bd9ce3e9f4332016bd93e96a855b23d04aafdd8/crates/interledger-store-redis/src/store.rs#L101 . The `..` is the concatenation operator in Lua, which works with both strings and numbers but not booleans. So the question is why `id` ends up being a boolean in some cases. (I don't think it's a case of `id` being the value 0 or 1, because that's not really how Lua represents booleans.) The `id` variable is received from the `redis.call` function, whose documentation lists the circumstances under which it will return which Lua types:
>    Redis integer reply -> Lua number
>    Redis bulk reply -> Lua string
>    Redis multi bulk reply -> Lua table (may have other Redis data types nested)
>    Redis status reply -> Lua table with a single ok field containing the status
>    Redis error reply -> Lua table with a single err field containing the error
>    Redis Nil bulk reply and Nil multi bulk reply -> Lua false boolean type
So it looks like redis is returning `false` to indicate that `redis.call('HGET', 'usernames', ARGV[1])` didn't find anything, which is strange because this `redis.call` only happens when the following returns true: `redis.call('HEXISTS', 'usernames', username)`.

[...]

PR #448, opened 2019-10-22
Lua tweaks

bstrie:
The discussion at https://github.com/interledger-rs/interledger-rs/issues/439 got me thinking about our Lua scripts, and this cleans them up a bit. I did notice that the test suite was printing the same "attempt to concatenate" error message as in https://github.com/interledger-rs/interledger-rs/issues/439 and that this error message disappeared following these commits, so I can't confirm that that issue is fixed but it should at least help us track it down further.

### DIFF
--- a/crates/interledger-store-redis/src/store.rs
+++ b/crates/interledger-store-redis/src/store.rs
@@ -94,162 +94,23 @@ lazy_static! {
 
     /// This lua script fetches an account associated with a username. The client
     /// MUST ensure that the returned account is authenticated.
-    static ref ACCOUNT_FROM_USERNAME: Script = Script::new("
-    local username = ARGV[1]
-    if redis.call('HEXISTS', 'usernames', username) then
-        local id = redis.call('HGET', 'usernames', ARGV[1])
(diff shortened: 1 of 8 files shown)


crates/interledger-store-redis/src/store.rs
```diff
@@ -94,162 +94,23 @@ lazy_static! {
...
     /// MUST ensure that the returned account is authenticated.
-    static ref ACCOUNT_FROM_USERNAME: Script = Script::new("
-    local username = ARGV[1]
-    if redis.call('HEXISTS', 'usernames', username) then
-        local id = redis.call('HGET', 'usernames', ARGV[1])
-        return redis.call('HGETALL', 'accounts:' .. id)
-    else
-        return nil
-    end");
+    static ref ACCOUNT_FROM_USERNAME: Script = Script::new(include_str!("lua/account_from_username.lua"));
```
dora-gt: 👀
I didn't know `include_str`

bstrie: It's great! https://doc.rust-lang.org/std/macro.include_str.html

PR #451, opened 2019-10-22
chore: update secrecy to v0.5.0

bstrie:
The secrecy crate recently released both 0.4.1 and 0.5.0. We're on 0.4.0, but it turns out that 0.4.1 is accidentally a semver-incompatible release, which means that any `cargo update` right now is going to fail. I figure now's as good a time as any to update the dependency.


tarcieri:
Sorry about that. Some fallout from the `zeroize` 1.0 release. The good news is `zeroize` is stable now.

bstrie:
No worries, it's a fine price to pay for the 1.0 release of a good crate. :)

Issue #452, opened 2019-10-23
`GET /` returns an incorrect version number


dora-gt:
Originally pointed out at:
https://github.com/interledger-rs/interledger-rs/pull/444#discussion_r337678746

Now it returns the version like `0.1.1beta.3` but according to [Semantic Versioning](https://semver.org/#spec-item-9) it should be `0.1.1-beta.3`.

`docs/api.md` should be fixed as well.

PR #453, opened 2019-10-23
fix(health-check api): fix to return a correct semantic version

dora-gt:
Resolves: #452

Can't we just use `CARGO_PKG_VERSION`?


dora-gt:
Just taking a quick look at:
https://github.com/rust-lang/cargo/blob/master/src/cargo/core/compiler/compilation.rs#L259
It seems that `CARGO_PKG_VERSION` is always given, I changed to use this.

dora-gt:
I think this doesn't cause any serious problem right now, so I'm merging this.

Issue #419, opened 2019-10-14
Include asset details in SPSP response


emschwartz:
The `delivered_amount` should say what asset code and scale that number represents.

We might also want to include the sent amount in the response.

emschwartz:
@gakonst Just to be clear, this is the response to _sending_ a payment, not the SPSP query response

PR #456, opened 2019-10-23
Enhance details when sending an SPSP payment

gakonst:
Closes https://github.com/interledger-rs/interledger-rs/issues/419. 

Uses the `ConnectionAssetDetails` frame to return additional useful information to the sender.

Running the simple example:

```
Sending payment of 500 from Alice (on Node A) to Bob (on Node B)...

{
   "delivered_asset_scale" : 9,
   "delivered_asset_code" : "ABC",
   "sent_amount" : 500,
   "sent_asset_code" : "ABC",
   "from" : "example.node_a.alice",
   "to" : "example.node_b.bob.-p3zU4tXsDRCBLg8vt_U6iiyQ5pgZk4MfoCaG1wZDW8",
   "delivered_amount" : 500,
   "sent_asset_scale" : 9
}
```

### DIFF
--- a/crates/interledger-api/src/routes/accounts.rs
+++ b/crates/interledger-api/src/routes/accounts.rs
@@ -363,21 +363,16 @@ where
         .and(deserialize_json())
         .and(with_incoming_handler.clone())
         .and_then(
-            |account: A, pay_request: SpspPayRequest, incoming_handler: I| {
+            move |account: A, pay_request: SpspPayRequest, incoming_handler: I| {
                 pay(
                     incoming_handler,
-                    account,
+                    account.clone(),
                     &pay_request.receiver,
                     pay_request.source_amount,
                 )
-                .and_then(|delivered_amount| {
-                    debug!(
-                        "Sent SPSP payment and delivered: {} of the receiver's units",
-                        delivered_amount
-                    );
-                    Ok(warp::reply::json(&json!({
-                        "delivered_amount": delivered_amount
-                    })))
+                .and_then(move |receipt| {
+                    debug!("Sent SPSP payment, receipt: {:?}", receipt);
+                    Ok(warp::reply::json(&json!(receipt)))
                 })
                 .map_err::<_, Rejection>(|err| {
                     error!("Error sending SPSP payment: {:?}", err);
--- a/crates/interledger-api/tests/accounts.rs
+++ b/crates/interledger-api/tests/accounts.rs
@@ -531,7 +531,6 @@ fn accounts_test() {
                 assert!(res.error_for_status_ref().is_ok(), "{}", &content);
                 let receipt: StreamDelivery = serde_json::from_str(&content).unwrap();
                 assert_eq!(receipt.from, Address::from_str(ILP_ADDRESS_1).unwrap());
-                println!("{}", receipt.to.to_owned());
                 assert!(receipt.to.to_string().starts_with(&format!("{}.{}", NODE_ILP_ADDRESS, USERNAME_2)));
                 assert_eq!(receipt.sent_amount, amount);
                 assert_eq!(receipt.sent_asset_scale as usize, ASSET_SCALE);
--- a/crates/interledger-api/tests/accounts.rs
+++ b/crates/interledger-api/tests/accounts.rs
@@ -11,7 +11,7 @@ use tokio::runtime::Builder as RuntimeBuilder;
 
 use ilp_node::InterledgerNode;
 use interledger::{packet::Address, service::Username};
-use interledger_stream::Receipt;
+use interledger_stream::StreamDelivery;
 
 // Integration tests of accounts APIs
 // These are very rough tests. It confirms only that the paths and HTTP methods are working correctly.
@@ -529,7 +529,7 @@ fn accounts_test() {
             .and_then(move|mut res| {
                 let content = res.text().wait().expect("Error getting response!");
                 assert!(res.error_for_status_ref().is_ok(), "{}", &content);
-                let receipt: Receipt = serde_json::from_str(&content).unwrap();
+                let receipt: StreamDelivery = serde_json::from_str(&content).unwrap();
                 assert_eq!(receipt.from, Address::from_str(ILP_ADDRESS_1).unwrap());
                 println!("{}", receipt.to.to_owned());
                 assert!(receipt.to.to_string().starts_with(&format!("{}.{}", NODE_ILP_ADDRESS, USERNAME_2)));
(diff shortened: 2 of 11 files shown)


crates/interledger-api/tests/accounts.rs
```diff
@@ -527,12 +528,16 @@ fn accounts_test() {
...
-                    panic!("Invalid response JSON! {}", &content);
-                }
+                let receipt: Receipt = serde_json::from_str(&content).unwrap();
+                assert_eq!(receipt.from, Address::from_str(ILP_ADDRESS_1).unwrap());
+                println!("{}", receipt.to.to_owned());
+                assert!(receipt.to.to_string().starts_with(&format!("{}.{}", NODE_ILP_ADDRESS, USERNAME_2)));
+                assert_eq!(receipt.sent_amount, amount);
+                assert_eq!(receipt.sent_asset_scale as usize, ASSET_SCALE);
+                assert_eq!(receipt.sent_asset_code, ASSET_CODE);
+                assert_eq!(receipt.delivered_amount, amount * 1000);
```
gakonst: (asset scale 9 -> 12 means that the receiver should receive 1000 * sent_amount in their scale)

crates/interledger-stream/src/client.rs
```diff
@@ -251,7 +298,22 @@ where
...
                 // TODO check that the sequence matches our outgoing packet
-                self.delivered_amount += packet.prepare_amount();
+
+                // Update the asset scale & asset code via the received
+                // frame. https://github.com/interledger/rfcs/pull/551
+                // ensures that this won't change, so we only need to
+                // perform this loop once.
+                if self.receipt.delivered_asset_scale.is_none() {
+                    for frame in packet.frames() {
+                        if let Frame::ConnectionAssetDetails(frame) = frame {
```
kincaidoneil: What if the asset details frame is received within a Reject packet? (e.g. if the first packet had too large of an amount and got an F08 or something)

crates/interledger-stream/src/client.rs
```diff
@@ -11,13 +11,54 @@ use interledger_packet::{
...
+use serde::{Deserialize, Serialize};
 use std::{
     cell::Cell,
     cmp::min,
     str,
     time::{Duration, SystemTime},
 };
 
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub struct Receipt {
```
kincaidoneil: I'm indifferent to how it's represented internally, but in the returned JSON from the API call, could we change it to something like this, which might be a bit more idiomatic?

```json
{
    "sent": {
        "amount": "...",
        "asset_scale": 9,
        "asset_code": "..."
    },
    "delivered": {
        "amount": "...",
        "asset_scale": 6,
        "asset_code": "..."
    },
    "from": "...",
    "to": "...",
}
```

(Also, not sure how this ends up being serialized, but I think in the returned JSON the amounts should be strings since JavaScript can't support u64 numbers, but the scale can be a number because it's u8.)

crates/interledger-stream/src/server.rs
```diff
@@ -190,7 +195,11 @@ where
 // TODO send asset code and scale back to sender also
```
kincaidoneil: This TODO can be removed

crates/interledger-stream/src/server.rs
```diff
@@ -229,6 +238,11 @@ fn receive_money(
         }
     }
 
+    response_frames.push(Frame::ConnectionAssetDetails(ConnectionAssetDetailsFrame {
```
kincaidoneil: Why is this sent back on every packet? Couldn't we just send it once in response to e.g. the `ConnectionNewAddress` frame? (I think that's what JS Stream essentially does)

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -363,21 +363,16 @@ where
...
-                .and_then(|delivered_amount| {
-                    debug!(
-                        "Sent SPSP payment and delivered: {} of the receiver's units",
-                        delivered_amount
-                    );
-                    Ok(warp::reply::json(&json!({
-                        "delivered_amount": delivered_amount
-                    })))
+                .and_then(move |receipt| {
+                    debug!("Sent SPSP payment, receipt: {:?}", receipt);
```
emschwartz: We shouldn't use the term receipt, because it's likely that we will implement a protocol for cryptographically signed receipts at some point

crates/interledger-api/tests/accounts.rs
emschwartz: I'd leave out the "slippage" field for now

crates/interledger-stream/src/client.rs
emschwartz: :+1: this could be in either a fulfill or reject

crates/interledger-api/tests/accounts.rs
gakonst: yeah just mentioning it as an idea for the future maybe

crates/interledger-stream/src/client.rs
gakonst: As long as we receive 1 Fulfill packet this will still work, so I do not think there is any change required for this

emschwartz: What if the first packet is rejected and the frame isn't sent in every response?

kincaidoneil:
> This seems like a fine addition, though unrelated to this PR

👍

> I think it should always be the sender's job to determine how much the receiver should receive. That way, the receiver behavior is very predictable and does not change

Make sense, and I agree that should suffice in the vast majority of cases, paying invoices, etc. My only counter example might be is the if the sender is sending a gift or donation to the receiver (so the sender is paying up to some amount in their units, rather than paying up to some amount defined by the recipient in the recipient's units), the sender may be indifferent to the rate, so the receiver might have their own preference for how strict of an exchange rate they want to accept.

crates/interledger-stream/src/client.rs
gakonst: The `ConnectionAssetDetails` frame is sent after a [`ConnectionNewAddress` frame](https://github.com/interledger-rs/interledger-rs/blob/db9efb308efba327c429b587f37845bdb811c80e/crates/interledger-stream/src/server.rs#L244) is received.

The sender sends a `ConnectionNewAddress` frame if the [`should_send_source_account` variable is true](https://github.com/interledger-rs/interledger-rs/blob/db9efb308efba327c429b587f37845bdb811c80e/crates/interledger-stream/src/client.rs#L173). This variable starts as true, and [ONLY gets set to false when we receive a fulfill.](https://github.com/interledger-rs/interledger-rs/blob/db9efb308efba327c429b587f37845bdb811c80e/crates/interledger-stream/src/client.rs#L296). 

As a result, I think that there is indeed no need to add any handling logic for rejects.

emschwartz: That is the current behavior of the Rust STREAM client, not part of the spec for how STREAM _must_ be implemented. We may change this behavior later and other implementations may handle it differently

PR #462, opened 2019-10-24
feat: config option to make exchange rate polling more forgiving

### LINKED ISSUE #304: Tolerate failed rate backend requests
Right now, rates are updated at the same interval they're invalidated as too old (60 seconds, by default). The problem is any single failed request results in the connector being unavailable until the next request succeeds, since existing rates are discarded if the request fails. Rates should be updated more frequently than their maximum acceptable age to tolerate one-off request failures with the rate backend.

_Originally posted by @kincaidoneil in https://github.com/interledger-rs/interledger-rs/pull/295_

bstrie:
closes #304

### DIFF
--- a/crates/ilp-node/src/node.rs
+++ b/crates/ilp-node/src/node.rs
@@ -67,6 +67,10 @@ fn default_exchange_rate_poll_interval() -> u64 {
     60_000
 }
 
+fn default_exchange_rate_failed_polls_before_invalidation() -> u64 {
+    5
+}
+
 fn deserialize_optional_address<'de, D>(deserializer: D) -> Result<Option<Address>, D::Error>
 where
     D: Deserializer<'de>,
@@ -185,6 +189,10 @@ pub struct InterledgerNode {
     /// Defaults to 60000ms (60 seconds).
     #[serde(default = "default_exchange_rate_poll_interval")]
     pub exchange_rate_poll_interval: u64,
+    /// The number of consecutive failed polls to the exchange rate provider
+    /// that the connector will tolerate before invalidating the exchange rate cache.
+    #[serde(default = "default_exchange_rate_failed_polls_before_invalidation")]
+    pub exchange_rate_failed_polls_before_invalidation: u64,
     /// API to poll for exchange rates. Currently the supported options are:
     /// - [CoinCap](https://docs.coincap.io)
     /// - [CryptoCompare](https://cryptocompare.com) (note this requires an API key)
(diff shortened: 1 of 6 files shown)


crates/ilp-node/src/node.rs
```diff
@@ -183,6 +186,10 @@ pub struct InterledgerNode {
     /// Defaults to 60000ms (60 seconds).
     #[serde(default = "default_exchange_rate_poll_interval")]
     pub exchange_rate_poll_interval: u64,
+    /// The number of consecutive failed polls to the exchange rate provider
+    /// that the connector will tolerate before invalidating the exchange rate cache.
+    #[serde(default = "default_exchange_rate_poll_lenience")]
+    pub exchange_rate_poll_lenience: u32,
```
emschwartz: How about `exchange_rate_poll_failure_tolerance`?

crates/interledger-api/tests/node_settings.rs
```diff
@@ -46,6 +46,7 @@ fn node_settings_test() {
         secret_seed: random_secret(),
         route_broadcast_interval: Some(200),
         exchange_rate_poll_interval: 60000,
+        exchange_rate_failed_polls_before_invalidation: 5,
```
emschwartz: The names of the variable in the tests need to be updated

crates/interledger-service-util/src/exchange_rates_service.rs
```diff
@@ -8,12 +8,14 @@ use interledger_service::*;
 // TODO remove the dependency on interledger_settlement, that doesn't really make sense for this minor import
 use interledger_settlement::{Convert, ConvertDetails};
 use log::{debug, error, trace, warn};
+use parking_lot::RwLock;
```
emschwartz: Can we just use an AtomicU32 instead?

bstrie: Sure, though that wouldn't solve the potential race condition.

emschwartz: I'm not too worried about that particular race condition

crates/interledger-service-util/src/exchange_rates_service.rs
```diff
@@ -270,10 +272,10 @@ where
             .map_err(move |_| {
                 // Note that a race between the read on this line and the check on the line after
                 // is quite unlikely as long as the interval between polls is reasonable.
-                let failed_polls: u32 = { *consecutive_failed_polls.read() };
+                let failed_polls = consecutive_failed_polls.load(Ordering::Relaxed);
```
emschwartz: What about just using [`fetch_add`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU32.html#method.fetch_add) and then checking the return. Whether or not the `failed_polls < failed_polls_before_invalidation`, the number of failed polls has gone up

PR #469, opened 2019-10-26
Don't clone routing table for every packet

emschwartz:
While reviewing the routing table logic with @bstrie, we realized that it was cloning the whole routing table, including account IDs for every account, while processing each packet. This changes the `RouterStore#routing_table()` method to return an `Arc<HashMap>` to avoid copying the underlying data.

This also changes the prefix type to be `String`s instead of `Bytes`, because ILP address prefixes must be UTF8 and we had a lot of unnecessary parsing logic scattered throughout the code to display the prefix.

The first two commits are unrelated but they're small changes I think we should make.

Open questions:
- Should we use the [String crate](https://crates.io/crates/string) instead of `std::string::String` to avoid copying those values around (for example, when creating route broadcasts)?

### DIFF
--- a/crates/ilp-node/src/main.rs
+++ b/crates/ilp-node/src/main.rs
@@ -198,7 +198,7 @@ fn merge_args(config: &mut Config, matches: &ArgMatches) {
 fn get_env_config(prefix: &str) -> Config {
     let mut config = Config::new();
     config
-        .merge(config::Environment::with_prefix(prefix))
+        .merge(config::Environment::with_prefix(prefix).separator("__"))
         .unwrap();
 
     if prefix.to_lowercase() == "ilp" {
--- a/crates/ilp-node/tests/btp.rs
+++ b/crates/ilp-node/tests/btp.rs
@@ -29,7 +29,7 @@ fn two_nodes_btp() {
     let node_b_settlement = get_open_port(Some(3021));
 
     let mut runtime = RuntimeBuilder::new()
-        .panic_handler(|_| panic!("Tokio worker panicked"))
+        .panic_handler(|err| std::panic::resume_unwind(err))
         .build()
         .unwrap();
 
--- a/crates/interledger-api/src/routes/node_settings.rs
+++ b/crates/interledger-api/src/routes/node_settings.rs
@@ -103,7 +103,7 @@ where
         .and_then(|store: S| {
             // Convert the account IDs listed in the routing table
             // to the usernames for the API response
-            let routes = store.routing_table();
+            let routes = store.routing_table().clone();
             store
                 .get_accounts(routes.values().cloned().collect())
                 .map_err::<_, Rejection>(|_| {
@@ -113,15 +113,9 @@ where
                 .and_then(move |accounts| {
                     let routes: HashMap<String, String> = HashMap::from_iter(
                         routes
-                            .keys()
-                            .zip(accounts.into_iter().map(|a| a.username().to_string()))
-                            .filter_map(|(prefix, username)| {
-                                if let Ok(prefix) = str::from_utf8(prefix.as_ref()) {
-                                    Some((prefix.to_string(), username))
-                                } else {
-                                    None
-                                }
-                            }),
+                            .iter()
+                            .map(|(prefix, _)| prefix.to_string())
+                            .zip(accounts.into_iter().map(|a| a.username().to_string())),
                     );
 
                     Ok(warp::reply::json(&routes))
--- a/crates/interledger-ccp/src/server.rs
+++ b/crates/interledger-ccp/src/server.rs
@@ -8,7 +8,6 @@ use crate::{
     routing_table::RoutingTable,
     CcpRoutingAccount, RouteManagerStore, RoutingRelation,
 };
-use bytes::Bytes;
 use futures::{
     future::{err, join_all, ok, Either},
     Future, Stream,
@@ -55,7 +54,7 @@ fn hash(preimage: &[u8; 32]) -> [u8; 32] {
     out
 }
 
-type NewAndWithdrawnRoutes = (Vec<Route>, Vec<Bytes>);
+type NewAndWithdrawnRoutes = (Vec<Route>, Vec<String>);
 
 pub struct CcpRouteManagerBuilder<I, O, S> {
     /// The next request handler that will be used both to pass on requests that are not CCP messages.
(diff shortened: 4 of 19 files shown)


crates/interledger-api/src/routes/node_settings.rs
```diff
@@ -113,15 +113,9 @@ where
                 .and_then(move |accounts| {
                     let routes: HashMap<String, String> = HashMap::from_iter(
                         routes
-                            .keys()
-                            .zip(accounts.into_iter().map(|a| a.username().to_string()))
-                            .filter_map(|(prefix, username)| {
-                                if let Ok(prefix) = str::from_utf8(prefix.as_ref()) {
```
gakonst: Yeah wanted to point this out in the previous PR that it did not make much sense for the prefixes to be `Bytes`, this is a good change

crates/interledger-ccp/src/server.rs
```diff
@@ -338,18 +337,18 @@ where
...
                 } else if route.prefix.len() <= address_scheme.len() + 1 {
                     // note the + 1 is due to address_scheme not including a trailing "."
                     warn!("Got route broadcast for the global prefix: {:?}", route);
                     false
-                } else if route.prefix.starts_with(self.ilp_address.read().as_ref()) {
+                } else if route.prefix.starts_with(&ilp_address as &str) {
                     trace!("Ignoring route broadcast for a prefix that starts with our own address: {:?}", route);
                     false
-                } else if route.path.contains(self.ilp_address.read().as_ref()) {
+                } else if route.path.iter().any(|p| p == &ilp_address as &str) {
```
gakonst: I know you really like iterators & functional style calls, but I'm generally curious, does using `any(<condition>)` have better performance over `contains`?

crates/interledger-ccp/src/server.rs
```diff
@@ -888,24 +896,24 @@ where
...
+    local_routes: &HashMap<String, A>,
+    configured_routes: &HashMap<String, A>,
     incoming_tables: &HashMap<A::AccountId, RoutingTable<A>>,
-    prefix: &[u8],
+    prefix: &str,
 ) -> Option<(A, Route)> {
     // Check if we have a configured route for that specific prefix
     // or any shorter prefix ("example.a.b.c" will match "example.a.b" and "example.a")
     // Note that this logic is duplicated from the Address type. We are not using
     // Addresses here because the prefixes may not be valid ILP addresses ("example." is
```
gakonst: Maybe we should make prefixes not include the `.` and just use `segments()` in this function?

crates/interledger-store-redis/src/store.rs
```diff
@@ -1937,7 +1928,7 @@ type RouteVec = Vec<(String, AccountId)>;
 // TODO replace this with pubsub when async pubsub is added upstream: https://github.com/mitsuhiko/redis-rs/issues/183
 fn update_routes(
     connection: RedisReconnect,
-    routing_table: Arc<RwLock<HashMap<Bytes, AccountId>>>,
+    routing_table: Arc<RwLock<Arc<HashMap<String, AccountId>>>>,
```
gakonst: Passing an `Arc<RwLock<Arc<T>>>` seems a bit weird to me here, what is the benefit of 2 nested Arcs here?

crates/interledger-ccp/src/server.rs
emschwartz: The prefix isn't necessarily a valid ILP address so we can't use the `Address` type

crates/interledger-store-redis/src/store.rs
emschwartz: Good question. The outer `Arc` allows us to keep a reference to the `RwLock` that we can pass around the redis store. 

I added the inner `Arc` so that the [`routing_table`](https://github.com/interledger-rs/interledger-rs/blob/5b10f3edb61c413e7fb930d0093a90e7421ee25e/crates/interledger-store-redis/src/store.rs#L957-L959) method can return an `Arc<HashMap>`. 

I originally tried to implement it so that `routing_table` just returned a `&HashMap` but the compiler didn't allow it because that would mean returning a reference to a temporary variable created in that function (the read lock). We could make that method return a read lock, but I thought that exposed too much of the specific implementation.

Returning an `Arc<HashMap>` allows the caller of the `routing_table` method to keep their copy for as long as they'd like (it will only be dropped when there are no more references to it) while also allowing the store to swap out the contents of its internal `RwLock` when it updates the routes.

Happy to reconsider this setup but if I can convince you that it's the right way to implement it, I'll add some comments to explain this choice for posterity.

crates/ilp-node/src/main.rs
```diff
@@ -198,7 +198,7 @@ fn merge_args(config: &mut Config, matches: &ArgMatches) {
 fn get_env_config(prefix: &str) -> Config {
     let mut config = Config::new();
     config
-        .merge(config::Environment::with_prefix(prefix))
+        .merge(config::Environment::with_prefix(prefix).separator("__"))
```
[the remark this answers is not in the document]
emschwartz: This allows nested config properties to be passed in as env vars by using a double underscore separator. For example, `ILP_PROMETHEUS__BIND_ADDRESS`

crates/ilp-node/tests/btp.rs
```diff
@@ -29,7 +29,7 @@ fn two_nodes_btp() {
     let node_b_settlement = get_open_port(Some(3021));
 
     let mut runtime = RuntimeBuilder::new()
-        .panic_handler(|_| panic!("Tokio worker panicked"))
+        .panic_handler(|err| std::panic::resume_unwind(err))
```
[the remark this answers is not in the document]
emschwartz: That passes through the panic instead of swallowing it like Tokio normally does

crates/interledger-router/src/lib.rs
```diff
@@ -27,5 +26,5 @@ pub trait RouterStore: AccountStore + Clone + Send + Sync + 'static {
     /// keep the routing table in memory and use PubSub or polling to keep it updated.
     /// This ensures that individual packets can be routed without hitting the underlying store.
     // TODO avoid using HashMap because it means it'll be cloned a lot
-    fn routing_table(&self) -> HashMap<Bytes, <Self::Account as Account>::AccountId>;
+    fn routing_table(&self) -> Arc<HashMap<String, <Self::Account as Account>::AccountId>>;
```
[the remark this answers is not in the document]
emschwartz: That comment does need to be updated. The function does return a clone of the Arc, but that is used to avoid cloning the underlying data

crates/ilp-node/src/main.rs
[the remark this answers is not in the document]
dora-gt: Oh wow, it works even for Option values? nice.
Thank you.

PR #471, opened 2019-10-28
ci(ci): publish binaries when tags are pushed

dora-gt:
**This branch is based on [`feature/dx/testable-md-integrating-into-ci`](https://github.com/interledger-rs/interledger-rs/pull/305)**

image

image

- Publishes binaries for two platforms
    - General Linux OSs using `musl`
    - macOS using [`osxcross`](https://github.com/tpoechtrager/osxcross)
- Binaries are built only when something is tagged with name `ilp-node-.*` or `ilp-cli-.*`.
- Release notes are auto-generated by [conventional-changelog](https://www.npmjs.com/package/conventional-changelog-cli).
- It seems that distributing macOS binaries is no problem but I'm not sure that publishing the docker image using `osxcross` ([interledgerrs/circleci-rust-mac](https://hub.docker.com/r/interledgerrs/circleci-rust-mac)) is legitimate. So I'm contacting Apple whether it is OK or not.
    - We could make it `private` repo but that needs credits on DockerHub.
    - Also we could use [`macOS executor`](https://circleci.com/docs/2.0/executor-types/#using-macos) on CircleCI but it also requires credits on CircleCI.

The workflow is like:
image
which means
- `test-md` is not required to publish artifacts because it should be tested on their branch before merging into master or tagged.
- Buiding binaries jobs and building docker image job are processed in parallel.


dora-gt:
A collaborator of `cargo release` kindly answered the question.
https://github.com/sunng87/cargo-release/issues/172

He suggested that using `release.toml` in `ilp-node` would solve the problem. We could tag the `ilp-node` crate with just a tag `vX.X.X`, setting `tag-prefix` empty. The toml would look like
```
tag-prefix = ""
```
This seems to work well. If there are no objections, I would like to do so.

bstrie:
Very cool to see conventional-changelog in action. Sadly I don't feel qualified to review since I have no experience with CI or release distribution.

emschwartz:
> I'm not sure that publishing the docker image using osxcross (interledgerrs/circleci-rust-mac) is legitimate

@dora-gt I don't think we should publish docker images for mac. There should be at most 1 docker image per binary we're releasing and it should use the most minimal linux distro for it.

dora-gt:
Sorry, my bad. It seems already possible to use `macOS executor`, maybe because we are already paying. I'll try to fix.

dora-gt:
I'm merging this (into `feature/dx/testable-md-integrating-into-ci`) because I could use `macOS executor` of CircleCI and now we don't need to use any docker images to build macOS binaries.

PR #476, opened 2019-10-28
STREAM Receiver: fulfill packets meant for the JS stream server

emschwartz:
Before putting interledger.rs into production, we want to run it side by side with the `ilp-connector` and `ilp-spsp-server` (Javascript implementation) and compare the results. 

This PR adds a feature that makes the Rust STREAM receiver try using the same method used by the JS receiver to generate the shared secret from the incoming packets. When enabled, this allows the Rust receiver to fulfill packets that were created with the details generated by the JS STREAM server.

I also modified the magic bytes used by the receiver to generate the shared secret to match the JS implementation. This is not generally necessary because it only matters that the magic bytes are the same between the server that generates the STREAM details and the one that fulfills the packet (normally the same implementation). We could use a different magic bytes string for the JS method as we do for the normal Rust flow, but it doesn't break anything to change this to match the JS implementation.

Note we may not want to keep this feature in for the long-term but I think it's reasonable to include in the short term.


crates/interledger-stream/src/server.rs
```diff
@@ -70,19 +77,33 @@ impl ConnectionGenerator {
...
-        if local_part.len() == 32 {
-            let (random_bytes, auth_tag) = local_part.split_at(18);
-            let shared_secret = hmac_sha256(&self.secret_generator[..], &random_bytes[..]);
-            let dest: &[u8] = destination_account.as_ref();
-            let derived_auth_tag = &hmac_sha256(&shared_secret[..], &dest[..dest.len() - 19])[..14];
-            if derived_auth_tag == auth_tag {
-                return Ok(shared_secret);
+        if let Ok(local_part) = base64::decode_config(local_part, base64::URL_SAFE_NO_PAD) {
+            if local_part.len() == 32 {
+                let (random_bytes, auth_tag) = local_part.split_at(18);
```
gakonst: These are not really random bytes, are they?

crates/interledger-stream/src/server.rs
```diff
@@ -70,19 +77,33 @@ impl ConnectionGenerator {
...
+                let shared_secret = hmac_sha256(&self.secret_generator[..], &random_bytes[..]);
+                let dest: &[u8] = destination_account.as_ref();
+                let derived_auth_tag =
+                    &hmac_sha256(&shared_secret[..], &dest[..dest.len() - 19])[..14];
+                if derived_auth_tag == auth_tag {
+                    return Ok(shared_secret);
+                }
             }
         }
+        #[cfg(feature = "fulfill-packets-for-js-receiver")]
```
gakonst: Why don't we just always generate the shared secret the way the js receiver does it anyway, instead of doing it both ways?

crates/interledger-stream/src/server.rs
emschwartz: It's a nonce generated by the sender

You're right that "random bytes" probably isn't the best way to describe it

crates/interledger-stream/src/server.rs
emschwartz: Implemented in https://github.com/interledger-rs/interledger-rs/pull/476/commits/83ccf456b96cef00d8cdeeb61178672a8554dc5d

emschwartz: Implemented in https://github.com/interledger-rs/interledger-rs/pull/476/commits/83ccf456b96cef00d8cdeeb61178672a8554dc5d

crates/interledger-stream/src/server.rs
```diff
@@ -41,49 +47,39 @@ impl ConnectionGenerator {
...
-            let (random_bytes, auth_tag) = local_part.split_at(18);
-            let shared_secret = hmac_sha256(&self.secret_generator[..], &random_bytes[..]);
-            let dest: &[u8] = destination_account.as_ref();
-            let derived_auth_tag = &hmac_sha256(&shared_secret[..], &dest[..dest.len() - 19])[..14];
-            if derived_auth_tag == auth_tag {
-                return Ok(shared_secret);
-            }
-        }
-        Err(())
+        // Note this computes the HMAC with the token _encoded as UTF8_,
```
gakonst:
```suggestion
        // Note this computes the HMAC with the token _encoded as UTF8,
```

crates/interledger-stream/src/server.rs
```diff
@@ -167,25 +159,33 @@ where
...
+                match response {
+                    Ok(ref _fulfill) => store.publish_payment_notification(PaymentNotification {
+                        to_username,
+                        from_username,
+                        amount,
+                        destination: destination.clone(),
+                        timestamp: DateTime::<Utc>::from(SystemTime::now()).to_rfc3339(),
+                    }),
+                    Err(ref reject) => {
+                        if reject.code() == ErrorCode::F06_UNEXPECTED_PAYMENT {
```
gakonst: Can't you do `map_err(|reject| { ... }` and handle the error there in the same way it was written before, since you're not avoiding using `result(response)` in either case?

crates/interledger-stream/src/server.rs
```diff
@@ -486,9 +482,36 @@ mod receiving_money {
         let shared_secret = connection_generator
             .rederive_secret(&prepare.destination())
             .unwrap();
-        let result = receive_money(&shared_secret, &ilp_address, "ABC", 9, prepare);
+        let result = receive_money(&shared_secret, &ilp_address, "ABC", 9, &prepare);
         assert!(result.is_err());
     }
+
+    #[test]
+    fn fulfills_packets_sent_to_javascript_receiver() {
```
gakonst: :100:

crates/interledger-stream/src/server.rs
```diff
@@ -210,17 +210,27 @@ fn receive_money(
...
-            }
-            .build()
-        })?;
+
+    // Note that we are copying the Prepare packet data. This is a bad idea
+    // in cases where STREAM is used to send a significant amount of data.
+    // This implementation doesn't currently support handling the STREAM data
+    // so copying the bytes of the other STREAM frames shouldn't be a big
+    // performance hit in practice.
+    // The data is copied so that we can take the Prepare packet by
```
gakonst: Doesn't this mean that we will copy each time? Wouldn't it be better if we left this function as is, and just clone inside `send_request`?

crates/interledger-stream/src/server.rs
emschwartz: The underscores are intentional to emphasize "encoded as UTF8"

gakonst: nvm did not see the previous underscore

crates/interledger-stream/src/server.rs
emschwartz: That is worse because we need to clone the `next` service in order to pass it into the future combinator

PR #478, opened 2019-10-29
Various exchange rate fixes

emschwartz:
See the commit messages for details

### DIFF
--- a/crates/ilp-cli/Cargo.toml
+++ b/crates/ilp-cli/Cargo.toml
@@ -13,5 +13,5 @@ http = { version = "0.1.18", default-features = false }
 reqwest = { version = "0.9.21", default-features = false, features = ["default-tls"] }
 serde = { version = "1.0.101", default-features = false }
 serde_json = { version = "1.0.41", default-features = false }
-tungstenite = { version = "0.9.1", default-features = true }
+tungstenite = { version = "0.9.1", default-features = false, features = ["tls"] }
 url = { version = "2.1.0", default-features = false }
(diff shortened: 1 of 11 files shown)


crates/ilp-cli/Cargo.toml
```diff
@@ -13,5 +13,5 @@ http = { version = "0.1.18", default-features = false }
 reqwest = { version = "0.9.21", default-features = false, features = ["default-tls"] }
 serde = { version = "1.0.101", default-features = false }
 serde_json = { version = "1.0.41", default-features = false }
-tungstenite = { version = "0.9.1", default-features = false }
+tungstenite = { version = "0.9.1", default-features = true }
```
bstrie: For consistency I would like to leave this as false and add `features = ["tls"]` instead; that is indeed the only feature that tungstenite exposes at the moment and so this is currently semantically equivalent, but part of my (ill-communicated) impetus for the mass `default-features = false` is to prevent automatically expanding our dependencies' surface area without our explicit consideration, in the event that a future version of a dependency adds a new default feature. Does this sound like a reasonable policy? Sadly I can't exactly say that this approach is a widely-used precedent; a lot of the community-wide best practices regarding dependencies are still up in the air.

crates/interledger-btp/Cargo.toml
```diff
@@ -22,7 +22,7 @@ rand = { version = "0.7.2", default-features = false, features = ["std"] }
 stream-cancel = { version = "0.4.4", default-features = false }
 tokio-executor = { version = "0.1.8", default-features = false }
 tokio-timer = { version = "0.2.11", default-features = false }
-tokio-tungstenite = { version = "0.9.0", default-features = false, features = ["connect", "tls"] }
+tokio-tungstenite = { version = "0.9.0", default-features = true }
```
bstrie: Related to the previous comment.

PR #486, opened 2019-10-31
feat(http api): return the node version instead of the interledger-api crate version

dora-gt:
We could do some hack when compiling using `build.rs` etc. but it seemed more straightforward to pass the version to API. I'm not so sure there is the case of using `interledger-api` crate for another use but I used `Option` for `node_version`. If that looks too much or unnecessary, I'd be happy to change.

The output looks like:
```
{"ilp_address":"local.host","status":"Ready","version":"0.5.1"}
```
which shows `ilp-node`'s version.

### DIFF
--- a/crates/ilp-node/src/node.rs
+++ b/crates/ilp-node/src/node.rs
@@ -404,7 +404,7 @@ impl InterledgerNode {
                                 api.default_spsp_account(username);
                             }
                             // add an API of ILP over HTTP and add rejection handler
-                            let api = api.into_warp_filter()
+                            let api = api.into_warp_filter(Some(env!("CARGO_PKG_VERSION").to_owned()))
                                 .or(IlpOverHttpServer::new(incoming_service.clone().wrap(|request, mut next| {
                                     let http = debug_span!(target: "interledger-node", "http");
                                     let _http_scope = http.enter();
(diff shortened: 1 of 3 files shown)


gakonst:
I am not sure about this, although it's a small change, it seems weird to me to require a version string when calling a function. @bstrie is there a better way to do this?

bstrie:
I suppose it's not that weird, considering that ilp-node will have to pass this information to interledger-api one way or another. One alternative could be to pass the information along in environment variables. Such an approach could potentially be useful in the future, as I remember one of us being wary of exposing any kind of server version information, since it could be used by attackers. If this were passed in via environment variables, then the server operator could choose to set `ILP_NODE_VERSION="some hash of the real version that is useless to an attacker"`, and ilp-node could check for the existence of such a variable and, if the variable is unset, simply pass along its own version by default. Does that strike anyone as useful?

dora-gt:
Actually I tried to inject an env var using `build.rs` but it didn't work because the `build.rs` of `ilp-node` is not run before one of `interledger-api`.

So, what I could do then was to run the `build.rs` of `interledger-api` and let it get the version of `ilp-node`, which seemed weird to me because `interledger-api` should NOT know about upper concepts like `ilp-node`.

Also, it seemed reasonable for me to inject some dependencies when creating the API instance although the "dependency" is just "a" version here.

bstrie:
I'd say this is totally fine, there's no reason we couldn't revisit this later if we wanted to.

PR #487, opened 2019-10-31
Node logging fixes

emschwartz:


crates/ilp-node/src/node.rs
```diff
@@ -269,11 +269,17 @@ impl InterledgerNode {
                 .and_then(move |btp_accounts| {
                     let outgoing_service =
                         outgoing_service_fn(move |request: OutgoingRequest<Account>| {
-                            error!(target: "interledger-node", "No route found for outgoing account ");
+                            // Don't log anything for failed route updates sent to child accounts
+                            // because there's a good chance they'll be offline
+                            if request.prepare.destination().scheme() != "peer"
```
dora-gt: I don't think this is required right now but this could be encapsulated and more understandable like
```
request.prepare.destination().is_scheme_peer()
// or
request.prepare.destination().is_scheme(Scheme::Peer)
```

crates/ilp-node/src/trace.rs
```diff
@@ -88,14 +92,26 @@ pub fn trace_outgoing<A: Account>(
...
     let _details_scope = details_span.enter();
 
+    // Don't log anything for failed route updates sent to child accounts
+    // because there's a good chance they'll be offline
+    let ignore_rejects = request.prepare.destination().scheme() == "peer"
+        && request.to.routing_relation() == RoutingRelation::Child;
     next.send_request(request)
-        .then(trace_response)
+        .then(move |result| {
+            if result.is_ok() || ignore_rejects {
```
dora-gt:
```suggestion
            if result.is_ok() || ! ignore_rejects {
```
Is this...?

dora-gt: Also I feel this should be more specific like... "If the node could not **connect** to the child". It seems that this code includes just pure rejections that could be sent from **connected** children.

emschwartz: Updated to make it only ignore F02 errors

PR #491, opened 2019-11-01
refactor(three_nodes test, .circleci/config.yml): fixed account usernames to be clearer, etc

dora-gt:
To investigate more about #490, fixed some lines to output more detailed information.
I could not reproduce errors on my computer, so I need more information in the CI process.

### DIFF
--- a/crates/ilp-node/tests/three_nodes.rs
+++ b/crates/ilp-node/tests/three_nodes.rs
@@ -1,8 +1,8 @@
-use futures::{future::join_all, Future};
+use futures::{future::join_all, stream::*, sync::mpsc, Future};
 use ilp_node::InterledgerNode;
 use serde_json::json;
 use tokio::runtime::Builder as RuntimeBuilder;
-use tracing::error_span;
+use tracing::{debug, error_span};
 use tracing_futures::Instrument;
 
 mod redis_helpers;
@@ -13,6 +13,8 @@ use interledger::packet::Address;
 use interledger::stream::StreamDelivery;
 use std::str::FromStr;
 use test_helpers::*;
+
+const LOG_TARGET: &str = "interledger-tests-three-nodes";
 
 #[test]
 fn three_nodes() {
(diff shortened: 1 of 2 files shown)


crates/ilp-node/tests/three_nodes.rs
```diff
@@ -1,5 +1,6 @@
-use futures::{future::join_all, Future};
+use futures::{future::join_all, stream::*, sync::mpsc, Future};
 use ilp_node::InterledgerNode;
+use log::debug;
```
emschwartz: You should just use `tracing::debug` instead. It's functionality is a superset of `log`'s

crates/ilp-node/tests/three_nodes.rs
```diff
@@ -42,35 +45,35 @@ fn three_nodes() {
 
     let alice_on_alice = json!({
         "ilp_address": "example.alice",
-        "username": "alice",
+        "username": "alice_on_alice",
```
emschwartz: I think it's even clearer if the nodes are identified as `a` and `b` so it's `alice_on_a`, etc

crates/ilp-node/tests/three_nodes.rs
```diff
@@ -137,16 +139,24 @@ fn three_nodes() {
...
 
+    let mut node1_finish_sender = finish_sender.clone();
     runtime.spawn(
         node1
             .serve()
             .and_then(move |_| alice_fut)
-            .and_then(|_| Ok(()))
+            .and_then(move |_| {
+                node1_finish_sender
+                    .try_send(1)
```
emschwartz: What is this for?

crates/ilp-node/tests/three_nodes.rs
dora-gt: Indeed. I'll fix.

crates/ilp-node/tests/three_nodes.rs
dora-gt: I'm not sure this has positive effects but it just ensures that tasks of `node3` starts after `node1` and `node2` finished their tasks. It seemed more robust than waiting `1000ms` or something though it looks actually almost the same.

crates/ilp-node/tests/three_nodes.rs
dora-gt: Fixed.

crates/ilp-node/tests/three_nodes.rs
dora-gt: Fixed.

dora-gt:
I'm merging this.
I'll check logs if the tests fail again.

PR #492, opened 2019-11-04
ci(circleci): enables run-md-test.sh on CircleCI

dora-gt:
[Separated PR](https://github.com/interledger-rs/interledger-rs/pull/305#pullrequestreview-310992490) of #305 "Enables `run-md-test.sh` on CircleCI". The following PRs should be merged at once, because latter PRs optimize some of former PRs (e.g. cache).

- #492 ci(circleci): enables run-md-test.sh on CircleCI
- #493 feat(examples): fixed to use CLI arguments instead of ENV vars, etc
- #494 feat(.circleci): tag docker images
- #495 ci(ci): publish binaries when tags are pushed

Note that some of these will be simplified when examples start using binaries. Some of docker images will be removed, examples get simpler.

<details><summary>My opinions</summary>

- Why do we need `circleci-rust-dind.dockerfile`?
    - It is because
        1. Installing dependencies every time doesn't make sense
        1. there aren't any Docker images that contain Rust, musl with SSL support and Docker commands.
- Why are there so many changes?
    - It is because we are
        1. Testing markdowns
        1. Building Docker images, tag it and publish it
        1. Building binaries
        1. Auto-releasing the artifacts
    - and we definitely should do CI/CI. We should not do these manually because people forget to test markdowns, which makes always examples broken and forget to publish Docker images and binaries. That is also cumbersome. Why don't we just automate it?

</details>

PR #493, opened 2019-11-04
feat(examples): fixed to use CLI arguments instead of ENV vars, etc

dora-gt:

    - It is because
    - It is because we are
        1. Testing markdowns
        1. Building binaries

</details>

### DIFF
--- a/docker/CI/README.md
+++ b/docker/CI/README.md
@@ -18,7 +18,7 @@ In order to build new docker images, we build the source code once like `cargo b
 
 The docker images `node.dockerfile` and `ilp-cli.dockerfile` contain platform-agnostic (although it is not completely agnostic) binaries which means it does not rely on dynamic libraries and uses `musl` instead. So we have to provide `musl` environment for `Cargo`. That is why `circleci-rust-dind.dockerfile` contains `musl`.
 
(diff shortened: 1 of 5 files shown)


docker/CI/README.md
```diff
@@ -18,7 +18,7 @@ In order to build new docker images, we build the source code once like `cargo b
...
 
-Also why `circleci-rust-dind.dockerfile` is an independent image is that there isn't any suitable image which contains `Cargo`, `musl` and `dind` (docker in docker). `dind` is required to test the Docker mode because `run-md.sh` runs docker commands such as `docker run`. It requires docker commands inside the docker image. CircleCI spins up a Docker environment somewhere outer the container. So it would be called rater "docker command in container" than "docker in docker".
+Also why `circleci-rust-dind.dockerfile` is an independent image is that there isn't any suitable image which contains `Cargo`, `musl` and `dind` (docker in docker). `dind` is required to test the Docker mode because `run-md.sh` runs docker commands such as `docker run`. It requires docker commands inside the docker image. CircleCI spins up a Docker environment somewhere outside the container. So it would be called rater "docker command in container" than "docker in docker".
```
gakonst:
```suggestion
Currently, there exists no Docker image which contains `Cargo`, `musl` and `dind` together. `dind` is required to test Docker mode because `run-md.sh` runs the tests using Docker inside CircleCI (hence requiring us to use "docker in docker")
```

gakonst:
@dora-gt can we perhaps base this on master and merge it? This looks good to merge now and is not dependent on any of the docker stuff

dora-gt:

dora-gt:
Oops, I didn't notice your comment. I'm sorry. Maybe we could merge #506 instead.
Does that work for you?

PR #494, opened 2019-11-04
feat(.circleci): tag docker images

dora-gt:

    - It is because
    - It is because we are
        1. Testing markdowns
        1. Building binaries

</details>


.circleci/README.md
```diff
@@ -14,7 +14,7 @@ So we need new docker images of:
...
+In order to build new docker images, we build the source code once like `cargo build --bin ilp-node --bin ilp-cli`, not twice because we want to reduce compilation time. After the build, we just copy the binaries to docker containers using `.circleci/node.dockerimage` and `.circleci/ilp-cli.dockerimage`. That is why we don't use normal `node.dockerfile` and `ilp-cli.dockerfile`. It compiles the source files of `interledger-rs` **inside** the container, so we need **two** compilation of the source although that seems natural for normal uses.
 
 The docker images `node.dockerfile` and `ilp-cli.dockerfile` contain platform-agnostic (although it is not completely agnostic) binaries which means it does not rely on dynamic libraries and uses `musl` instead. So we have to provide `musl` environment for `Cargo`. That is why `circleci-rust-dind.dockerfile` contains `musl`.
```
gakonst: I seems a little odd to say "platform-agnostic, but not really" here? not sure what would be a better way to phrase it

.circleci/config.yml
```diff
@@ -83,9 +83,15 @@ jobs:
           docker_layer_caching: true
       - checkout
       # If something wrong is cached, use these lines
-      #- run:
-      #    name: Clear docker cache
-      #    command: docker images --no-trunc --format '{{.ID}}' interledgerrs/settlement-engines | xargs --no-run-if-empty docker rmi
+#      - run:
```
gakonst: Think indentation is off by 1 to the right here

.circleci/config.yml
```diff
@@ -83,9 +83,15 @@ jobs:
...
+#          name: Clear docker cache
+#          command: docker images --no-trunc --format '{{.ID}}' interledgerrs/settlement-engines | xargs --no-run-if-empty docker rmi
+      - run:
+          # Although we use `docker_layer_caching`, we need to pull these latest images
```
gakonst: Not a docker expert here, if we're clearing the cache and then pulling, can't we remove both steps and use the cache? Why do we need to pull the images if we use the caching feature?

.circleci/config.yml
```diff
@@ -155,13 +172,24 @@ workflows:
...
+      - build:
+          filters:
+            tags:
+              only: # we need this to kick builds when tags are given
+                - /.*/
+      - test-md:
+          filters:
+            tags:
+              only: # we need this to kick builds when tags are given
+                - /.*/
```
gakonst: Will this only build & test if we give a tag? If I'm opening a PR, it won't be tagged, doesn't this mean that tests won't run for it?

.circleci/README.md
dora-gt: I think it is just OK because this markdown is going to be removed soon.
If we use binaries for examples, we no longer need these tricks.

.circleci/config.yml
dora-gt: I'm not sure but if I remove `#`, it looks just good. 🤔

comment-out

Different on your IDE?

.circleci/config.yml
dora-gt: Imagine that docker images consist of some layers.

1. base files
1. some additional files
1. some more additional files

The above has 3 layers for example.
If we disable whole caching, we always download the three layers. It is a waste of time. So I used the cache.

**However**, it seems that CircleCI caches `latest` image and it does not try to download the truly `latest` image. In other words, OLD `latest` is cached. So we have to `pull` the `latest` image. That said, if the cache is enabled, it only downloads **not cached layers**.

For example,

1. base files
1. some additional files
1. some more additional files
1. some **latest** change

In this case, it only downloads layer **4**. It seems reasonable.

.circleci/config.yml
dora-gt: It only builds if some tags are given. This is something tricky on CircleCI.

- CircleCI builds basically ALL the pushed branches **even if no `branches` section is given**.
- CircleCI does **NOT** build pushed tags if there is no `tags` section.
- Giving `tags` section does not affect pushed branches.

🤷‍♂️

dora-gt:

Issue #230, opened 2019-08-19
Auto-publish artifacts when new versions are tagged


emschwartz:
We should start tagging versions of the components of this project. This includes:
- Docker image
- `interledger` crate
- `interledger-*` crates

It seems like the version tags on the repo should line up with the Docker image and the `interledger` crate.

I would propose that we version the sub-crates separately and let their versions be whatever they may be.

Thoughts? @gakonst @dora-gt @bstrie

dora-gt:
I understand the version policy. Here, we should consider:

1. The versioning policy
    - Bump up the ONLY `ilp-node` crate; the other crates are not affected.
1. The release policy
    - Should we follow the Rust's release policy?
        - `Nightly` builds
        - `Beta` releases
        - `Stable` releases
1. The artifacts
    - Binaries
        - Should we provide binaries? for what platforms? how?
    - Docker Images
        - We should properly tag the image versions. I don't think this is a big issue. Just using the `ilp-node`'s crate version would be appropriate.
    - Documents
        - We should NOT use `nightly` builds for documents because it might cause errors. We should use `stable` (or at least `beta`) releases for documents.

So the questions are:
1. Should we follow the Rust's release policy (and its branching model)?
1. How, for what platform, should we provide binaries?
    - GitHub's assets
    - interledger.rs site

What do you think? @interledger-rs/contributors

dora-gt:
My thoughts.

1. Should we follow the Rust's release policy (and its branching model)?

Maybe YES, we could also adopt some other strategies like [A successful Git branching model](https://nvie.com/posts/a-successful-git-branching-model/) but it seems that both work (IMO) and we live in the Rust world, it seems natural to adopt Rust's releasing model if there isn't any problem.

JFYI: the main difference AFAIK between "A successful Git branching model" and Rust's one is that "A successful Git branching model" expects that the master is always `stable`. The commits should be tested and should have worked a while. In the Rust's model, it seems that the `master` branch is `unstable` and being continuously tested.

2. How, for what platforms, should we provide binaries?

My answer is:

- What platforms
    - Linux based OSs (`musl`)
    - macOS
- How
    - GitHub assets.
    - Additionally, package managers could be considered. e.g. `brew`
    - If we had the `interledger-rs` website(#378), that would be another option.

I think we should focus on what is really needed or used. I considered:

- Most developers seem to be using Linux based OS. Then providing `musl` binaries would suffice.
[...]

PR #495, opened 2019-11-04
ci(ci): publish binaries when tags are pushed

dora-gt:

    - It is because
    - It is because we are
        1. Testing markdowns
        1. Building binaries

</details>


.circleci/config.yml
```diff
@@ -108,45 +64,37 @@ jobs:
...
           path: /tmp/run-md-test
           destination: run-md-test
-      - save_cache:
-          key: cargo-cache-2-{{ checksum "Cargo.lock" }}-{{ .Branch }}
-          paths:
-            - /home/circleci/.cargo
   update-docker-images:
     docker:
       - image: interledgerrs/circleci-rust-dind:latest
+    resource_class: medium+
```
gakonst: @emschwartz guess @dora-gt was running into the out of memory issues, tagging you to make sure you know that we'll be using bigger boxes if this gets merged

.circleci/config.yml
```diff
@@ -155,22 +103,118 @@ jobs:
...
+    steps:
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - checkout
+      - run:
+          # Builds release profile binaries
+          # The artifact is stored in release-builds directory, which is persistent
+          name: Build binary
+          command: |
+            crate_name=$(./.circleci/release/get_crate_name.sh ${CIRCLE_TAG})
```
gakonst: nit, I think it'd be better to use 
```
`pwd`/.circleci/release/...
```
 instead of `./`

.circleci/config.yml
```diff
@@ -155,22 +103,118 @@ jobs:
...
+  build-release-binary-darwin:
+    macos:
+      xcode: "10.2.0"
+    resource_class: medium+
+    steps:
+      - checkout
+      - run:
+          name: Install Rust
+          command: |
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.38.0
```
gakonst: If we switch the repo to a newer rust, Mac compilation will probably fail (writing this so that we don't forget to update this when we update to 1.39)

.circleci/release/get_crate_name.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# Returns a crate name to build from a git tag name.
+# The tag name is assumed to be output by `cargo release` or be tagged manually.
+
+if [[ $1 =~ ^(.*)-v([0-9]+)\.([0-9]+)\.([0-9]+)(-([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?(\+[0-9A-Za-z-]+)?$ ]]; then
```
gakonst: could you add a comment on what this regex does?

scripts/release.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+# Just not to publish crates
```
gakonst: Don't we also want to publish our subcrates as standalone crates?

.circleci/config.yml
dora-gt: What is the pro of this?

.circleci/config.yml
dora-gt: Linux builds also fail but it is the desired behavior. I believe we should upgrade the version explicitly. I'd be happy to upgrade the docker images if we change the Rust version.

dora-gt: Added comment:
https://github.com/interledger-rs/interledger-rs/pull/495/files#diff-c737838ab35a25093f7f2489f4e420e3R1-R5

.circleci/release/get_crate_name.sh
dora-gt: It is `(something)-(semantic version)`. I'll add comments.

scripts/release.sh
dora-gt: As discussed here:https://github.com/interledger-rs/interledger-rs/issues/230
It seems better to push crates by ourselves because `cargo release` publishes crates even if nothing is changed.

dora-gt:

PR #496, opened 2019-11-05
Use binaries for examples

dora-gt:
This PR is based on #495. #495 should be merged first.

Fix examples to use compiled binaries.

- Make examples use compiled binaries
    - Also source code mode is provided
- Remove Docker mode markdown test
    - Binary mode and source code mode are tested
- Remove unnecessary Docker images
- Refactor CI
- Fix three nodes example to use `ilp-cli`

Release names have to be renamed by `update_markdown_bin.sh`. Currently pseudo release names are used.

I'll try not to use `circleci-rust-dind.dockerfile` in another PR.


examples/eth-settlement/README.md
```diff
@@ -105,73 +68,119 @@ init
...
+<!--!
+    if [ $(is_macos) -eq 1 ]; then
 -->
 ```bash
-cargo build --bin ilp-node
+pushd ~/.interledger/bin &>/dev/null
+
+# install ilp-node
+ILP_NODE_RELEASE_NAME=ilp-node-tag-sample
+if [ ! -e "ilp-node" ]; then
```
gakonst: When running the examples on CI, the binaries will be stored under `./target/debug/xxx`, maybe we should allow providing the path where all the binaries are as an environment variable? i.e. `if [ ! -e $ILP_BIN/ilp-node ]`

examples/eth-settlement/README.md
```diff
@@ -105,73 +68,119 @@ init
...
+If you would prefer compiling from the source code, compile interledger.rs and CLI as follows.
+
+<!--!
+if [ ${SOURCE_MODE} -eq 1 ]; then
+    printf "Installing interledger.rs... (This may take a couple of minutes)\n"
+-->
+```bash
+# just make sure you are running this in the top directory of interledger.rs
+pushd $(dirname $(cargo locate-project | grep -Eo '/(.*)Cargo.toml')) &>/dev/null
+if [ -z "$(which ilp-node)" ]; then
```
gakonst: https://github.com/interledger-rs/interledger-rs/pull/496/files#r343806910 same comment here

examples/eth-settlement/README.md
```diff
@@ -105,73 +68,119 @@ init
...
+#### Compile from the Source Code
+If you would prefer compiling from the source code, compile interledger.rs and CLI as follows.
+
+<!--!
+if [ ${SOURCE_MODE} -eq 1 ]; then
+    printf "Installing interledger.rs... (This may take a couple of minutes)\n"
+-->
+```bash
+# just make sure you are running this in the top directory of interledger.rs
+pushd $(dirname $(cargo locate-project | grep -Eo '/(.*)Cargo.toml')) &>/dev/null
```
gakonst: Can't you   just do `pushd $pwd`?

examples/eth-settlement/README.md
```diff
@@ -105,73 +68,115 @@ init
...
+if [ ! -e "ilp-node" ]; then
+    curl -L https://github.com/interledger-rs/interledger-rs/releases/download/${ILP_NODE_RELEASE_NAME}/ilp-node-x86_64-unknown-linux-musl.tar.gz | tar xzv
+fi
 
-### 1. Build interledger.rs
-First of all, let's build interledger.rs. (This may take a couple of minutes)
+# install ilp-cli
+ILP_CLI_RELEASE_NAME=ilp-cli-tag-sample
+if [ ! -e "ilp-cli" ]; then
+    curl -L https://github.com/interledger-rs/interledger-rs/releases/download/${ILP_CLI_RELEASE_NAME}/ilp-cli-x86_64-unknown-linux-musl.tar.gz | tar xzv
```
gakonst: I think we can remove the ILP_CLI_RELEASE_NAME in the examples and just have a `latest` tag which is always used and points to the latest version.

scripts/update_markdown_bin.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
gakonst: if we use the `latest` tag as written in my previous comment then this should be unneeded

examples/eth-settlement/README.md
dora-gt: It is the binary mode, it downloads the latest **compiled** binary.

examples/eth-settlement/README.md
dora-gt: On the other hand, this is **source code mode**. I changed this line to use `alias`.

examples/eth-settlement/README.md
dora-gt: I can't because it might be in `interledger-rs` or `interledger-rs/examples/eth-settlement` but I changed this line, so it no longer matters now.

examples/eth-settlement/README.md
dora-gt: It would be nice and easier!
I'll try to fix so. Thank you.

scripts/update_markdown_bin.sh
dora-gt: Indeed. Thank you so much.

dora-gt:

PR #497, opened 2019-11-05
Refactor ILP Settlement into Settlement Core

gakonst:
Description & rationale TBD. There's some code across helpers which is duplicated and ideally I'd like to remove as well.

Basically I wanted to package all settlement related functionalities in the `interledger_settlement` crate. If you're an engine implementer, you'd just import that (w/o the `settlement_api` feature), and use the utilities from settlement_core.

One might think that it'd have been better that the Engine related functionalities, e.g. the trait and the common backends should go to https://github.com/interledger-rs/settlement-engines, but my thinking here was that somebody perhaps would want to have their own implementation and not be bound to the settlement-engines crate (which could be considered as the "(un)official" engines repository.


crates/interledger-settlement/src/settlement_core/api.rs
```diff
@@ -0,0 +1,376 @@
+/// # Settlement Engine API
```
gakonst: This file is moved over from the settlement-engines repository.

crates/interledger-settlement/src/settlement_core/backends_common/redis/mod.rs
```diff
@@ -0,0 +1,352 @@
+use bytes::Bytes;
```
gakonst: This is moved over from the engines repo. It is a duplicate implementation of the IdempotentStore for the RedisStore for the node.

crates/interledger-settlement/src/settlement_core/types.rs
```diff
@@ -0,0 +1,372 @@
+use bytes::Bytes;
```
gakonst: These are all the settlement related types in one file for better organization

emschwartz:
@gakonst what do you think about moving all settlement-related stuff into the other repo, renaming that repo to just `settlement`, and having separate crates in that repo for:
- `settlement-engine-core`
- `settlement-api` (this could include both the connector and SE sides of the API with feature flags to turn one off if you don't want to import it)
- `settlement-eth`, etc

gakonst:
That was my initial idea and is implemented in https://github.com/interledger-rs/settlement-engines/tree/settlement-monorepo. The problem there was that functions which required a `SettlementAccount` did not work due to a version conflict between the `interleger_service` imported as a relative path dependency, and as an external dependency as version `0.3.0` in the other repo.

gakonst:
@emschwartz as per our discussion, we leave this as is, and feature-gate the `backends_common` module

crates/interledger-api/src/lib.rs
```diff
@@ -5,7 +5,7 @@ use interledger_packet::Address;
 use interledger_router::RouterStore;
 use interledger_service::{Account, AddressStore, IncomingService, OutgoingService, Username};
 use interledger_service_util::{BalanceStore, ExchangeRateStore};
-use interledger_settlement::{SettlementAccount, SettlementStore};
+use interledger_settlement::settlement_core::types::{SettlementAccount, SettlementStore};
```
emschwartz:
```suggestion
use interledger_settlement::core::types::{SettlementAccount, SettlementStore};
```

crates/interledger-settlement/Cargo.toml
```diff
@@ -11,7 +11,7 @@ repository = "https://github.com/interledger-rs/interledger-rs"
 bytes = { version = "0.4.12", default-features = false }
 futures = { version = "0.1.29", default-features = false }
 hyper = { version = "0.12.35", default-features = false }
-interledger-http = { path = "../interledger-http", version = "^0.3.0", default-features = false, features = ["idempotency"] }
+interledger-http = { path = "../interledger-http", version = "^0.3.0", default-features = false }
```
emschwartz: Does this crate needs to depend on `interledger-http`?

crates/interledger-settlement/src/settlement_core/engines_api.rs
```diff
@@ -0,0 +1,376 @@
+/// # Settlement Engine API
+///
+/// Web service which exposes settlement related endpoints as described in RFC536,
+/// See [forum discussion](https://forum.interledger.org/t/settlement-architecture/545) for more context.
```
emschwartz: The PR has been merged so this should point to the actual RFC

crates/interledger-settlement/src/settlement_core/engines_api.rs
```diff
@@ -0,0 +1,376 @@
...
+use futures::Future;
+use hyper::Response;
+use interledger_http::error::default_rejection_handler;
+
+use serde::{Deserialize, Serialize};
+
+use warp::{self, reject::Rejection, Filter};
+
+#[derive(Serialize, Deserialize, Debug, Clone, Hash)]
+pub struct CreateAccount {
```
emschwartz: Does this need to be a struct?

gakonst: I could make it a `json!({"id", id})` I believe

crates/interledger-settlement/Cargo.toml
gakonst: Yeah because we need the errors for handling stuff in Warp

Issue #474, opened 2019-10-28
SQLite store backend


bstrie:
Breaking down this task into more manageable goals:

- [ ] configuration option for selecting redis or sqlite (possibly via parsing the scheme of the given store URL?)
- [ ] crate features for enabling/disabling redis and sqlite backends (both enabled by default?)
- [ ] abstract redis-specific logic out of ilp-node
- [ ] dynamically determine whether to instantiate a prior existing node or a new node based on whether the sqlite file already exists (or configuration parameter, for use with redis as well?)
- [ ] "hello world" demo for a node that answers the `/` (status) endpoint without having the redis crate enabled at all
- [ ] one-by-one implement the store traits for SQLite (todo: list all these out and prioritize)

emschwartz:
>  abstract redis-specific logic out of ilp-node

I wouldn't work on much abstraction here. Instead, I would replace [these lines](https://github.com/interledger-rs/interledger-rs/blob/a418426634739a433ad08ccea490477f3b9f8e49/crates/ilp-node/src/node.rs#L262-L264) with a method on the node called `connect_to_store` or something like that. That method would parse the `database_url` parameter to see whether it's a `redis`, `redis+unix`, or `sqlite` URL and use the appropriate builder to create and connect to the store. That method would return an `impl Future<Item = ..., Error = ()>` where `...` is a thing that implements all the store traits.

>  dynamically determine whether to instantiate a prior existing node or a new node based on whether the sqlite file already exists (or configuration parameter, for use with redis as well?)

This should only be necessary if the SQLite store does anything to create the tables on startup. If the tables are created dynamically when they're used, its behavior would be the same whether you gave it a path to a file that didn't yet exist or one a path to a previously used database.

bstrie:
@KevinWMatthews I haven't pushed any of my WIP commits yet, but I'll try to get something up today that people can look at.

bstrie:
@KevinWMatthews I've got the initial groundwork laid in https://github.com/interledger-rs/interledger-rs/issues/474.

sappenin:
@bstrie Any updates here?

gakonst:
Given that we haven't had any progress on this, we decided with @sappenin to move forward with PostGRes and tokio-postgres for this task

sappenin:
[...]

Issue #391, opened 2019-10-07
Investigate whether some usage of generics may be unnecessary


bstrie:
In the interest of reducing compilation times and reducing the number of crates that need to be touched when making changes, it's worth investigating whether any of our generic types aren't pulling their weight by being generic.

For example, the `AccountId` associated type on the `Account` trait is a candidate for removal, since we ourselves only ever specify it as `u64` and it's unclear what alternative implementations would want to use instead and what benefits they would get from doing so.

PR #502, opened 2019-11-05
WIP: SQLite backend store

bstrie:
Getting this PR even to this rough state was harder than anticipated, since it turns out that we're sort of between a rock and a hard place as far as the type system is concerned. You definitely want to read this whole saga first before looking at the code because even I have a hard time believing the lengths to which it goes to get such a seemingly simple thing to work, but it may well still be the best option. I'm open to suggestions!

The first significant step to getting the new SQLite backend operational was to decouple ilp-node from the Redis backend. As originally envisioned, [described here](https://github.com/interledger-rs/interledger-rs/issues/474#issuecomment-547112513), this would have taken the form of a function taking in a configuration parameter (for Redis vs Sqlite) and spitting out the appropriate backend (`RedisStore` or `SqliteStore`), its concrete type hidden behind an `impl Future`.

Sadly we can't just do that. `impl Trait`, contrary to `dyn Trait`, merely hides the type from the caller rather than erasing it completely, which means that *within* the function all return paths still have to have the same type. This means `if foo { RedisStore } else { SqliteStore }` isn't going to work with `impl Trait`.

What if we try `dyn Trait` then? No, that doesn't work, for a different reason: it irrevocably turns the returned Store into a trait object, which means that the value is now subject to the usual object safety rules, which makes it impossible to call `.clone()` on our store, and sadly the node is designed such that it clones the store handle thoroughly, and attempting to change that would be a massive and risky undertaking.

Ok, so neither `-> impl Trait` nor `-> dyn Trait` will work. What if instead of trying to make this new function coalesce both stores into one type, we make the rest of the service_chain into a function that takes a generic argument, then we just branch on the config parameter and call the generic function with a different input type in each branch? At first this looks like it will work, but remember that the connection to the backend store is an asynchronous operation, so we're already handed back an `impl Future`, and even if we can manage to make the different `Store` traits all line up, what we can't do is make all the `Account` associated types on those traits *also* line up, because they're behind the opaque `impl Future` and there's no way to tell the compiler that all of them should unify into the same concrete type (a major point of `impl Trait` is to hide things like that, after all).

At this point we can take a step back and reconsider the problem: we don't actually need to pass in the `impl Future<Item = Store>` to the service chain, we can just pass in the appropriate `StoreBuilder`--which we can get our hands on *without* descending into async code--so we won't be constrained by the opaque type when writing our generic parameters. This is progress, but sadly we now have another problem, which is that our generic StoreBuilder trait needs to define a method to actually construct a store, and its return type will have to look like `impl Future<Item = Self::Store>`, where `Self::Store` is an associated type. And in the definition of StoreBuilder trait, that associated type is going to have to be defined like this:
```rust
trait StoreBuilder {
    type Store: NodeStore + HttpStore + EtcStore;
```
...which would be fine, *except* for the fact that many of the `FooStore`s are already themselves generic over an Account type, which means that the `StoreBuilder` trait actually has to be defined like:
```rust
trait StoreBuilder {
    type Store: NodeStore<A> + HttpStore<A> + EtcStore<A>;
```
...which is a classic example of Generic Associated Types, which Rust won't have support for anytime soon. So we can't do this either... *unless* we were to get rid all the associated types on the `Store` traits, which wouldn't be *completely* unthinkable, and in fact has been considered before (#391).

But let's step back further and stop trying to be fancy. We're just trying to get this done quick and dirty, so let's just extract all the service chaining into a normal non-generic function, copy/paste it once, and then call each of the two functions with concrete `RedisStore` and `SqliteStore` types so that we don't have to worry about the type system. I tried that, and it regressed compile times by over ***two minutes*** (and all of it at the very end in ilp-node, so you'd be suffering it constantly). The build time for a clean `cargo build` went from 4m30s to 6m40s. That seemed like a complete non-starter.

I also considered using feature flags in conjunction with `cfg` to just hardcode a given binary into having exactly one of the Redis-backed service chain or the Sqlite-backed service chain, but this seemed like it would be a massive pain for developing, testing, and distributing (I'm also not sure if you can even use feature flags like that in non-library crates).

In any case, after reflecting on it I figured the least-bad of all these options would be to remove the generic parameters from all the `Store` traits, so that's what this PR is, and this giant wall of text is here to hopefully convince you (and maybe also me) that this wasn't a completely unreasonable course of action. I didn't take the time to polish it because I wasn't sure if people were going to strongly object to the general approach, but it does all seem to work as far as I have tested.

This PR also includes a complete skeleton of the SqliteStore to prove that the type system isn't going to throw another fastball at us. :) Don't worry about reviewing nits just yet, this round is more for getting broad strokes of feedback than doing detail work.


gakonst:
What was the reasoning behind the tight coupling between the Account & AccountId inside interledger_service? I guess compilation times / code churn? @emschwartz just tried compiling it and he said it is not much faster

bstrie:
Closing this in favor of pursuing a macro approach while we search for other ways to reduce compilation times, since consensus is that we don't want to completely remove Account as a generic parameter (although AccountId might be a different story, but that's left for a different PR).

PR #503, opened 2019-11-06
fix(cryptocompare): ignore currencies that don't have the RAW field

dora-gt:
Currently the `exchange_rates` test fails because `MEXC` doesn't have the `RAW` field.

```
2019-11-06T02:31:55.616780071+00:00 ERROR interledger_service_util::exchange_rate_providers::cryptocompare: Error getting exchange rate response body from CryptoCompare, incorrect type: Error(Json(Error("missing field `RAW`", line: 1, column: 104464)))    
```

https://circleci.com/gh/interledger-rs/interledger-rs/3188

### DIFF
--- a/crates/interledger-service-util/src/exchange_rate_providers/cryptocompare.rs
+++ b/crates/interledger-service-util/src/exchange_rate_providers/cryptocompare.rs
@@ -79,8 +79,11 @@ pub fn query_cryptocompare(
             let rates = res
                 .data
                 .into_iter()
-                .filter(|asset| asset.raw.is_some())
-                .map(|asset| (asset.coin_info.name.to_uppercase(), asset.raw.unwrap().usd.price))
+                .filter_map(|asset| if let Some(raw) = asset.raw {
+                    Some((asset.coin_info.name.to_uppercase(), raw.usd.price))
+                  } else {
+                    None
+                  })
                 .chain(once(("USD".to_string(), 1.0)));
             Ok(HashMap::from_iter(rates))
         })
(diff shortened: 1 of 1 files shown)


crates/interledger-service-util/src/exchange_rate_providers/cryptocompare.rs
```diff
@@ -79,7 +79,8 @@ pub fn query_cryptocompare(
             let rates = res
                 .data
                 .into_iter()
-                .map(|asset| (asset.coin_info.name.to_uppercase(), asset.raw.usd.price))
+                .filter(|asset| asset.raw.is_some())
+                .map(|asset| (asset.coin_info.name.to_uppercase(), asset.raw.unwrap().usd.price))
```
emschwartz: Unwrapping will cause the node to panic. :-1: 

```suggestion
                .filter_map(|asset| if let Some(raw) = asset.raw {
                  Some((asset.coin_info.name.to_uppercase(), raw.usd.price))
                } else {
                  None
                })
```

dora-gt: Even if the assets are filtered by `is_some`?
Is that possible?

emschwartz: No, you're right. I didn't see that. Could you change it anyway? I think it's clearer to do it in one step instead of two

gakonst: Pushed a change that addresses this in https://github.com/interledger-rs/interledger-rs/pull/503/commits/b55a71935124628d098387cdf7b44da3772f57d7

PR #504, opened 2019-11-07
Remove the custom docker image

dora-gt:
This PR makes the CI process get rid of the custom Docker image.

Now it doesn't depend on any custom docker images although it does on `circleci/rust` and `clux/muslrust`. Instead of using a custom docker image, it installs dependencies inside the CI procedure every time.


dora-gt:

PR #506, opened 2019-11-08
all-in-one CI update PR

dora-gt:
Because it was difficult to grasp the entire structure with many PRs, I made another PR that includes all the PRs. We could merge either.

This PR includes:
- #496 Use binaries for examples
- **#504 Remove the custom docker image**

The features:
- Test markdown files in the CI process
- Update on examples to use binaries, ilp-cli, and cli arguments instead of env vars
- Remove CI cache (because it is faster)
- Update docker images, tag it and push
- Build `musl` and `macOS` binaries, macOS build uses `macOS executor`
- Release binaries when tags are pushed
- **Don't depend on any custom docker images**


dora-gt:
I'm trying to fix: https://github.com/interledger-rs/interledger-rs/pull/496#discussion_r343845274

scripts/run-md.sh
```diff
@@ -44,7 +44,7 @@ cat "$MD_FILE" | awk 'BEGIN { one_line_comment="^<!--!(.*)-->$" }
   # if the line matches one line comment (<!--! foo -->), just print it
   else if ($0 ~ one_line_comment) { p = $0; sub("^<!--! *", "", p); sub(" *-->$", "", p); print p }
 }' > "$TMP_SCRIPT"
-bash "$TMP_SCRIPT"
+bash -O expand_aliases "$TMP_SCRIPT"
```
dora-gt: `alias` is not applied in `non-interactive` shells.

examples/eth-xrp-three-nodes/README.md
```diff
@@ -74,100 +88,121 @@ init
...
+if [ ${SOURCE_MODE} -ne 1 ]; then
+    if [ $(is_linux) -eq 1 ]; then
 -->
+```bash
+pushd ~/.interledger/bin &>/dev/null
 
-### 1. Build interledger.rs
+# install ilp-node
+if [ ! -e "ilp-node" ]; then
+    curl -L https://github.com/interledger-rs/interledger-rs/releases/download/ilp-node-latest/ilp-node-x86_64-unknown-linux-musl.tar.gz | tar xzv
```
dora-gt: The `latest` tag is always updated.

.circleci/config.yml
```diff
@@ -56,30 +45,285 @@ jobs:
...
+      - run:
+          name: Install Dependencies
+          command: |
+            # install system dependeicies
+            sudo apt-get update
+            sudo apt-get install -y curl jq redis-server redis-tools lsof libssl-dev
+
+            # install rust components
+            cargo install cargo-audit
+            rustup component add clippy
```
emschwartz: Testing the examples definitely doesn't require these tools

.circleci/release/get_docker_image_tag.sh
```diff
@@ -0,0 +1,23 @@
...
+# If no tag is given, it is considered as a `latest` (or possibly could be said `nightly`) build.
+# This script requires `jq`
+
+crate_name=$1
+git_tag=$2
+
+if [ -n "$git_tag" ]; then
+    # If it is a tag of semantic version expression
+    if [[ "$git_tag" =~ ^.*v([0-9]+)\.([0-9]+)\.([0-9]+)(-([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?(\+[0-9A-Za-z-]+)?$ ]] ; then
+        cargo read-manifest --manifest-path crates/${crate_name}/Cargo.toml | jq -r .version
```
emschwartz: nit: if we don't need jq for a lot of other things, it might be better to avoid using a tool that needs to be explicitly installed

examples/README.md
```diff
@@ -23,14 +23,8 @@ $ ../../scripts/run-md.sh README.md
 $ (some command) | ../../scripts/run-md.sh
 ```
 
-You could also run the examples using Docker. Try specifying `USE_DOCKER=1` as follows.
+The script basically downloads compiled binaries and uses it. You could also make the script use the source code.  Try specifying `SOURCE_MODE=1` as follows.
```
emschwartz:
```suggestion
By default, the script will download compiled binaries to use with the examples. If you want it to build the project from the source code instead, you can set `SOURCE_MODE=1` as follows:
```

examples/eth-settlement/README.md
```diff
@@ -56,72 +68,111 @@ init
...
+for port in 8545 7770 8770 3000 3001; do
+    if lsof -tPi :${port} ; then
+        kill `lsof -tPi :${port}`
+    fi
+done
+-->
+
+### 1. Prepare interledger.rs
+
+First of all, we have to prepare `interledger.rs`. You could either:
```
emschwartz:
```suggestion
First of all, we have to prepare `interledger.rs`. You can either:
```

examples/eth-settlement/README.md
```diff
@@ -56,72 +68,111 @@ init
...
+
+1. Download compiled binaries
+1. Compile from the source code
+
+If you would like to play more deeply with interledger.rs, compiling from the source is considerable. It would take less time to download compiled binaries otherwise.
```
emschwartz:
```suggestion
Compiling the source code relatively slow, so we recommend downloading the pre-built binaries unless you want to modify some part of the code.
```

examples/eth-settlement/README.md
```diff
@@ -253,7 +339,7 @@ cargo run --features "ethereum" -- ethereum-ledger \
 &> logs/node-alice-settlement-engine.log &
 
 # Start Bob's settlement engine
-cargo run --features "ethereum" -- ethereum-ledger \
+interledger-settlement-engines ethereum-ledger \
```
emschwartz: This will need to be updated for https://github.com/interledger-rs/settlement-engines/pull/17

.circleci/config.yml
```diff
@@ -56,30 +45,285 @@ jobs:
...
+    resource_class: medium+
+    environment:
+      # setting BASH_ENV to custom file because .bashrc of circleci/rust doesn't work with
+      # non-interactive shell.
+      BASH_ENV: /home/circleci/.bash_circleci
+      CARGO_HOME: /home/circleci/.cargo
+    steps:
+      - checkout
+      - run:
+          name: Install Dependencies
```
gakonst: I think we can use the `default` feature of yaml to not repeat the code for dependency installation, 

https://discuss.circleci.com/t/using-defaults-syntax-in-config-yaml-aka-yaml-anchors/16168/3

examples/eth-xrp-three-nodes/README.md
```diff
@@ -74,100 +88,121 @@ init
...
 fi
+
+for port in 8545 7770 8770 9770 3000 3001 3002 3003; do
+    if lsof -tPi :${port} ; then
+        kill `lsof -tPi :${port}`
+    fi
+done
+-->
+
+### 1. Prepare interledger.rs
```
emschwartz: I know we had this conversation before about whether the examples should all be standalone files but there is a fair amount copied between them. What do you think about moving the instructions for installing everything into `examples/README.md`? You run that once and then each of the examples just tell you what you need to do to run the actual example

@dora-gt @gakonst @bstrie

scripts/release.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+# Just not to publish crates
```
emschwartz: Does this need to be a separate script (as opposed to just running this command wherever it's used)?

scripts/run-md-test.sh
```diff
@@ -1,22 +1,16 @@
 #!/bin/bash
 
-function init_test() {
+function clear_environment() {
     clear_logs
-    clear_docker
     free_ports
     clear_redis_file
+    clear_binaries
```
emschwartz: Why are the binaries removed at the end?

.circleci/config.yml
```diff
@@ -56,30 +45,285 @@ jobs:
...
+            mkdir -p /tmp/release-builds/x86_64-unknown-linux-musl
+            ./.circleci/release/build_musl_binary.sh ${crate_name} /tmp/release-builds/x86_64-unknown-linux-musl/${crate_name}
+      - persist_to_workspace:
+          root: /tmp/release-builds
           paths:
-            - /home/circleci/.cargo
+            - x86_64-unknown-linux-musl/*
+      - store_artifacts:
+          path: /tmp/release-builds/x86_64-unknown-linux-musl
+          destination: release-builds/x86_64-unknown-linux-musl
```
gakonst: Looks like we cannot get rid of musl because all docker images are built on Alpine and Alpine uses musl and does not have glibc

.circleci/config.yml
```diff
@@ -56,30 +45,285 @@ jobs:
...
+      - run:
+          name: Install Rust
+          command: |
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.38.0
+      - run:
+          # Builds release profile binaries
+          # The artifact is stored in release-builds directory, which is persistent
+          name: Build binary
+          command: |
+            crate_name=$(./.circleci/release/get_crate_name.sh ${CIRCLE_TAG})
```
gakonst: Same comment regarding yaml default tags

.circleci/config.yml
```diff
@@ -56,30 +45,285 @@ jobs:
...
+
+workflows:
+  version: 2
+  test-and-release:
+    jobs:
+      - build:
+          filters:
+            tags:
+              only: # we need this to kick builds when tags are given
+                - /.*/
```
gakonst: Will this not match when there is no tag? Why?

.circleci/release/get_docker_image_tag.sh
```diff
@@ -0,0 +1,23 @@
...
+# with the git tag.
+# If no tag is given, it is considered as a `latest` (or possibly could be said `nightly`) build.
+# This script requires `jq`
+
+crate_name=$1
+git_tag=$2
+
+if [ -n "$git_tag" ]; then
+    # If it is a tag of semantic version expression
+    if [[ "$git_tag" =~ ^.*v([0-9]+)\.([0-9]+)\.([0-9]+)(-([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?(\+[0-9A-Za-z-]+)?$ ]] ; then
```
gakonst: Why do we need to do this regex match?

examples/eth-settlement/README.md
gakonst: I will take care of updating that

.circleci/release/build.sh
```diff
@@ -0,0 +1,23 @@
...
+    exit 1
+fi
+
+if [ -z "${crate_name}" ]; then
+    printf "%s\n" "crate_name is required."
+    exit 1
+fi
+
+mkdir -p .cargo
+printf "[build]\ntarget = \"${target_name}\"\n" >> .cargo/config
```
gakonst: I think this whole build script can be removed and replaced with `cargo build --target <target_name>` commands

.circleci/config.yml
```diff
@@ -56,30 +45,285 @@ jobs:
...
+          command: |
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.38.0
+      - run:
+          # Builds release profile binaries
+          # The artifact is stored in release-builds directory, which is persistent
+          name: Build binary
+          command: |
+            crate_name=$(./.circleci/release/get_crate_name.sh ${CIRCLE_TAG})
+            printf "Building crate_name: %s, tag: %s\n" "${crate_name}" "${CIRCLE_TAG}"
+            ./.circleci/release/build.sh x86_64-apple-darwin ${crate_name}
```
gakonst: IMO we should just call `cargo build --release --target x86_x64-apple-darwin --bin $crate_name` instead of having a `build.sh` script

.circleci/release/build_musl_binary.sh
```diff
@@ -0,0 +1,35 @@
...
+fi
+
+docker run -dt --name builder "${docker_image_name}"
+
+docker cp ./Cargo.toml builder:/usr/src/Cargo.toml
+docker cp ./crates builder:/usr/src/crates
+docker cp ./.circleci/release/build.sh builder:/usr/src/build.sh
+
+# "--workdir" requires API version 1.35, but the Docker daemon API version of CircleCI is 1.32
+echo "cd /usr/src/; ./build.sh x86_64-unknown-linux-musl \"${crate_name}\"" | docker exec -i builder /bin/bash
```
gakonst: Think you can replace the pipe with:

```suggestion
docker exec -i builder "cd /usr/src && cargo build --release --target x86_x64-unknown-linux-musl $crate_name"
```

.circleci/release/github_release.sh
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env bash
+
+# Intended to be run in the top directory.
+#
+# This script creates a GitHub release and attach specified asset files to the release.
+# [WARNING] If there is any release that has the same name, it will be replaced.
+#
+# Env var:
+#   GITHUB_OAUTH_TOKEN: OAuth token for GitHub
```
gakonst: Where is this set?

.circleci/config.yml
```diff
@@ -56,30 +45,285 @@ jobs:
...
-      #     command: "bash <(curl -s https://codecov.io/bash)"
-      - save_cache:
-          key: cargo-cache-1-{{ checksum "Cargo.toml" }}-{{ .Branch }}
+  test-md:
+    docker:
+      - image: circleci/rust
+    resource_class: medium+
+    environment:
+      # setting BASH_ENV to custom file because .bashrc of circleci/rust doesn't work with
+      # non-interactive shell.
```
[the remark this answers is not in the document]
dora-gt: Because it contains:

```
# If not running interactively, don't do anything
case $- in
    *i*) ;;
      *) return;;
esac
```

I could override it but hesitated to do that because there might be some reasons.

.circleci/config.yml
```diff
@@ -56,30 +45,285 @@ jobs:
...
-      #       # --exclude-line="^(\)*\}*,*)$"
-      #     environment:
-      #       # Configure the redis tests to use a unix socket instead of TCP
-      #       REDISRS_SERVER_TYPE: unix
-      # - run:
-      #     name: Upload Code Coverage
-      #     command: "bash <(curl -s https://codecov.io/bash)"
-      - save_cache:
-          key: cargo-cache-1-{{ checksum "Cargo.toml" }}-{{ .Branch }}
+  test-md:
```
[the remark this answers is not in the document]
dora-gt: We could, but it forces us to wait for 10min or so additionally to finish the build.
It doesn't look acceptable.

.circleci/config.yml
dora-gt: Indeed, I'll remove.

.circleci/release/get_docker_image_tag.sh
dora-gt: I think parsing JSON in shell scripts is like... ☠️

.circleci/release/build.sh
dora-gt: OMG I was doing super redundantly. Thank you so much.

examples/eth-xrp-three-nodes/README.md
dora-gt: I think it should be separated because now it is repeating the same thing.

scripts/release.sh
dora-gt: Yeah, I just added this script to be more explicit.
No more meaning than that. If you don't need this, I'll remove.

scripts/run-md-test.sh
dora-gt: If I don't remove it, the markdowns don't download the binaries, which means we are not testing it (the coverage would decrease). It could be also said for `settlement-engines` but if I remove `settlement-engines` as well, we have to compile again and again and the entire testing time would increase like 5 min or so. Actually I was in two minds but I didn't add lines to remove `settlement-engines` intentionally.

.circleci/config.yml
dora-gt: Ohh interesting. It looks nice but actually the dependencies are slightly different ones, so it might be OK to leave it as-is. Thanks.

.circleci/config.yml
dora-gt: I'm sorry I don't understand. Why do you want to get rid of `musl`?

.circleci/config.yml
dora-gt: As I wrote before it is just the specifications of CircleCI.

https://github.com/interledger-rs/interledger-rs/pull/494#discussion_r342876535

examples/eth-xrp-three-nodes/README.md
dora-gt: Though I'd like to do that in another PR. I want to merge this first because it has a lot of important improvements.

scripts/run-md-test.sh
```diff
@@ -25,6 +19,13 @@ function free_ports() {
         fi
     done
 
+    # ports of redis
```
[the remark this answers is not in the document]
dora-gt: ?! I'm sorry my bad, it might be a merge mistake.

examples/simple/README.md
```diff
@@ -433,22 +354,20 @@ printf "Bob's balance: "
...
-run_hook_before_kill
+run_post_test_hook
 if [ $TEST_MODE -ne 1 ]; then
     prompt_yn "Do you want to kill the services? [Y/n] " "y"
 fi
 printf "\n"
 if [ "$PROMPT_ANSWER" = "y" ] || [ $TEST_MODE -eq 1 ] ; then
-    if [ "$USE_DOCKER" -ne 1 ]; then
-        exec 2> /dev/null
+    exec 2> /dev/null
```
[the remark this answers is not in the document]
dora-gt: Hmmmm it outputs like:

```
Do you want to kill the services? [Y/n] y
/var/folders/83/hfv5cgm17yv_kr75v339bqpc0000gn/T/tmp.HhIVplo5: line 213: 12846 Terminated: 15          ilp-node --ilp_address example.node_a --secret_seed 8852500887504328225458511465394229327394647958135038836332350604 --admin_auth_token admin-a --redis_url redis://127.0.0.1:6379/ --http_bind_address 127.0.0.1:7770 --settlement_api_bind_address 127.0.0.1:7771 >&logs/node_a.log
```

I don't think this is so nice...

examples/simple/README.md
```diff
@@ -181,26 +207,28 @@ export RUST_LOG=interledger=debug
...
-cargo run --bin ilp-node &> logs/node_a.log &
-
-ILP_ADDRESS=example.node_b \
-ILP_SECRET_SEED=1604966725982139900555208458637022875563691455429373719368053354 \
-ILP_ADMIN_AUTH_TOKEN=admin-b \
-ILP_REDIS_URL=redis://127.0.0.1:6380/ \
-ILP_HTTP_BIND_ADDRESS=127.0.0.1:8770 \
-ILP_SETTLEMENT_API_BIND_ADDRESS=127.0.0.1:8771 \
-cargo run --bin ilp-node &> logs/node_b.log &
+# Be aware that we are using `--` to differentiate arguments for `cargo` from `ilp-node`.
```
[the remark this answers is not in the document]
dora-gt: I'm sorry and thank you. It should be removed.

examples/simple/README.md
```diff
@@ -41,61 +53,110 @@ init
...
+
+#### Compile from the Source Code
+If you would prefer compiling from the source code, compile interledger.rs and CLI as follows.
+
+<!--!
+if [ ${SOURCE_MODE} -eq 1 ]; then
+    printf "Building interledger.rs... (This may take a couple of minutes)\n"
 -->
 ```bash
+# These aliases make our command invocations more natural
```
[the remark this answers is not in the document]
dora-gt: It would be better.

examples/simple/README.md
```diff
@@ -41,61 +53,110 @@ init
...
+
+- Linux based OSs
+- macOS
+
+First, let's make a directory to install binaries.
+
+```bash
+mkdir -p ~/.interledger/bin
+
+# Also write this line in .bash_profile etc if needed
```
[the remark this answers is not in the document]
dora-gt: I think it is OK because they are just using binary mode, which means they'd not want to install from source code.

Also I wanted to call just `ilp-node`. I mean, I didn't want to write `if` in markdowns.

examples/simple/README.md
```diff
@@ -41,61 +53,110 @@ init
...
 
-printf "\n"
+### 1. Prepare interledger.rs
+
+First of all, we have to prepare `interledger.rs`. You could either:
+
+1. Download compiled binaries
+1. Compile from the source code
+
+If you would like to play more deeply with interledger.rs, compiling from the source is considerable. It would take less time to download compiled binaries otherwise.
```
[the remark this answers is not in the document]
dora-gt: (I'm applying Evan's suggestion)

.circleci/config.yml
dora-gt: It had meaning before but actually it doesn't have now and you are right.
I'll fix.

.circleci/release/github_release.sh
dora-gt: image

dora-gt:
Thank you for the comments. I'll address the comments tomorrow.

@emschwartz 
The PRs are originally separated for @gakonst, so If gakonst wouldn't mind, I'll close the other PRs.
https://github.com/interledger-rs/interledger-rs/pull/305#pullrequestreview-310992490

scripts/run-md-test.sh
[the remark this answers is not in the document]
dora-gt: Ah I remembered. It was killing `nc` for Docker mode. We no longer need it anyway.

.circleci/config.yml
dora-gt: Removed.

scripts/run-md-test.sh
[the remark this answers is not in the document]
dora-gt: Fixed.

examples/simple/README.md
```diff
@@ -303,86 +260,64 @@ export ILP_CLI_API_AUTH=admin-a
...
 
 printf "Creating Node A's account on Node B...\n"
-ilp-cli --quiet --node http://localhost:8770 accounts create node_a \
+ilp-cli --node http://localhost:8770 accounts create node_a \
     --auth admin-b \
     --asset-code ABC \
     --asset-scale 9 \
-    --ilp-over-http-incoming-token node_a-password
+    --ilp-over-http-incoming-token node_a-password \
+    >logs/account-node_b-node_a.log 2>/dev/null
```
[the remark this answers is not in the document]
dora-gt: Fixed.

examples/simple/README.md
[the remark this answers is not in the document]
dora-gt: Fixed.

examples/simple/README.md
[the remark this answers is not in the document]
dora-gt: Fixed.

.circleci/release/build.sh
dora-gt: Fixed.

.circleci/config.yml
dora-gt: Fixed.

.circleci/release/build_musl_binary.sh
dora-gt: Fixed though I modified a bit because it doesn't work.

dora-gt:
I think I've addressed comments. Please let me know if there is a critical one.

.circleci/release/get_docker_image_tag.sh
emschwartz: It looks like it might already be installed by default on the CircleCI Rust docker image: https://github.com/CircleCI-Public/circleci-dockerfiles/blob/master/rust/images/1.37.0-stretch/Dockerfile#L45

.circleci/config.yml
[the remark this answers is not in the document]
emschwartz: What makes it take so long? Actually running the examples or compiling the binaries?

.circleci/config.yml
```diff
@@ -56,30 +45,281 @@ jobs:
...
           paths:
-            - /home/circleci/.cargo
+            - x86_64-unknown-linux-musl/*
+      - store_artifacts:
+          path: /tmp/release-builds/x86_64-unknown-linux-musl
+          destination: release-builds/x86_64-unknown-linux-musl
+  build-release-binary-darwin:
+    macos:
+      xcode: "10.2.0"
+    resource_class: medium+
```
emschwartz: Does this need to be run on a large @gakonst ?

.circleci/release/get_docker_image_tag.sh
dora-gt: Oh sorry, I'll remove.

.circleci/config.yml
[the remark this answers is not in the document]
dora-gt: Hmmm, maybe both.

1. Compilation time
    - We separated `interledger-settlement-engines` so it added a few minutes to compile.
1. Testing time
    - Also, now it is testing both binary mode and source code mode, so it takes about 8 min.

scripts/run-md.sh
bstrie: > `alias` is not applied in `non-interactive` shells.

Allegedly it is possible to use`shopt -s expand_aliases` to allow alises to work in non-interactive shells, but I was never able to get this work for the examples (this is why the READMEs define `function ilp-cli` rather than `alias ilp-cli`). However, even if `shopt -s expand_aliases` could be made to work, that would only work in bash and not other shells like zsh.

examples/eth-xrp-three-nodes/README.md
bstrie: Agreed that it would be nice to share more things between the examples, but also agreed that that should be left for a follow-up PR.

.circleci/config.yml
```diff
@@ -56,30 +45,277 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Install Rust
+          command: |
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.38.0
```
bstrie: We need to begin running on 1.39, which was released last week.

examples/simple/README.md
```diff
@@ -41,61 +53,113 @@ init
 
 printf "Stopping Interledger nodes...\n"
 
-if [ "$USE_DOCKER" -eq 1 ]; then
```
bstrie: Getting rid of these ifs is so good 🙌

examples/simple/README.md
```diff
@@ -41,61 +53,113 @@ init
...
+<!--!
+if [ ${SOURCE_MODE} -eq 1 ]; then
+    printf "Building interledger.rs... (This may take a couple of minutes)\n"
 -->
 ```bash
+# These aliases make our command invocations more natural
+# Be aware that we are using `--` to differentiate arguments for `cargo` from `ilp-node` or `ilp-cli`.
+# Arguments before `--` are used for `cargo`, after are used for `ilp-node`.
+
+alias ilp-node="cargo run --quiet --bin ilp-node --"
```
bstrie: How did you get these aliases to work? I always ran into strange and undebuggable error messages when trying to use aliases here, which is why I secretly used a function instead.

.circleci/config.yml
dora-gt: I'll fix to use `stable` instead for now.

examples/simple/README.md
dora-gt: I just added `-O expand_aliases` to `bash` option.

https://github.com/interledger-rs/interledger-rs/pull/506#discussion_r344256127

.circleci/config.yml
dora-gt:
```
info: syncing channel updates for 'stable-x86_64-apple-darwin'
info: latest update on 2019-11-07, rust version 1.39.0 (4560ea788 2019-11-04)
```
Confirmed.

examples/simple/README.md
dora-gt: Btw `run-md.sh` uses `bash` explicitly.

dora-gt:
Maybe time to merge?

bstrie:
@dora-gt Anything preventing this being merged?

dora-gt:
Oh sorry, no, I'll merge.

dora-gt:
image

dora-gt:
Also merged: https://github.com/interledger-rs/settlement-engines/pull/14

PR #509, opened 2019-11-09
Optional feature to publish packets to Google PubSub

emschwartz:
This is an experimental feature for publishing fulfilled packets to a message queue (in this case, Google's hosted PubSub service). This can be used as an alternative to real-time balance tracking because the amounts of the fulfilled packets can be summed up later to determine the balances with specific peers.

It's possible this will be removed later or we may abstract the behavior more so it's not so tied to Google Cloud.

### DIFF
--- a/crates/ilp-node/Cargo.toml
+++ b/crates/ilp-node/Cargo.toml
@@ -11,6 +11,9 @@ default-run = "ilp-node"
 [features]
 default = ["balance-tracking"]
 balance-tracking = []
+# This is an experimental feature that enables submitting packet
+# records to Google Cloud PubSub. This may be removed in the future.
+google-pubsub = ["base64", "chrono", "parking_lot", "reqwest", "serde_json", "yup-oauth2"]
 
 [dependencies]
 bytes = { version = "0.4.12", default-features = false }
@@ -34,6 +37,14 @@ libc = { version = "0.2.62", default-features = false }
 warp = { version = "0.1.20", default-features = false, features = ["websocket"] }
 secrecy = { version = "0.5.0", default-features = false, features = ["alloc", "serde"] }
 uuid = { version = "0.8.1", default-features = false, features = ["v4"] }
+
+# For google-pubsub
+base64 = { version = "0.10.1", default-features = false, optional = true }
+chrono = { version = "0.4.9", default-features = false, features = [], optional = true}
+parking_lot = { version = "0.9.0", default-features = false, optional = true }
+reqwest = { version = "0.9.22", default-features = false, features = ["default-tls"], optional = true }
+serde_json = { version = "1.0.41", default-features = false, optional = true }
+yup-oauth2 = { version = "3.1.1", default-features = false, optional = true }
 
 [dev-dependencies]
 approx = { version = "0.3.2", default-features = false }
--- a/crates/ilp-node/src/google_pubsub.rs
+++ b/crates/ilp-node/src/google_pubsub.rs
@@ -1,0 +1,186 @@
+use base64;
+use chrono::Utc;
+use futures::{
+    future::{ok, Either},
+    Future,
+};
+use interledger::{
+    packet::Address,
+    service::{Account, BoxedIlpFuture, OutgoingRequest, OutgoingService, Username},
+};
+use parking_lot::Mutex;
+use reqwest::r#async::Client;
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, sync::Arc};
+use tokio::spawn;
+use tracing::{error, info};
+use yup_oauth2::{service_account_key_from_file, GetToken, ServiceAccountAccess};
+
+static TOKEN_SCOPES: Option<&str> = Some("https://www.googleapis.com/auth/pubsub");
+
+/// Configuration for the Google PubSub packet publisher
+#[derive(Deserialize, Clone, Debug)]
+pub struct PubsubConfig {
+    /// Path to the Service Account Key JSON file.
+    /// You can obtain this file by logging into [console.cloud.google.com](https://console.cloud.google.com/)
+    service_account_credentials: String,
+    project_id: String,
+    topic: String,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct PubsubMessage {
+    message_id: Option<String>,
+    data: Option<String>,
+    attributes: Option<HashMap<String, String>>,
+    publish_time: Option<String>,
+}
+
+#[derive(Serialize)]
+struct PubsubRequest {
+    messages: Vec<PubsubMessage>,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct PacketRecord {
+    prev_hop_account: Username,
+    prev_hop_asset_code: String,
+    prev_hop_asset_scale: u8,
+    prev_hop_amount: u64,
+    next_hop_account: Username,
+    next_hop_asset_code: String,
+    next_hop_asset_scale: u8,
+    next_hop_amount: u64,
+    destination_ilp_address: Address,
+    fulfillment: String,
+    timestamp: String,
+}
+
+/// Create an Interledger service wrapper that publishes records
+/// of fulfilled packets to Google Cloud PubSub.
+///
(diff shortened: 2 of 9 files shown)


crates/ilp-node/Cargo.toml
```diff
@@ -31,6 +36,14 @@ warp = { version = "0.1.20", default-features = false, features = ["websocket"]
 secrecy = { version = "0.5.0", default-features = false, features = ["alloc", "serde"] }
 uuid = { version = "0.8.1", default-features = false, features = ["v4"] }
 
+# For google-pubsub
+base64 = { version = "0.10.1", default-features = false, optional = true }
+chrono = { version = "0.4.9", default-features = false, features = [], optional = true}
```
gakonst: if `default-features = false`, think you don't need to do `features = []`

crates/ilp-node/src/node.rs
```diff
@@ -211,6 +213,8 @@ pub struct InterledgerNode {
     /// If this configuration is not provided, the node will not collect metrics.
     #[serde(default)]
     pub prometheus: Option<PrometheusConfig>,
+    #[cfg(feature = "google-pubsub")]
```
gakonst: Doesn't this also need `#[serde(default)]`?

crates/ilp-node/src/google_pubsub.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// Path to the Service Account Key JSON file.
+    /// You can obtain this file by logging into [console.cloud.google.com](https://console.cloud.google.com/)
+    service_account_credentials: String,
+    project_id: String,
+    topic: String,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct PubsubMessage {
```
gakonst:
```suggestion
/// A PubsubMessage as defined in [Google's API documentation] (https://cloud.google.com/pubsub/docs/overview)
struct PubsubMessage {
```

crates/ilp-node/src/google_pubsub.rs
```diff
@@ -0,0 +1,190 @@
...
+                                };
+                                let data = base64::encode(&serde_json::to_string(&record).unwrap());
+
+                                client
+                                    .post(api_endpoint.as_str())
+                                    .bearer_auth(token.access_token)
+                                    .json(&PubsubRequest {
+                                        messages: vec![PubsubMessage {
+                                            // TODO should there be an ID?
+                                            message_id: None,
```
gakonst: This looks always empty and is not used by the Google API, can we remove it?

crates/ilp-node/src/google_pubsub.rs
```diff
@@ -0,0 +1,190 @@
...
+                                            data: Some(data),
+                                            attributes: None,
+                                            publish_time: None,
```
gakonst: This is not included anywhere in the API & is always None, what is the reason to have it as a field? I'd expect that the backend generates a timestamp when it receives the request

crates/ilp-node/src/google_pubsub.rs
```diff
@@ -0,0 +1,190 @@
...
+                                let data = base64::encode(&serde_json::to_string(&record).unwrap());
+
+                                client
+                                    .post(api_endpoint.as_str())
+                                    .bearer_auth(token.access_token)
+                                    .json(&PubsubRequest {
+                                        messages: vec![PubsubMessage {
+                                            // TODO should there be an ID?
+                                            message_id: None,
+                                            data: Some(data),
```
gakonst: Can we remove the Option here if it's always populated?

crates/ilp-node/src/google_pubsub.rs
```diff
@@ -0,0 +1,190 @@
...
+                                            attributes: None,
+                                            publish_time: None,
+                                        }],
+                                    })
+                                    .send()
+                                    .map_err(|err| {
+                                        error!(
```
gakonst: Should we be scoping the error logs in some way, similar to how you do `%status` in line 175?

crates/ilp-node/src/google_pubsub.rs
```diff
@@ -0,0 +1,190 @@
...
+
+                                client
+                                    .post(api_endpoint.as_str())
+                                    .bearer_auth(token.access_token)
+                                    .json(&PubsubRequest {
+                                        messages: vec![PubsubMessage {
+                                            // TODO should there be an ID?
+                                            message_id: None,
+                                            data: Some(data),
+                                            attributes: None,
```
gakonst: Is Google's API able to handle a JSON which does not contain the `attributes` field, or do we have to provide it as an empty dictionary? If they can handle it, we can remove this attribute

crates/ilp-node/src/google_pubsub.rs
```diff
@@ -0,0 +1,190 @@
...
+        info!("Fulfilled packets will be submitted to Google Cloud Pubsub (project ID: {}, topic: {})", config.project_id, config.topic);
+        Some((client, api_endpoint, token_fetcher))
+    } else {
+        None
+    };
+
+    move |request: OutgoingRequest<A>, mut next: O| -> BoxedIlpFuture {
+        match &utilities {
+            // Just pass the request on if no Google Pubsub details were configured
+            None => Box::new(next.send_request(request)),
```
gakonst: What if we just removed this if/else and made the function accept only a `PubsubConfig` (instead of an Option)? Removes the need for the `map`, and allows us to have one less service in the chain at runtime if the binary is compiled with the feature flag, but no pubsub config is provided

crates/ilp-node/src/google_pubsub.rs
emschwartz: I included it because it's part of the struct definition but you're right that we aren't using it. I don't feel that strongly about it either way. We could also remove the other fields we aren't using

crates/ilp-node/src/google_pubsub.rs
emschwartz: This doesn't work because you can't have the if and else branches return different types. We might be able to make that change when we enable boxing the services so that both branches can return a boxed service.

gakonst:
CI error is related to  tracing_core::field::Value` is not satisfied

gakonst:
Bumped up the CI boxes as discussed so that we do not waste timetoo much time on this

crates/ilp-node/Cargo.toml
```diff
@@ -31,6 +36,14 @@ warp = { version = "0.1.20", default-features = false, features = ["websocket"]
 secrecy = { version = "0.5.0", default-features = false, features = ["alloc", "serde"] }
 uuid = { version = "0.8.1", default-features = false, features = ["v4"] }
 
+# For google-pubsub
+base64 = { version = "0.10.1", default-features = false, optional = true }
```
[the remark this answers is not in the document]
bstrie: I'm willing to rethink the approach here, what we've got now is just an extreme case of trying to keep compilation times down at all costs. I'm not sure how we'd implement that suggestion in particular though.

crates/ilp-node/src/google_pubsub.rs
```diff
@@ -0,0 +1,190 @@
...
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, sync::Arc};
+use tokio::spawn;
+use tracing::{error, info};
+use yup_oauth2::{service_account_key_from_file, GetToken, ServiceAccountAccess};
+
+static TOKEN_SCOPE: &str = "https://www.googleapis.com/auth/pubsub";
+lazy_static! {
+    static ref SCOPES: Vec<String> = vec![TOKEN_SCOPE.to_string()];
+}
```
bstrie: You should be able to replace these lines with:
```rust
static SCOPES: Option<&str> = Some("https://www.googleapis.com/auth/pubsub");
```
Since the only places that SCOPES gets used is a function whose input is "something that can be turned into an iterator" yielding "something that can be turned into a string", and `Option<&str>` satisfies that.

crates/ilp-node/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+#![type_length_limit = "1119051"]
```
bstrie: 😨

Issue #472, opened 2019-10-28
Document node configuration parameters


emschwartz:

dora-gt:
I'm doing this for now but I think also we should consider a more structured explanation like #240

PR #512, opened 2019-11-11
feat(node cli explanation, docs): add node config explanation, fix node cli explanation

dora-gt:
Fix: #472

Add node config explanation, and also fix the wrong explanation of CLI.
Maybe I should add the same thing for `interledger-settlement-engines`.

(I didn't add the pubsub config because it seemed an experimental feature)

### DIFF
--- a/crates/ilp-node/src/main.rs
+++ b/crates/ilp-node/src/main.rs
@@ -106,8 +106,24 @@ pub fn main() {
                 This amount is kept as the node operator's profit, or may cover \
                 fluctuations in exchange rates.
                 For example, take an incoming packet with an amount of 100. If the \
-                exchange rate is 1:2 and the spread is 0.01, the amount on the \
-                    outgoing packet would be 198 (instead of 200 without the spread).")
+                exchange rate is 1:0.5 and the spread is 0.01, the amount on the \
+                    outgoing packet would be 198 (instead of 200 without the spread)."),
+        Arg::with_name("prometheus.bind_address")
+            .long("prometheus.bind_address")
+            .takes_value(true)
+            .help("IP address and port to host the Prometheus endpoint on."),
+        Arg::with_name("prometheus.histogram_window")
+            .long("prometheus.histogram_window")
+            .takes_value(true)
+            .help("Amount of time, in milliseconds, that the node will collect data \
+                points for the Prometheus histograms. Defaults to 300000ms (5 minutes)."),
+        Arg::with_name("prometheus.histogram_granularity")
+            .long("prometheus.histogram_granularity")
+            .takes_value(true)
+            .help("Granularity, in milliseconds, that the node will use to roll off \
+                old data. For example, a value of 1000ms (1 second) would mean that the \
+                node forgets the oldest 1 second of histogram data points every second. \
+                Defaults to 10000ms (10 seconds)."),
         ]);
 
     let mut config = get_env_config("ilp");
--- a/docs/api.md
+++ b/docs/api.md
@@ -250,7 +250,13 @@ Health check.
 
 Admin only.
 
-Sets the exchange rates for the node.
+Sets the exchange rates for the node. Note that the rates should be expressed by strength. For example, the present rates might look like:
+
+- USD: 1
+- EUR: 1.105
+- BTC: 8564.7348
+
+If we are relaying currency `A` of an account of `Alice` to currency `B` of an account of `Bob` and the rates of `A` and `B` are `1` and `2` respectively, then the currency `B` is stronger than `A`. So `Bob` will receive the amount multiplied by `0.5` (`1 / 2`). For instance, if `Alice` sends `100` to `Bob`, `Bob` will receive `50`, not `200`.
 
 #### Request
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,8 +1,8 @@
-### Configuration
+## How to Specify Configurations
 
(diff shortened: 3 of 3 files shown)


crates/ilp-node/src/main.rs
```diff
@@ -106,7 +106,7 @@ pub fn main() {
                 This amount is kept as the node operator's profit, or may cover \
                 fluctuations in exchange rates.
                 For example, take an incoming packet with an amount of 100. If the \
-                exchange rate is 1:2 and the spread is 0.01, the amount on the \
+                exchange rate is 1:0.5 and the spread is 0.01, the amount on the \
```
dora-gt: @gakonst should check if possible.

docs/configuration.md
```diff
@@ -1,45 +1,124 @@
...
+    - `CoinCap`
+    - Exchange rate API to poll for exchange rates. If this is not set, the node will not poll for rates and will instead use the rates set via the HTTP API. Note that [CryptoCompare](#using-cryptocompare) can also be used **when the node is configured via a config file or stdin**, because an API key must be provided to use that service.
+- exchange_rate_poll_interval
+    - Non-negative Integer (in milliseconds)
+    - `60000`
+    - Interval, defined in milliseconds, on which the node will poll the `exchange_rate_provider` (if specified) for exchange rates.
+- exchange_rate_spread
+    - Float
+    - `0.01`
+    - Spread, as a fraction, to add on top of the exchange rate. This amount is kept as the node operator's profit, or may cover fluctuations in exchange rates. For example, take an incoming packet with an amount of 100. If the exchange rate is 1:0.5 and the spread is 0.01, the amount on the outgoing packet would be 198 (instead of 200 without the spread).
```
dora-gt: Also here @gakonst .

dora-gt:
(sry I forgot to add some params, I'm adding)

crates/ilp-node/src/main.rs
```diff
@@ -106,8 +106,24 @@ pub fn main() {
                 This amount is kept as the node operator's profit, or may cover \
                 fluctuations in exchange rates.
                 For example, take an incoming packet with an amount of 100. If the \
-                exchange rate is 1:2 and the spread is 0.01, the amount on the \
-                    outgoing packet would be 198 (instead of 200 without the spread).")
+                exchange rate is 1:0.5 and the spread is 0.01, the amount on the \
+                    outgoing packet would be 198 (instead of 200 without the spread)."),
+        Arg::with_name("prometheus.bind_address")
```
dora-gt: It seemed that we could not specify prometheus configs from CLI, I added these.

docs/configuration.md
gakonst: Yeah as I wrote in the other comment, 1:2 rate means for every 1 unit of the incoming asset we send out 2. So let's make it `1:2`. 

(The bug was really at the implementation level, we were fetching `USD/XRP` and `USD/ETH`)

```suggestion
    - Spread, as a fraction, to add on top of the exchange rate. This amount is kept as the node operator's profit, or may cover fluctuations in exchange rates. For example, take an incoming packet with an amount of 100. If the exchange rate is 1:2 and the spread is 0.01, the amount on the outgoing packet would be 198 (instead of 200 without the spread).
```

crates/ilp-node/src/main.rs
```diff
@@ -106,8 +106,24 @@ pub fn main() {
                 This amount is kept as the node operator's profit, or may cover \
                 fluctuations in exchange rates.
                 For example, take an incoming packet with an amount of 100. If the \
-                exchange rate is 1:2 and the spread is 0.01, the amount on the \
-                    outgoing packet would be 198 (instead of 200 without the spread).")
+                exchange rate is 1:0.5 and the spread is 0.01, the amount on the \
```
gakonst:
```suggestion
                exchange rate is 1:2 and the spread is 0.01, the amount on the \
```

crates/ilp-node/src/main.rs
dora-gt: I'm sorry but I don't understand. If the rates are:

- `1:2` (`incoming currency rate against USD`:`outgoing currency rate against USD`)

the outgoing amount will be `49.5` because the rate is now calculated by `rates[0] / rates[1]`.

dora-gt: Ah, you might be referring the **rate** as just an **amount rate** but I think we should not use that because we cannot specify that value for our API. We have to specify `1` and `0.5` instead of `1` and `2` for our API. If we use `1` and `2` here, it might confuse people.

I think we should add an explanation for that.

dora-gt:
Added an explanation of rates.

https://github.com/interledger-rs/interledger-rs/pull/512/files#diff-0f426376b964e609c5c4618eb8bed3b2R253-R259

docs/api.md
```diff
@@ -250,7 +250,13 @@ Health check.
...
 Admin only.
 
-Sets the exchange rates for the node.
+Sets the exchange rates for the node. Note that the rates should be expressed by strength. For example, the present rates might look like:
+
+- USD: 1
+- EUR: 1.105
+- BTC: 8564.7348
+
+If we are relaying currency `A` of an account of `Alice` to currency `B` of an account of `Bob` and the rates of `A` and `B` are `1` and `2` respectively, then the currency `B` is stronger than `A`. So `Bob` will receive the amount multiplied by `0.5` (`1 / 2`). For instance, if `Alice` sends `100` to `Bob`, `Bob` will receive `50`, not `200`.
```
gakonst: Correct

PR #514, opened 2019-11-17
Allow Deposits and Withdrawals from the Settlement System

gakonst:
Based on the ideas from: https://forum.interledger.org/t/standardized-deposit-and-withdrawal/824

This is a rough sketch of how the system could look like, I'm opening this PR to get some feedback and iterate on the API.

# `GET /accounts/:username/deposits`
1. User queries that endpoint
2. Connector queries settlement engine
3. Engine responds to connector with payment info
4. Connector responds to user with payment info

This could either be a LN invoice, or just an ETH address. User then sends a payment from their wallet to that address. The engine is watching for incoming transactions, and once the tx is verified, it will get the account id associated with the sending address. 

Note: This requires overloading the CreateAccount API with additional information beyond the account id. When adding a user who is not running a connector or an engine, the node operator should provide extra data for that user, which would be interpreted by the engine in certain ways. That could be the user's address, or channel information
 
# `POST /accounts/:username/withdrawals` 
1. User queries that endpoint with some amount
2. Connector checks if the user's balance will not become negative and below `min_balance`. If so, it subtracts that amount from their `balance`/`prepaid_amount` pair (similarly to how it's done with `prepare_balances_for_prepare`)
3. Connector sends a withdrawal request to the engine's `/settlements` endpoint (similar to how it's done via `settlement_client.send_settlement` in the `balance_service`)

### DIFF
--- a/crates/interledger-api/src/http_retry.rs
+++ b/crates/interledger-api/src/http_retry.rs
@@ -2,9 +2,9 @@
 use futures::future::Future;
 use futures_retry::{ErrorHandler, FutureRetry, RetryPolicy};
 use http::StatusCode;
+use interledger_settlement::core::types::CreateAccount;
 use log::trace;
 use reqwest::r#async::Client as HttpClient;
-use serde_json::json;
 use std::{default::Default, fmt::Display, time::Duration};
 use url::Url;
 
@@ -39,10 +39,14 @@ impl Client {
             .path_segments_mut()
             .expect("Invalid settlement engine URL")
             .push(ACCOUNTS_ENDPOINT);
+        let req = CreateAccount {
+            id: id.to_string(),
+            extra,
+        };
         trace!(
-            "Sending account {} creation request to settlement engine: {:?}",
-            id,
-            se_url.clone()
+            "Sending account creation request to settlement engine at URL {:?}: {:?}",
+            se_url.clone(),
+            req,
         );
 
         // The actual HTTP request which gets made to the engine
@@ -50,7 +54,7 @@ impl Client {
         let create_settlement_engine_account = move || {
             client
                 .post(se_url.as_ref())
-                .json(&json!({ "id": id.to_string(), "extra": extra }))
+                .json(&req)
                 .send()
                 .and_then(move |response| {
                     // If the account is not found on the peer's connector, the
--- a/crates/interledger-api/src/http_retry.rs
+++ b/crates/interledger-api/src/http_retry.rs
@@ -32,7 +32,7 @@ impl Client {
         &self,
         engine_url: Url,
         id: T,
-        extra: Option<Vec<u8>>,
+        extra: Option<String>,
     ) -> impl Future<Item = StatusCode, Error = reqwest::Error> {
         let mut se_url = engine_url.clone();
         se_url
--- a/crates/interledger-api/src/http_retry.rs
+++ b/crates/interledger-api/src/http_retry.rs
@@ -32,6 +32,7 @@ impl Client {
         &self,
         engine_url: Url,
         id: T,
+        extra: Option<Vec<u8>>,
     ) -> impl Future<Item = StatusCode, Error = reqwest::Error> {
         let mut se_url = engine_url.clone();
         se_url
@@ -49,7 +50,7 @@ impl Client {
         let create_settlement_engine_account = move || {
             client
                 .post(se_url.as_ref())
-                .json(&json!({"id" : id.to_string()}))
+                .json(&json!({ "id": id.to_string(), "extra": extra }))
                 .send()
                 .and_then(move |response| {
                     // If the account is not found on the peer's connector, the
--- a/crates/interledger-api/src/routes/wallet.rs
+++ b/crates/interledger-api/src/routes/wallet.rs
@@ -1,0 +1,548 @@
+use futures::{
+    future::{err, ok, Either},
+    Future,
+};
+use hyper::{Response, StatusCode};
+use interledger_http::{error::*, HttpAccount, HttpStore};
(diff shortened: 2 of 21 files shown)


crates/interledger-api/src/routes/wallet.rs
```diff
@@ -0,0 +1,548 @@
...
+            if account.settlement_engine_details().is_some() {
+                Ok(account.clone())
+            } else {
+                let error_msg = format!("Account {} does not have settlement engine details configured. Will not attempt to withdraw", account.id());
+                error!("{}", error_msg);
+                Err(ApiError::from_api_error_type(&NO_ENGINE_CONFIGURED_ERROR_TYPE).detail(error_msg))
+            }
+        })
+        .and_then(move |account| {
+            // TODO: is there any race condition where this future completes and executes the request on the engine simultaneously with another call?
```
bstrie: Is it a problem if two tasks are attempting to withdraw simultaneously? For the Redis store implementation, it looks like the withdrawal Lua script makes sure to check that the balance won't become negative, and Lua scripts are atomic within Redis so even if multiple tasks are attempting to withdraw there shouldn't be a race condition from that.

gakonst: Yup I believe I added all the necessary checks, have added this comment to make sure I (& reviewers) revisit this since it's a critical thing that withdrawals happen only once. I think it should be safe.

crates/interledger-api/src/http_retry.rs
```diff
@@ -32,24 +32,29 @@ impl Client {
...
         engine_url: Url,
         id: T,
+        extra: Option<String>,
     ) -> impl Future<Item = StatusCode, Error = reqwest::Error> {
         let mut se_url = engine_url.clone();
         se_url
             .path_segments_mut()
             .expect("Invalid settlement engine URL")
             .push(ACCOUNTS_ENDPOINT);
+        let req = CreateAccount {
```
carlhua: So this abstracts away the settlement engine level handling - effectively giving control to the settlement engine implementer to handle how it should create, store and manage ledger specific deposit information at the account level. This way the settlement engine layer can also monitor the tx based on the the specific tx monitoring implementation of the settlement engine. Am I correct? This is my first time looking over Rust interledger so forgive my ignorance here.

crates/interledger-api/src/routes/wallet.rs
```diff
@@ -0,0 +1,553 @@
...
+                let error_type = ApiErrorType {
+                    r#type: &ProblemType::Default,
+                    status: StatusCode::INTERNAL_SERVER_ERROR,
+                    title: "Withdraw funds error", 
+                };
+                ApiError::from_api_error_type(&error_type).detail(error_msg)
+            })
+            .and_then(move |_| {
+                let client = SettlementClient::new();
+                client.send_settlement(account.clone(), amount)
```
carlhua: Overall I think its extremely critical to ensure these operations complete/cancel in an atomic fashion. I am unsure if this implementation here realizes such requirements. But given this is an initial sketch I think the idea here is clear at the functional level. In general I think anywhere we touch real ledger balance need to have a certain level of rigor. no action here just some thoughts.

crates/interledger-api/src/http_retry.rs
```diff
@@ -32,24 +32,29 @@ impl Client {
...
+            id: id.to_string(),
+            extra,
+        };
         trace!(
-            "Sending account {} creation request to settlement engine: {:?}",
-            id,
-            se_url.clone()
+            "Sending account creation request to settlement engine at URL {:?}: {:?}",
```
carlhua: I also think we should consider expanding features such as limiting the withdraw frequency or give options to set min balance - or even enable automatic withdraw. However, I can also see that these features potentially do not belong inside interledger and should be implemented at the connector application level (user facing). just some thoughts, no action.

crates/interledger-api/src/routes/wallet.rs
```diff
@@ -0,0 +1,553 @@
...
+                Ok(account.clone())
+            } else {
+                let error_msg = format!("Account {} does not have settlement engine details configured. Will not attempt to withdraw", account.id());
+                error!("{}", error_msg);
+                Err(ApiError::from_api_error_type(&NO_ENGINE_CONFIGURED_ERROR_TYPE).detail(error_msg))
+            }
+        })
+        .and_then(move |account| {
+            // TODO: is there any race condition where this future completes and executes the request on the engine simultaneously with another call?
+            store_clone.withdraw_funds(account_id, amount)
```
carlhua: I am assuming the account ID for deposit is provided by settlement engine so it can monitor. However, where can the user specify the withdraw account address?

gakonst: During account creation, the address can be provided in the `settlement_extra` field, as done here: https://github.com/interledger-rs/settlement-engines/pull/20/files#diff-39f21ff9417cf765c0e1435811dcedebR155

crates/interledger-api/src/routes/wallet.rs
gakonst: Yeah good catch, should add an `or_else` similar to how it's done in the balance_service if the settlement fails: https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-service-util/src/balance_service.rs#L168-L171

crates/interledger-api/src/http_retry.rs
gakonst: This is correct

crates/interledger-api/src/http_retry.rs
gakonst: Definitely agree these are useful features, unclear where they should be added though as you said

gakonst:
Closing until we get clarity on how we will approach this feature.

Issue #498, opened 2019-11-05
HTTP API's Content-Type header validation is too strict


latenssi:
Atleast _axios_ (an npm module for http requests) uses `Content-Type:  application/json;charset=utf-8` by default and this is rejected by current implementation of `deserialize_json` in [interledger-http](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-http/src/lib.rs#L47) because of the extra `;charset=utf-8`.

We are using a nodejs service to update rates in our ilp-nodes and this strict content-type requirement forces us to specify the content-type explicitly:
`Content-Type:  application/json;charset=utf-8` => `Content-Type:  application/json`

It is nothing major but an unnecessary inconvenience in my opinion.

dora-gt:
Sorry for the inconvenience. This should be fixed by #517.

PR #517, opened 2019-11-18
feat(http api): accepts content-type with arguments

dora-gt:
Fix: #498 

Now the `deserialize_json` filter accepts `content-type` header with `charset` arguments.


dora-gt:
Indeed, I added tests.

Issue #500, opened 2019-11-05
Update client.rs to use Outgoing Bearer token instead of AuthToken for outgoing IlpOverHttp requests.


sappenin:
This relates to #357, but currently, if an IlpOverHttp account is configured with an `outgoing` bearer token, the Rust implementation is inferring token-format unnecessarily.

Per IlpOverHttp Auth Recs [here](https://github.com/interledger/rfcs/pull/559/files#diff-4b23f32e485ebb46ea9cf36892850ffbR44), the `token-user` (which is the Rust Connector when it's making outgoing IlpOverHttp requests to a peer) should treat auth tokens as opaque and not attempt to do any parsing of them. This will allow non-Rust connectors to construct token-schemes without needing to coordinate the scheme with any other ILP implementations (such as the Rust implementation, in this case).

One manifestation of this in today's code, for example, is if I configure an account on the Rust connector with an outgoing bearer token of `foo:bar:baz`. While this value will be stored into the database, the Rust implementation will attempt to parse this token into an `AuthToken`, which will assume the String adheres to that formatting -- the implementation will only send `foo:bar` to the peer, which will trigger an HTTP 401 error.

More formally, this issue requests to change `client.rs` to not use the current `AuthToken` when assembling its `Authorization` header, but instead simply use `token` as found [here](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-http/src/client.rs#L70).

dora-gt:
Regarding the RFC, shouldn't we get rid of prefixing with usernames? It seems that `Bearer` tokens (`b64token`) don't allow `:` in its credentials. So our bearer tokens don't conform to normative `Bearer` tokens, which doesn't look nice. If we want to do so, we could issue paths per user like `/accounts/:username/ilp`. Then we could treat our tokens as totally opaque.

```
     b64token    = 1*( ALPHA / DIGIT /
                       "-" / "." / "_" / "~" / "+" / "/" ) *"="
     credentials = "Bearer" 1*SP b64token
```

Also, I'm curious about how we differentiate the type of authentication that the bearer sent? It seems that there are no type definitions in the `Authorization ` header.

Anyway, I'll look into.

Issue #357, opened 2019-09-30
Incoming and outgoing tokens should match


emschwartz:
Right now the `ilp_over_btp_outgoing_token` has to be in the form `username:password` while the `ilp_over_btp_incoming_token` has to be in the form `password`. We should change it so both values can be exactly the same because it's way too confusing otherwise.

We should also make sure the same applies to `ilp_over_http`

dora-gt:
I thought about this and I'm wondering which is the best approach:

1. Outgoing requests use "our" account username as the outgoing token username.
1. Split the outgoing token into `outgoing_token` and `outgoing_username`.
1. The incoming token includes an incoming username.

My opinion is "2" maybe.

1 is not good because the usernames might differ on our node and their node. (Or... should it be always a **relation name** like "node_a_b" or something? more confusing?)
2 might look a bit different from incoming tokens but it is inevitable because it represents another account on another node.
3 looks a bit redundant and confusing because it has another `username` though the account is clear already.

What do you think? @interledger-rs/contributors

gakonst:
I am in favor of 2 with a slight twist:

1. Make the `outgoing_token` always be just the `token`, wihtout any password.
2. Make the `outgoing_username` optional. 
  - If it does not exist, it should use our `username`. This would require adding an extra parameter somewhere, probably on `InterledgerNode` config, which would be our Node's `username`.
  - If it exists, it should use that instead as the username.

dora-gt:
Thank you @gakonst, what do you mean by "wihtout any password"?

emschwartz:
Right now the whole thing we're doing with including the username in the token is pretty janky. I think we might want to fix that, along with making the configuration options clearer.

> Make the outgoing_token always be just the token, wihtout any password.

I think @gakonst meant username rather than password

> Make the outgoing_username optional

If this were provided, I would expect it to use basic auth instead of bearer token auth

dora-gt:
@emschwartz Understood. One more question, do you mean, when the incoming service (HTTP/BTP) gets requests, it should be identified by ONLY the "token"?

I mean, the outgoing tokens should be used for "authentication" (identifying whose request it is) without usernames? Then, for example, `get_account_from_http_auth(username, password)` is going to be like `get_account_from_http_auth(access_token)`.

[...]

PR #518, opened 2019-11-19
fix(outgoing token): fix not to discard remaining commma separated values

dora-gt:
Fix: #500

⚠️ Note that procedures of authorization should be entirely refactored once the RFC of `Interledger HTTP Authentication Profiles` by @sappenin is merged, considering also #357.

Formerly, some parts of outgoing tokens are discarded unexpectedly. For example, if we parse an
outgoing token of `interledger:rust:test`, it had been sent as `interledger:rust`. This fixes to
send it as `interledger:rust:test`.

### DIFF
--- a/crates/interledger-service/src/auth/token.rs
+++ b/crates/interledger-service/src/auth/token.rs
@@ -69,11 +69,15 @@ impl Auth {
             Some(part) => Username::from_str(part)?,
             None => return Err("no username found when parsing auth token".to_owned()),
         };
-        let password = match parts.next() {
-            Some(part) => part.to_owned(),
-            None => return Err("no password found when parsing auth token".to_owned()),
-        };
-        Ok(Auth { username, password })
+        let remaining: Vec<&str> = parts.collect();
+        if remaining.is_empty() {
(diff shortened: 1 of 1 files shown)


crates/interledger-service/src/auth/token.rs
```diff
@@ -69,9 +69,10 @@ impl Auth {
             Some(part) => Username::from_str(part)?,
             None => return Err("no username found when parsing auth token".to_owned()),
         };
-        let password = match parts.next() {
-            Some(part) => part.to_owned(),
-            None => return Err("no password found when parsing auth token".to_owned()),
+        let remaining: Vec<&str> = parts.collect();
+        let password = match remaining.len() {
```
emschwartz: Instead of using a match, it would be clearer to write:
```
if remaining.is_empty() {
  Ok(Auth { username, password })
} else {
  Err("no password found when parsing auth token".to_owned())
}
```

dora-gt: Ooops, I wrote so ugly code. Fixed.

PR #519, opened 2019-11-19
Bs sqlite2

bstrie:


crates/ilp-node/Cargo.toml
```diff
@@ -21,7 +21,7 @@ clap = { version = "2.33.0", default-features = false }
 config = { version = "0.9.3", default-features = false, features = ["json", "toml", "yaml"] }
 futures = { version = "0.1.29", default-features = false }
 hex = { version = "0.4.0", default-features = false }
-interledger = { path = "../interledger", version = "^0.6.0", default-features = false, features = ["node"] }
+interledger = { path = "../interledger", version = "^0.6.0", default-features = false, features = ["node", "store-redis", "store-sqlite"] }
```
emschwartz: nit: just call the features `redis` and `sqlite`

crates/interledger-store-sqlite/src/account.rs
```diff
@@ -0,0 +1,319 @@
...
+use std::str::{self, FromStr};
+use uuid::{parser::ParseError, Uuid};
+
+use url::Url;
+
+use secrecy::ExposeSecret;
+use secrecy::SecretBytes;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Account {
```
emschwartz: Can this go in common?

crates/interledger-store-sqlite/src/account.rs
```diff
@@ -0,0 +1,319 @@
...
+
+        self.account
+    }
+}
+
+// Rust does not allow implementing foreign traits on foreign data types.
+// As a result, we wrap Uuid in a local data type, and implement the necessary
+// traits for that.
+#[derive(Eq, PartialEq, Hash, Debug, Default, Serialize, Deserialize, Copy, Clone)]
+pub struct AccountId(Uuid);
```
emschwartz: I think you made a good point about the `AccountId` not needing to be generic so I think it would be fine (maybe in a separate PR) to change the base `Account` type to always use a UUID.

crates/interledger-store-sqlite/src/crypto.rs
```diff
@@ -0,0 +1,175 @@
+use bytes::Bytes;
```
emschwartz: This is also going to go in common, right?

crates/interledger-store-sqlite/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+
+pub use store::{SqliteStore, SqliteStoreBuilder};
+
+#[cfg(test)]
+mod tests {
+    use rusqlite::Connection;
+
+    #[test]
+    fn open() {
+        assert!(Connection::open_in_memory().is_ok());
```
emschwartz: Don't think this should be in the final submission

crates/ilp-node/src/node.rs
```diff
@@ -236,6 +237,19 @@ impl InterledgerNode {
         } else {
             Either::B(self.serve_node())
         }
+        // XXX
+        .join({
+            let redis_scheme = true; // TODO: best way of parsing URL from Connection Info
+            if redis_scheme {
+                let redis_secret = generate_redis_secret(&self.secret_seed);
+                let store = RedisStoreBuilder::new(self.redis_connection.clone(), redis_secret);
+                Either::A(self.chain_services(store))
```
emschwartz: Will this method (`chain_services`) work? I know you said there were problems getting the types to line up before. If it works, it seems like a reasonable way to go

bstrie:
Sorry, pushed by mistake, disregard for now. :P

PR #526, opened 2019-11-21
Adding rejectcounter to STREAM client

aanchal4:
Adding rejectcounter to STREAM client to keep track of rejected packets, so that it errors out nicely when the sender receives more that 50 continuous reject packets. Otherwise it gets into infinite loop of sending packets.


crates/interledger-stream/src/client.rs
```diff
@@ -19,6 +19,7 @@ use std::{
     str,
     time::{Duration, SystemTime},
 };
+const REJECT_COUNTER: u8 = 50;
```
emschwartz: This should be named something like `MAX_CONSECUTIVE_REJECTS` (the constant is the limit, not the counter itself)

crates/interledger-stream/src/client.rs
```diff
@@ -95,6 +96,7 @@ where
                 sequence: 1,
                 rejected_packets: 0,
                 error: None,
+                rejectcounter: 0,
```
emschwartz: In Rust, all variables and fields should be written in snake_case. Also, it's hard to understand the difference between this and `rejected_packets` so I think this one should be called `consecutive_rejects` and the other could be renamed to `total_rejects`

crates/interledger-stream/src/error.rs
```diff
@@ -8,4 +8,6 @@ pub enum Error {
     PollError(String),
     #[fail(display = "Error polling: {}", _0)]
     SendMoneyError(String),
+    #[fail(display = "Error connecting: {}", _0)]
+    RejectPacketsError(String),
```
emschwartz: nit: I would call this `TooManyRejectedPacketsError` or `TooManyTemporaryErrors` so it's clear what exactly the problem was (the sender can handle some rejects)

We could also just roll this into the `SendMoneyError`, since that's the error you'd get if you get a FXX error code -- though maybe it's better to split out the error types.

crates/interledger-stream/src/error.rs
```diff
@@ -8,4 +8,6 @@ pub enum Error {
     PollError(String),
     #[fail(display = "Error polling: {}", _0)]
     SendMoneyError(String),
+    #[fail(display = "Error connecting: {}", _0)]
```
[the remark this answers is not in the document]
emschwartz: Agreed. It should say something more like `Too many rejected packets: {}`

crates/interledger-stream/src/client.rs
aanchal4: Agreed and fixed

crates/interledger-stream/src/client.rs
aanchal4: Agreed and fixed

aanchal4: Also, I do not have a good rationale for `MAX_CONSECUTIVE_REJECTS` to be 50. Any insights on that?

aanchal4:
We do not think that capping the number of consecutive reject packets is the best way to deal with this issue.

kincaidoneil:
> We do not think that capping the number of consecutive reject packets is the best way to deal with this issue.

Could you elaborate more on why?

I imagine we still want a timeout as a catch-all (?), but I like the consecutive rejects approach since I think in many cases it'd fail faster (which is a better UX if the payment is never going to go through).

PR #527, opened 2019-11-21
refactor: split parts of store-redis into new store-common crate

bstrie:
In preparation for future store crates to make use of.


bstrie:
An alternative design to consider here would be to have just one `interledger-store` crate, and then move `interledger-redis` and the future `interledger-sqlite` to just being submodules of `interledger-store`.

bstrie:
Closing for now since we're going to go with the design where we have just one store crate, will resubmit when that's done.

PR #528, opened 2019-11-22
ci(circleci): fix to make a release when the branch is `master`

dora-gt:
I'm sorry but there was a bug in CI.

Currently the variable is referring to `CIRCLE_TAG`, which is not correct. It should be
`CIRCLE_BRANCH` if we want to make a release on the `master` branch.


dora-gt:
I'm merging this because this should be merged before #530.

PR #529, opened 2019-11-22
refactor: renamed interledger-store crate in prep for new stores

bstrie:


bstrie:
Pushed with the nit fixed, let's see if CI fails again (I couldn't manage to reproduce on my machine).

Issue #359, opened 2019-09-30
Document usage of redis-cell for rate limiting


emschwartz:
The `amount_per_minute_limit` and `packets_per_minute_limit` are enforced by an external redis module called [`redis-cell`](https://github.com/brandur/redis-cell). If you don't load that module in, the rate limiting call will fail and the node will return an unhelpful `T00` error.

We should provide some recommended redis configuration in the main README or wherever we explain how to set up and run a node. See [here](https://github.com/interledger-rs/interledger-rs/blob/0d098ab7421d6abedb5e367c68b190e9f2cfe121/crates/interledger-store-redis/redis-example.conf#L15-L16) for how it's loaded in for the tests.

PR #532, opened 2019-11-25
docs(readme): add an explanation for redis-cell

dora-gt:
Fix: #359

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ docker run -it interledgerrs/settlement-engines
 #### Prerequisites
 
 - Git
+- [Redis](https://redis.io/)
 - [Rust](https://www.rust-lang.org/tools/install) - latest stable version
 
 #### Install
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,6 +85,22 @@ Append the `--help` flag to see available options.
 
 See [configuration](./docs/configuration.md) for more details on how the node is configured.
 
+#### Configuration of Redis
+
+Because we utilize [redis-cell](https://github.com/brandur/redis-cell), you need to load the redis-cell module as follows.
```
emschwartz: Technically, this is only required if you use the rate limiting parameters on any of the accounts. If those aren't used, it won't cause a problem if redis-cell isn't loaded

dora-gt: I see. Hmmm, I think we should mention it but basically I think it would be recommended to load the module because admins might add accounts with rate limits regardless of whether the node has the module or not.

I'll add an explanation for this.

dora-gt: Updated.

PR #533, opened 2019-11-26
Fix dos

aanchal4:
This is an alternative (potentially better) solution based on MAX_TIME since last fulfill or start to the problem in PR #526

### DIFF
--- a/crates/interledger-stream/src/client.rs
+++ b/crates/interledger-stream/src/client.rs
@@ -17,8 +17,12 @@ use std::{
     cell::Cell,
     cmp::min,
     str,
-    time::{Duration, SystemTime},
+    time::{Duration, Instant, SystemTime},
 };
+
+// Maximum time we should wait since last fulfill before we error out to avoid
+// getting into an infinite loop of sending packets and effectively DoSing ourselves
+const MAX_TIME_SINCE_LAST_FULFILL: Duration = Duration::from_secs(30);
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
 pub struct StreamDelivery {
@@ -95,6 +99,7 @@ where
                 sequence: 1,
                 rejected_packets: 0,
                 error: None,
+                last_fulfill_time: Instant::now(),
             }
         })
 }
@@ -114,6 +119,7 @@ struct SendMoneyFuture<S: IncomingService<A>, A: Account> {
     sequence: u64,
     rejected_packets: u64,
     error: Option<Error>,
+    last_fulfill_time: Instant,
 }
 
 struct PendingRequest {
@@ -136,10 +142,9 @@ where
     A: Account,
 {
     fn try_send_money(&mut self) -> Result<bool, Error> {
-        // Fire off requests until the congestion controller tells us to stop or we've sent the total amount
+        // Fire off requests until the congestion controller tells us to stop or we've sent the total amount or maximum time since last fulfill has elapsed
         let mut sent_packets = false;
         loop {
-            // Determine the amount to send
             let amount = min(
                 self.source_amount,
                 self.congestion_controller.get_max_amount(),
@@ -279,6 +284,7 @@ where
         // TODO should we check the fulfillment and expiry or can we assume the plugin does that?
         self.congestion_controller.fulfill(amount);
         self.should_send_source_account = false;
+        self.last_fulfill_time = Instant::now();
 
         if let Ok(packet) = StreamPacket::from_encrypted(&self.shared_secret, fulfill.into_data()) {
             if packet.ilp_packet_type() == IlpPacketType::Fulfill {
@@ -385,6 +391,12 @@ where
         // TODO maybe don't have loops here and in try_send_money
         loop {
             self.poll_pending_requests()?;
+            if self.last_fulfill_time.elapsed() >= MAX_TIME_SINCE_LAST_FULFILL {
+                return Err(Error::TimeoutError(format!(
+                    "Time since last fulfill exceeded the maximum time limit of {:?} secs",
+                    self.last_fulfill_time.elapsed().as_secs()
+                )));
+            }
 
             if self.source_amount == 0 && self.pending_requests.get_mut().is_empty() {
                 if self.state == SendMoneyFutureState::SendMoney {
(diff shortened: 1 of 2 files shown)


crates/interledger-stream/src/client.rs
```diff
@@ -17,9 +17,12 @@ use std::{
     cell::Cell,
     cmp::min,
     str,
-    time::{Duration, SystemTime},
+    time::{Duration, SystemTime, Instant},
 };
 
+//May be we should have different thresholds for MAX_TIME_SINCE_START and MAX_TIME_SINCE_LASTFULFILL
+const MAX_TIME: Duration = Duration::from_secs(30);
```
emschwartz: This kind of constant name should be more specific. Max time for what?

crates/interledger-stream/src/client.rs
```diff
@@ -17,9 +17,12 @@ use std::{
     cell::Cell,
     cmp::min,
     str,
-    time::{Duration, SystemTime},
+    time::{Duration, SystemTime, Instant},
 };
 
+//May be we should have different thresholds for MAX_TIME_SINCE_START and MAX_TIME_SINCE_LASTFULFILL
```
emschwartz: Why?

crates/interledger-stream/src/client.rs
```diff
@@ -95,6 +98,7 @@ where
                 sequence: 1,
                 rejected_packets: 0,
                 error: None,
+                time_counter_start: Instant::now(),
```
emschwartz: What is this a time counter for? (the name of the variable should make that clear)

crates/interledger-stream/src/client.rs
```diff
@@ -136,10 +141,17 @@ where
     A: Account,
 {
     fn try_send_money(&mut self) -> Result<bool, Error> {
-        // Fire off requests until the congestion controller tells us to stop or we've sent the total amount
+        // Fire off requests until the congestion controller tells us to stop or we've sent the total amount or maximum time has elapsed
```
emschwartz: Maximum time since when

crates/interledger-stream/src/client.rs
```diff
@@ -136,10 +141,17 @@ where
...
         let mut sent_packets = false;
+        std::println!("I am outside");
```
emschwartz: Remove all print statements

crates/interledger-stream/src/client.rs
emschwartz: Maximum time since when

crates/interledger-stream/src/client.rs
```diff
@@ -136,10 +141,17 @@ where
...
         loop {
-            // Determine the amount to send
+            if self.time_counter_start.elapsed() >= MAX_TIME {
+                return Err(Error::MaxTimeExceeded(format!(
+                    "Time since start exceeded the maximum time limit of {:?} secs",
```
emschwartz: This is really the time since the last fulfill so the message should make that clear

crates/interledger-stream/src/client.rs
```diff
@@ -136,10 +141,17 @@ where
     A: Account,
 {
     fn try_send_money(&mut self) -> Result<bool, Error> {
-        // Fire off requests until the congestion controller tells us to stop or we've sent the total amount
+        // Fire off requests until the congestion controller tells us to stop or we've sent the total amount or maximum time has elapsed
         let mut sent_packets = false;
+        std::println!("I am outside");
         loop {
-            // Determine the amount to send
+            if self.time_counter_start.elapsed() >= MAX_TIME {
```
emschwartz: `try_send_money` is called by `poll` so it seems like this check should only be in one of those functions

crates/interledger-stream/src/error.rs
```diff
@@ -8,4 +8,7 @@ pub enum Error {
     PollError(String),
     #[fail(display = "Error polling: {}", _0)]
     SendMoneyError(String),
+    #[fail(display = "Error maximum time exceeded: {}", _0)]
+    MaxTimeExceeded(String),
```
emschwartz: nit: could just be `TimeoutError`

crates/interledger-stream/src/client.rs
```diff
@@ -17,9 +17,12 @@ use std::{
     cell::Cell,
     cmp::min,
     str,
-    time::{Duration, SystemTime},
+    time::{Duration, Instant, SystemTime},
 };
 
+//May be we should have different thresholds for MAX_TIME_SINCE_START and MAX_TIME_SINCE_LASTFULFILL
+const LASTFULFILL_MAX_TIME: Duration = Duration::from_secs(30);
```
emschwartz: nit: there should be an underscore between `LAST` and `FULFILL`. Also, I think something like `MAX_TIME_SINCE_LAST_FULFILL` would be clearer

It might be useful to include a comment above this constant explaining what this value is for and why we included the timeout

crates/interledger-stream/src/client.rs
emschwartz: This comment should be removed

crates/interledger-stream/src/client.rs
```diff
@@ -95,6 +98,7 @@ where
                 sequence: 1,
                 rejected_packets: 0,
                 error: None,
+                lastfulfill_start_time: Instant::now(),
```
emschwartz: Why is this called "last fulfill start time"? It doesn't seem like it has anything to do with "starting". Maybe just `last_fulfill_time`

crates/interledger-stream/src/client.rs
aanchal4: I thought of `MAX_TIME_SINCE_LAST_FULFILL` but then it's 5 letters, which is too long.

emschwartz: 5 is on the cusp for me. Better to have 5 and make it clearer than skimp and make it confusing. 6 is where I draw the line

crates/interledger-stream/src/client.rs
```diff
@@ -17,9 +17,13 @@ use std::{
     cell::Cell,
     cmp::min,
     str,
-    time::{Duration, SystemTime},
+    time::{Duration, Instant, SystemTime},
 };
 
+//Maximum time we should wait since last fulfill before we error out to avoid
```
emschwartz:
```suggestion
// Maximum time we should wait since last fulfill before we error out to avoid
```

PR #535, opened 2019-11-26
refactor: make `AccountId` no longer an associated type

bstrie:
As discussed at RustFest, as a limited alternative to removing the `Account` trait.


bstrie:
Because `AccountId` was formerly so tightly coupled with Redis-specific concerns, this PR also includes several commits that continue to disentangle Redis from the rest of the store crate (as will be necessary for the future SQLite store, which is related to this PR).

crates/interledger-service/src/lib.rs
```diff
@@ -36,8 +37,39 @@ use std::{
     str::FromStr,
     sync::Arc,
 };
+use uuid::{parser::ParseError, BytesError, Uuid};
 
-use serde::Serialize;
+#[derive(Eq, PartialEq, Hash, Debug, Default, Serialize, Deserialize, Copy, Clone)]
+pub struct AccountId(pub Uuid);
+
+impl AccountId {
```
emschwartz: What's the point of having an `AccountId` struct if it's just going to wrap a UUID? Is there a benefit over just using the UUID struct directly?

crates/interledger-settlement/src/api/fixtures.rs
```diff
@@ -13,10 +14,14 @@ pub static IDEMPOTENCY: &str = "AJKJNUjM0oyiAN46";
...
+    pub static ref TEST_ACCOUNT_0: TestAccount = TestAccount::new(
+        AccountId::from_slice(&[0; 16]).unwrap(),
+        "http://localhost:1234",
+        "example.account"
+    );
     pub static ref SERVICE_ADDRESS: Address = Address::from_str("example.connector").unwrap();
-    pub static ref MESSAGES_API: Matcher = Matcher::Regex(r"^/accounts/\d*/messages$".to_string());
+    pub static ref MESSAGES_API: Matcher =
+        Matcher::Regex(r"^/accounts/[0-9a-fA-F-]*/messages$".to_string());
     pub static ref SETTLEMENT_API: Matcher =
```
emschwartz: nit: these regexes should probably have `+` instead of `*` because `/accounts//messages` isn't valid

crates/interledger-service/src/lib.rs
bstrie: In this case it was simply because it made the migration easier. Now that it's done I wouldn't be opposed to getting rid of it. (I suppose one theoretical benefit would be to decouple our public interface from the uuid crate, but we already export a uuid error type here so we're not currently achieving that anyway).

crates/interledger-settlement/src/api/fixtures.rs
bstrie: It occurs to me that this could actually be something much stricter like `[0-9A-fa-f]{8}-[0-9A-fa-f]{4}-[0-9A-fa-f]{4}-[0-9A-fa-f]{4}-[0-9A-fa-f]{12}`, if we intend to only support UUIDs.

crates/interledger-service/src/lib.rs
emschwartz: I'll leave this one up to you. I could go either way on this

bstrie: I don't think the newtype is particularly pulling its weight here, I've pushed a commit getting rid of it.

PR #536, opened 2019-11-27
Improve SPSP client errors when SPSP server response is invalid

emschwartz:


crates/interledger-spsp/src/lib.rs
```diff
@@ -16,12 +16,13 @@ mod server;
 pub use client::{pay, query};
 pub use server::SpspResponder;
 
+// TODO should these error variants be renamed to remove the 'Error' suffix from each one?
```
bstrie: I would say to absolutely leave the "Error" suffix, because the debug-print output of a variant only includes the variant name and not the enum, so someone encountering an error is going to see `HttpError("foo")` rather than `spsp::Http::Stream("foo")`, and seeing just `Http("foo)` seems undesirable.

crates/interledger-spsp/src/client.rs
```diff
@@ -17,9 +17,13 @@ pub fn query(server: &str) -> impl Future<Item = SpspResponse, Error = Error> {
         .header("Accept", "application/spsp4+json")
         .send()
         .map_err(|err| Error::HttpError(format!("Error querying SPSP receiver: {:?}", err)))
+        .and_then(|res| {
+            res.error_for_status()
+                .map_err(|err| Error::HttpError(format!("Error querying SPSP receiver: {:?}", err)))
+        })
```
bstrie: I assume the goal here is that returning a 404 response isn't currently considered an `Err`, so we manually err if we get a non-200 response?

bstrie:
Heh, the CI error...
```
   = note: consider adding a `#![type_length_limit="1119075"]` attribute to your crate
```
that's 600,000 less than last time! 🎉

crates/interledger-spsp/src/client.rs
gakonst: Yeah that's exactly the case, previously it'd return OK even if the server returned an error

emschwartz:
@bstrie the fact that this PR hit the type length limit suggests that each future combinator is being added to the type name (this doesn't add another level of services, only another `and_then` on a future)

For some reason, boxing the futures for the `spsp::client::{query, pay}` methods doesn't seem to help.

bstrie:
> the fact that this PR hit the type length limit suggests that each future combinator is being added to the type name

That's the behavior that I expected, and why I surmise that switching to async/await could resolve this on its own by allowing us to stop using the combinators entirely.

I'm not sure why boxing doesn't work, but if the goal of the box here is to keep the compiler from ever seeing any type that's of an extreme length, then boxes would need to also be inserted in places other than at the very end, in order to keep the type of intermediate values to a resonable length.

Issue #437, opened 2019-10-21
ILP address of `Child` node is not working properly.


dora-gt:
If we set the `ilp_address` parameter when spinning up a `Child` node, it causes a weird behavior on the node. For example, if we set `ILP_ADDRESS=example.bob.charlie` for Charlie's node in ETH + XRP three nodes example, `Charlie` receives amount `0` for a payment of `500`.

This seems undesired behavior, so this should be fixed.

Ref:
- https://forum.interledger.org/t/receiver-is-sending-xrp-back-in-the-cross-eth-xrp-3-nodes-setup/814
- https://github.com/interledger-rs/interledger-rs/pull/436

PR #538, opened 2019-11-28
Clarifications on ILP Address of Child nodes

gakonst:
Closes https://github.com/interledger-rs/interledger-rs/issues/437.

I've also tested that manually adding `--ilp_address example.bob.charlie` when starting up the node does not affect the tests

### DIFF
--- a/examples/eth-xrp-three-nodes/README.md
+++ b/examples/eth-xrp-three-nodes/README.md
@@ -449,9 +449,11 @@ ilp-node \
 --settlement_api_bind_address 127.0.0.1:8771 \
 &> logs/node-bob.log &
 
-# Start Charlie's node 
-# You can add `--ilp_address example.bob.charlie` but it is not necessary
-# since the address will get fetched via ILDCP request when Bob is added as a parent.
+# Start Charlie's node. The --ilp_address field is omitted, so the node's address will 
+# be `local.host`. When a parent account is added, Charlie's node will send an ILDCP request
+# to them, which they will respond to with an address that they have assigned to Charlie.
+# Charlie will then proceed to set that as their node's address.
+# ILDCP documentation: https://interledger.org/rfcs/0031-dynamic-configuration-protocol/
 ilp-node \
 --secret_seed 1232362131122139900555208458637022875563691455429373719368053354 \
 --admin_auth_token hi_charlie \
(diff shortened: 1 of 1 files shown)


examples/eth-xrp-three-nodes/README.md
```diff
@@ -449,7 +449,9 @@ ilp-node \
 --settlement_api_bind_address 127.0.0.1:8771 \
 &> logs/node-bob.log &
 
-# Start Charlie's node
+# Start Charlie's node 
+# You can add `--ilp_address example.bob.charlie` but it is not necessary
+# since the address will get fetched via ILDCP request when Bob is added as a parent.
```
emschwartz: Maybe link to the ILDCP spec here

examples/eth-xrp-three-nodes/README.md
```diff
@@ -553,9 +543,17 @@ ilp-cli --node http://localhost:8770 accounts create charlie \
     --settle-to 0 \
     --routing-relation Child &> logs/account-bob-charlie.log &
 
-# We have to wait here to ensure that the parent account is created because
-# the child account tries to acquire its ILP address from the parent account.
-sleep 2
+printf "Adding Charlie's Account...\n"
+# there is no need to specify an address, it will get added as localhost.charlie 
+# initially, and then it will be converted to example.bob.charlie
```
emschwartz: When / how will it be "converted"?

gakonst: Once a parent is added, the Node's ILP Address gets updated: https://github.com/interledger-rs/interledger-rs/blob/54c3c2bb56aada603f864377a88a550c4f9a4764/crates/interledger-api/src/routes/accounts.rs#L524

Then, each account's address is updated: https://github.com/interledger-rs/interledger-rs/blob/1381e173dd70803a84551c64891346ea0336b0dc/crates/interledger-store/src/redis/mod.rs#L1320-L1359

emschwartz: Sorry I actually meant that you should explain that in the comment. It's not clear from the text what you mean

gakonst: Expanded on the comment and added link to the code inline

examples/eth-xrp-three-nodes/README.md
```diff
@@ -553,11 +545,22 @@ ilp-cli --node http://localhost:8770 accounts create charlie \
...
+    --auth hi_charlie \
+    --asset-code XRP \
+    --asset-scale 6 \
+    --max-packet-amount 100 \
+    --ilp-over-http-incoming-token charlie_password \
+    --ilp-over-http-url http://localhost:9770/ilp \
+    --settle-to 0 &> logs/account-charlie-charlie.log
 
 printf "Adding Bob's account on Charlie's node (XRP Parent relation)...\n"
+# Once a parent is added, the Node's ILP Address gets updated: https://github.com/interledger-rs/interledger-rs/blob/54c3c2bb56aada603f864377a88a550c4f9a4764/crates/interledger-api/src/routes/accounts.rs#L524
```
emschwartz: I wouldn't include commit hashes in the documentation because it's likely to become out of date and not get updated. I would just explain how it happens or point to some other documentation about it

Issue #499, opened 2019-11-05
Wrong error code (F08) used when request.prepare.amount is too small


latenssi:
In [exchange_rates_service](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-service-util/src/exchange_rates_service.rs) if the `request.prepare.amount` is so small that the conversion to `outgoing_amount` results in a float between 0 and 1 the `exchange_rates_service` will return **Error F08 "Amount Too Large"** as specified on [this line](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-service-util/src/exchange_rates_service.rs#L148). 

I think this is misleading. In my opinion the result in this case should be **Error F03 "Invalid Amount"** with an explanation as this happens when one tries to send an amount too small to be converted and relayed.

kincaidoneil:
I think this case should return `R01`, Insufficient Source Amount.

Per the [ILPv4 spec](https://interledger.org/rfcs/0027-interledger-protocol-4/):

> The amount received by a connector in the path was too little to forward (zero or less). Either the sender did not send enough money or the exchange rate changed. The sender MAY try again with a higher amount, but they SHOULD NOT do this indefinitely or a malicious connector could steal money from them.

latenssi:
@kincaidoneil I agree, that is a better error code

bstrie:
Copying Evan's comment from #540:

> I don't think that's the right error. The problem here is that the amount is bigger than the max u64. You'd definitely want to communicate to the sender that they need to send a smaller amount in order for the packet to go through

kincaidoneil:
@gakonst @bstrie @latenssi Didn't see this was caused by an overflow. I think your original suggestion of F03 makes more sense then!

gakonst:
To be clear: The OP's statement "the  request.prepare.amount is so small that the conversion to outgoing_amount results in a float between 0 and 1" is not correct, this happens because when Rust casts a large f64 to u64, it zeroes it, if it was larger than what can fit in u64.

PR #540, opened 2019-11-28
Return proper error code in precision loss edge case in exchange rates service

gakonst:
Fixes https://github.com/interledger-rs/interledger-rs/issues/499

### DIFF
--- a/crates/interledger-service-util/src/exchange_rates_service.rs
+++ b/crates/interledger-service-util/src/exchange_rates_service.rs
@@ -145,13 +145,29 @@ where
                     // is larger than the maximum value for a u64.
                     // When it gets cast to a u64, it will end up being 0.
                     if outgoing_amount != 0.0 && outgoing_amount as u64 == 0 {
-                        return Box::new(err(RejectBuilder {
-                            code: ErrorCode::F08_AMOUNT_TOO_LARGE,
-                            message: format!(
-                                "Could not cast outgoing amount to u64 {}",
-                                outgoing_amount,
+                        let (code, message) = if outgoing_amount < 1.0 {
+                            // user wanted to send a positive value but it got rounded down to 0
+                            (
+                                ErrorCode::R01_INSUFFICIENT_SOURCE_AMOUNT,
+                                format!(
+                                    "Could not cast to f64, amount too small: {}",
+                                    outgoing_amount
+                                ),
                             )
-                            .as_bytes(),
+                        } else {
+                            // amount that arrived was too large for us to forward
+                            (
+                                ErrorCode::F08_AMOUNT_TOO_LARGE,
+                                format!(
+                                    "Could not cast to f64, amount too large: {}",
+                                    outgoing_amount
+                                ),
+                            )
+                        };
+
+                        return Box::new(err(RejectBuilder {
+                            code,
+                            message: message.as_bytes(),
                             triggered_by: Some(&ilp_address),
                             data: &[],
                         }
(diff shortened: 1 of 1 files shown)


crates/interledger-service-util/src/exchange_rates_service.rs
```diff
@@ -146,7 +146,7 @@ where
                     // When it gets cast to a u64, it will end up being 0.
                     if outgoing_amount != 0.0 && outgoing_amount as u64 == 0 {
                         return Box::new(err(RejectBuilder {
-                            code: ErrorCode::F08_AMOUNT_TOO_LARGE,
+                            code: ErrorCode::R01_INSUFFICIENT_SOURCE_AMOUNT,
```
emschwartz: I don't think that's the right error. The problem here is that the amount is bigger than the max u64. You'd definitely want to communicate to the sender that they need to send a smaller amount in order for the packet to go through

gakonst: PR was made based on: https://github.com/interledger-rs/interledger-rs/issues/499#issuecomment-550015494

emschwartz: Ah my mistake, this is when it gets rounded down to 0. Maybe R01 is the right error in this case

gakonst: @emschwartz So the way Rust handles f64 being cast to u64 when it doesn't fit, is that it gets zero'ed out. 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c81a1a7a2cc882ea787fc7e7fe6b01d3

that said, insufficient source amount is definitely not correct, as it's the result of overflow. My follow up commit, made that an `INVALID_AMOUNT` error, which seems like it makes more sense.

emschwartz: Ok if it's the result of an overflow, `F08_AMOUNT_TOO_LARGE` makes more sense to me than `F03_INVALID_AMOUNT`. The problem is not that the sender sent an invalid number, it's that they sent enough to cause us to have an overflow and they should try again with a smaller amount. `F08` seems more helpful

crates/interledger-service-util/src/exchange_rates_service.rs
```diff
@@ -146,7 +146,7 @@ where
                     // When it gets cast to a u64, it will end up being 0.
                     if outgoing_amount != 0.0 && outgoing_amount as u64 == 0 {
```
emschwartz: @gakonst #499 points at the issue when the amount gets rounded down to 0 because it's less than 1. I think there should be two checks here.

```
if outgoing_amount != 0.0 && outgoing_amount as u64 == 0 {
  if outgoing_amount < 1.0 {
    // user wanted to send a positive value but it got rounded down to 0
    return R01
  } else {
    // amount that arrived was too large for us to forward
    return F08
  }
}
```

}

gakonst: OK I agree with this

crates/interledger-service-util/src/exchange_rates_service.rs
```diff
@@ -145,13 +145,29 @@ where
...
+                            // user wanted to send a positive value but it got rounded down to 0
+                            (
+                                ErrorCode::R01_INSUFFICIENT_SOURCE_AMOUNT,
+                                format!(
+                                    "Could not cast to f64, amount too small: {}",
+                                    outgoing_amount
+                                ),
+                            )
+                        } else {
+                            // amount that arrived was too large for us to forward
```
emschwartz:
```suggestion
                            // amount that arrived was too large and the overflowing float was
                            // converted to 0 when it was cast to a u64
```

crates/interledger-service-util/src/exchange_rates_service.rs
```diff
@@ -145,13 +145,29 @@ where
...
+                            (
+                                ErrorCode::F08_AMOUNT_TOO_LARGE,
```
kincaidoneil: AFAIK, this case would probably occur when a user accidentally fat fingers wayyyyy too much _and_ this connector doesn't have any max packet amount set (which would already be caught earlier in the pipeline), right?

By returning an F08 error without any associated max packet data, their STREAM sender will retry, decreasing the packet amount by a factor of 2. Eventually, it'd try an amount that'd convert below the max u64, but still a massive amount (!), and that poor user might lose money if they have automatic settlement or something.

Alternatively, if we returned an F03, their STREAM sender wouldn't have that behavior, which might provide a better sanity check in this (admittedly low probability) case.

TL;DR maybe this should be a "final final" error instead of a "retryable final" error.

emschwartz: I do not think this is the place to handle such an issue. There should be enough confirmation steps or sanity checks in the sender's client to prevent them sending a ridiculous amount.

Issue #376, opened 2019-10-02
Balance API should include asset details


emschwartz:
It's hard to know what you're looking at when you just get back the plain number. It might be nice to include the `asset_code` and `asset_scale` in the response

kincaidoneil:
Better yet, also provide a human-readable string? e.g `0.00348 XRP`?

emschwartz:
Do you think it should return a single string like that or are you suggesting combining the `balance` and `asset_scale` to just return a decimal number?

kincaidoneil:
> Do you think it should return a single string like that or are you suggesting combining the balance and asset_scale to just return a decimal number?

Mmmmm, probably just combining the balance and asset_scale to a decimal number, and leaving the code as a separate field.

PR #543, opened 2019-11-29
Show normalized amount and asset code when fetching an account's balance

gakonst:
Fixes https://github.com/interledger-rs/interledger-rs/issues/376

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,6 +82,9 @@ jobs:
           name: Run run-md Test
           command: |
             scripts/run-md-test.sh
+          environment:
+            # run scripts in source mode so that the branch's build is used
+            SOURCE_MODE: 1
       - store_artifacts:
           path: /tmp/run-md-test
           destination: run-md-test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,10 +81,7 @@ jobs:
       - run:
           name: Run run-md Test
           command: |
-            scripts/run-md-test.sh
-          environment:
-            # run scripts in source mode so that the branch's build is used
-            SOURCE_MODE: 1
+            scripts/run-md-test.sh '^.*$' 1
       - store_artifacts:
           path: /tmp/run-md-test
           destination: run-md-test
--- a/crates/ilp-node/tests/btp.rs
+++ b/crates/ilp-node/tests/btp.rs
@@ -155,8 +155,20 @@ fn two_nodes_btp() {
                         })
                         .and_then(move |_| {
                             get_balances().and_then(move |ret| {
-                                assert_eq!(ret[0], -1000);
-                                assert_eq!(ret[1], 1000);
+                                assert_eq!(
+                                    ret[0],
+                                    BalanceData {
+                                        asset_code: "XYZ".to_owned(),
+                                        balance: -1e-6
+                                    }
+                                );
+                                assert_eq!(
+                                    ret[1],
+                                    BalanceData {
+                                        asset_code: "XYZ".to_owned(),
+                                        balance: 1e-6
+                                    }
+                                );
                                 Ok(())
                             })
                         })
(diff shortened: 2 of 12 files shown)


crates/ilp-node/tests/test_helpers.rs
```diff
@@ -26,9 +26,10 @@ pub fn random_secret() -> String {
     hex::encode(bytes)
 }
 
-#[derive(serde::Deserialize)]
+#[derive(serde::Deserialize, Debug, PartialEq)]
```
dora-gt: I think this should not be tested by `eq` because it is a **float** value.
It should be tested with [`approx`](https://github.com/brendanzab/approx) or some appropriate ways.

crates/ilp-node/tests/btp.rs
```diff
@@ -168,8 +180,20 @@ fn two_nodes_btp() {
                         })
                         .and_then(move |_| {
                             get_balances().and_then(move |ret| {
-                                assert_eq!(ret[0], 1000);
-                                assert_eq!(ret[1], -1000);
+                                assert_eq!(
```
dora-gt: Comparing float values by `eq` is not a good idea though it might work in some particular situation like this.

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -278,17 +278,23 @@ where
...
                         })
-                })
-                .and_then(|balance: i64| {
-                    Ok(warp::reply::json(&json!({
-                        "balance": balance,
-                    })))
+                        .and_then(move |balance: i64| {
+                            Ok(warp::reply::json(&json!({
+                                // normalize to the base unit
+                                "balance": balance as f64 / 10_u64.pow(asset_scale.into()) as f64,
```
dora-gt: This is a breaking change, so I think it should be written in your commit comment.
It will be shown on the release notes.

examples/eth-settlement/README.md
```diff
@@ -2,10 +2,10 @@
 # For integration tests
 function post_test_hook() {
     if [ $TEST_MODE -eq 1 ]; then
-        test_equals_or_exit '{"balance":-500}' test_http_response_body -H "Authorization: Bearer alice:in_alice" http://localhost:7770/accounts/alice/balance
-        test_equals_or_exit '{"balance":0}' test_http_response_body -H "Authorization: Bearer bob:bob_password" http://localhost:7770/accounts/bob/balance
-        test_equals_or_exit '{"balance":0}' test_http_response_body -H "Authorization: Bearer alice:alice_password" http://localhost:8770/accounts/alice/balance
-        test_equals_or_exit '{"balance":500}' test_http_response_body -H "Authorization: Bearer bob:in_bob" http://localhost:8770/accounts/bob/balance
+        test_equals_or_exit '{"asset_code":"ETH","balance":-5e-16}' test_http_response_body -H "Authorization: Bearer alice:in_alice" http://localhost:7770/accounts/alice/balance
```
dora-gt: Thank you for fixing these.

crates/ilp-node/tests/btp.rs
gakonst: I think it should be fine in these cases? If we need to do more granular comparisons we can use approx then.

crates/interledger-api/src/routes/accounts.rs
bstrie: Do we have a conventional changelog prefix specifically for breaking changes? I've definitely made a few of those that I've only labeled as "refactor"...

.circleci/config.yml
```diff
@@ -82,6 +82,9 @@ jobs:
           name: Run run-md Test
           command: |
             scripts/run-md-test.sh
+          environment:
+            # run scripts in source mode so that the branch's build is used
+            SOURCE_MODE: 1
```
bstrie: Is this temporary or permanent?

crates/interledger-api/src/routes/accounts.rs
gakonst: I've added it like this, based on @dora-gt's previous commits:
image

.circleci/config.yml
gakonst: Replaced with https://github.com/interledger-rs/interledger-rs/pull/545/commits/c2232ffe70b81ef95cc3ed0477544e57738459e3

crates/ilp-node/tests/test_helpers.rs
bstrie: As long as this is only for tests and not something used for production I think the current approach suffices, since the operations are all still deterministic.

gakonst: Yeah that was my idea as well

dora-gt: OK, just wanted to mention.

crates/interledger-api/src/routes/accounts.rs
dora-gt: I remember we are asked by `cz` when we are committing some breaking changes.
Maybe gakonst's one is generated by that.

Issue #534, opened 2019-11-26
Routes update when deleting an account


leboiko:
When deleting an account on the node the routes are not being updated.

gakonst:
@leboiko could you give more context on the bug? Please provide a test case illustrating the behavior you are experiencing, and the behavior you expected to see!

emschwartz:
I'm going to close this now but feel free to reopen if someone can document how to reproduce this issue and describe the desired behavior

PR #548, opened 2019-12-02
fix(redis): delete the STATIC_ROUTES when removing an account

gakonst:
Closes https://github.com/interledger-rs/interledger-rs/issues/534.

Other than that, the routes are all deleted properly inside https://github.com/interledger-rs/interledger-rs/blob/1381e173dd70803a84551c64891346ea0336b0dc/crates/interledger-store/src/redis/mod.rs#L560-L586

### DIFF
--- a/crates/interledger-store/src/redis/mod.rs
+++ b/crates/interledger-store/src/redis/mod.rs
@@ -1358,6 +1358,13 @@ impl AddressStore for RedisStore {
                                 RedisAccountId(account.id()),
                             )
                             .ignore();
+
+                            pipe.hset(
+                                STATIC_ROUTES_KEY,
+                                new_ilp_address.as_bytes(),
+                                RedisAccountId(account.id()),
+                            )
+                            .ignore();
                         }
                     }
                     pipe.query_async(connection.clone())
--- a/crates/interledger-store/src/redis/mod.rs
+++ b/crates/interledger-store/src/redis/mod.rs
@@ -580,6 +580,9 @@ impl RedisStore {
             pipe.hdel(ROUTES_KEY, account.ilp_address.to_bytes().to_vec())
                 .ignore();
 
+            pipe.hdel(STATIC_ROUTES_KEY, account.ilp_address.to_bytes().to_vec())
+                .ignore();
+
             pipe.del(uncredited_amount_key(id));
 
             pipe.query_async(connection)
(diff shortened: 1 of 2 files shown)


crates/interledger-store/src/redis/mod.rs
```diff
@@ -1353,6 +1358,13 @@ impl AddressStore for RedisStore {
                                 RedisAccountId(account.id()),
                             )
                             .ignore();
+
+                            pipe.hset(
+                                STATIC_ROUTES_KEY,
```
emschwartz: This shouldn't be needed, because it's included in the "local routes"

gakonst: Are you sure? Removing it makes the test at https://github.com/interledger-rs/interledger-rs/blob/aba34b87a2b87fb9866af2e2902484455846a3f9/crates/interledger-store/tests/redis/routing_test.rs#L339-L342 fail. 

Consider the following: We're Bob, Alice is our parent, hence we're `example.alice.bob`. We've added Charlie as a child account, hence they are `example.alice.bob.charlie`, and we have added them as a static route (plus we also have a normal route to them since they're a child node). Assume we remove Alice, and replace her with Dylan. Now Charlie's address becomes: `example.dylan.bob.charlie`. His static route should also be changed, since static routes override normal routes. Otherwise, it'll still be the old route. Am I missing something?

emschwartz: Why would `Charlie` be added as a static route in the first place if they're a local account?

Also, it seems a bit strange to me that something in _static_ routes would ever be changed automatically

crates/interledger-store/src/redis/mod.rs
```diff
@@ -1319,8 +1322,10 @@ impl AddressStore for RedisStore {
                         if account.routing_relation() != RoutingRelation::Parent
                             && account.routing_relation() != RoutingRelation::Peer
                         {
-                            // remove the old route
+                            // remove the old route and static route (if exists)
                             pipe.hdel(ROUTES_KEY, &account.ilp_address as &str).ignore();
+                            pipe.hdel(STATIC_ROUTES_KEY, &account.ilp_address as &str)
```
emschwartz: I think the more important things to delete are any static routes that point to this account

crates/interledger-store/src/redis/mod.rs
gakonst: As a `child` account, we should still route things to Charlie. Is there any limitation which says that we should not add child accounts as static routes?

Here, the change in _static_ routes would happen only because the node's parent changed, which means that all routes (static as well) have to be adjusted for the new topology.

gakonst:
Closing this until the issue opener gives more context on what their exact issue was

PR #551, opened 2019-12-04
Make Redis fully optional for ilp-node

bstrie:
This should be the final part of laying the groundwork for alternative backing databases.


gakonst:
Do you think it'd be possible to abstract the E2E tests such that they run once with each available backend? So instead of putting everythign under ilp-node/tests/redis, we'd use ilp-node/tests (as is now), and then the test would:
1. import all backends 
2. put them in a trait object vector [trait probably would be NodeStore ])
3. run all tests for that backend (launch backend, launch node and plug it on the backend's driver, kill node and backend, move on to the next one)

bstrie:
> Do you think it'd be possible to abstract the E2E tests such that they run once with each available backend?

Yes, in the long run I would like to share this code between all the storage backends, since otherwise there would be a substantial amount of duplication. For now though I was planning on doing that later as part of the SQLite backend effort, once it's mature enough to start needing tests. At that point I intend to extract the generic bits out of the tests/redis directory, but since none of it's generic today (and I think that's lower priority than working on the SQLite store itself) I just shoved all of the tests into the redis-specific directory for the time being.

crates/ilp-node/src/main.rs
```diff
@@ -69,8 +71,8 @@ pub fn main() {
             .takes_value(true)
             .required(true)
             .help("HTTP Authorization token for the node admin (sent as a Bearer token)"),
-        Arg::with_name("redis_url")
-            .long("redis_url")
+        Arg::with_name("data_source_url")
```
emschwartz: I think this should be `database_url`

crates/ilp-node/src/node.rs
```diff
@@ -66,8 +72,15 @@ fn default_settlement_api_bind_address() -> SocketAddr {
...
     SocketAddr::from(([127, 0, 0, 1], 7770))
 }
-fn default_redis_url() -> ConnectionInfo {
-    DEFAULT_REDIS_URL.into_connection_info().unwrap()
+// We allow unreachable code on the below function because there must always be exactly one default
+// regardless of how many data sources the crate is compiled to support,
+// but we don't know which will be enabled or in which quantities or configurations.
+// This return-based pattern effectively gives us fallthrough behavior.
+#[allow(unreachable_code)]
+fn default_data_source_url() -> String {
```
emschwartz: As above, `database_url`

crates/ilp-node/src/node.rs
```diff
@@ -187,7 +185,7 @@ impl ExchangeRateConfig {
 }
 
 /// An all-in-one Interledger node that includes sender and receiver functionality,
-/// a connector, and a management API. The node uses Redis for persistence.
+/// a connector, and a management API.
```
emschwartz: Worth mentioning that it connects to the database specified (and maybe which databases are supported)

crates/ilp-node/src/node.rs
```diff
@@ -251,22 +245,82 @@ impl InterledgerNode {
         }
     }
 
-    #[allow(clippy::cognitive_complexity)]
-    fn serve_node(&self) -> impl Future<Item = (), Error = ()> {
-        let redis_secret = generate_redis_secret(&self.secret_seed);
-        let secret_seed = Bytes::from(&self.secret_seed[..]);
-        let http_bind_address = self.http_bind_address;
-        let settlement_api_bind_address = self.settlement_api_bind_address;
+    fn serve_node(self) -> Box<dyn Future<Item = (), Error = ()> + Send + 'static> {
```
emschwartz: Is the box here faster to compile than `impl Future...`?

crates/ilp-node/src/node.rs
```diff
@@ -251,22 +245,82 @@ impl InterledgerNode {
...
+            "redis" => Box::new(serve_redis_node(self, ilp_address)),
+            other => {
+                error!("unsupported data source scheme: {}", other);
+                Box::new(err(()))
+            }
+        }
+    }
+
+    #[allow(clippy::cognitive_complexity)]
+    pub fn chain_services<S>(
```
emschwartz: Can this be `pub(crate)` instead?

crates/ilp-node/src/node.rs
```diff
@@ -251,22 +245,82 @@ impl InterledgerNode {
...
+        self,
+        store: S,
+        ilp_address: Address,
+    ) -> impl Future<Item = (), Error = ()>
+    where
+        // Should we use a generic rather than the concrete Account type?
+        // I spent long enough banging my head against this that I don't care
+        // anymore, but feel free to take a whack at it.
```
emschwartz: If this method needs to be public, I'd feel slightly stronger about making it generic. I think it can be internal though, in which case using the concrete type seems fine (since the stores we actually support will use that type).

crates/ilp-node/src/redis_store.rs
```diff
@@ -0,0 +1,86 @@
...
+use tracing::{debug, debug_span, error, info};
+use uuid::Uuid;
+
+static REDIS_SECRET_GENERATION_STRING: &str = "ilp_redis_secret";
+
+pub fn default_redis_url() -> String {
+    String::from("redis://127.0.0.1:6379")
+}
+
+pub fn serve_redis_node(
```
emschwartz: It seems a little odd for this to use a function from the node and be called by the node. Could you include a comment explaining why it's set up this way? (Presumably so that it's turned on/off with the redis feature flag)

crates/ilp-node/src/redis_store.rs
```diff
@@ -0,0 +1,86 @@
...
+        &hmac::Key::new(hmac::HMAC_SHA256, secret_seed),
+        REDIS_SECRET_GENERATION_STRING.as_bytes(),
+    );
+    redis_secret.copy_from_slice(sig.as_ref());
+    redis_secret
+}
+
+#[doc(hidden)]
+#[allow(dead_code)]
+pub fn insert_account_with_redis_store(
```
emschwartz: Is this only used in the tests?

crates/ilp-node/tests/redis/accounts_api.rs
```diff
@@ -48,16 +45,17 @@ fn accounts_test() {
     let mut runtime = RuntimeBuilder::new().build().unwrap();
 
     let redis_test_context = TestContext::new();
-    let node_http_port = get_open_port(Some(7770));
+    let node_http_port = get_open_port(None);
```
emschwartz: nit: any specific reason to change this?

crates/ilp-node/tests/redis/accounts_api.rs
```diff
@@ -408,7 +407,7 @@ fn accounts_test() {
                     //assert_eq!(account.get("balance").expect("balance was expected"), &Value::Number(Number::from(0)));
                     assert_eq!(
                         balance.get("balance").expect("balance was expected"),
-                        &Value::Number(Number::from(0))
+                        &Value::Number(Number::from(-100))
```
emschwartz: Why is this being changed?

crates/interledger-service-util/src/expiry_shortener_service.rs
```diff
@@ -39,6 +39,11 @@ impl<O> ExpiryShortenerService<O> {
     }
 }
 
+// TODO: the bounds in the module give really bad error messages;
```
emschwartz: This is a simple fix, can you just copy the bounds from here and put them above?

crates/ilp-node/src/node.rs
bstrie: The box is necessary here for a tricky reason. With an `impl Future` return type, when returning different futures from different branches we'd need to wrap them in something like `Either` in order to get the types to line up. But in this case the actual branches that exist in the trailing `match` are determined wholly by which features are enabled, so we need a graceful way to support returning either one or several branches. Thus I used the `Box` here to erase the return types of each branch so that the types line up gracefully.

If we really didn't want to use the box here for whatever reason, I can imagine a functioning alternative that looks something like:
```rust
#[cfg(not(feature = "redis"), not(feature = "sqlite")]
fn serve_node [...]

#[cfg(feature = "redis"), not(feature = "sqlite")]
fn serve_node [...]

#[cfg(not(feature = "redis"), feature = "sqlite")]
fn serve_node [...]

#cfg(feature = "redis, feature = "sqlite")]
fn serve_node [...]
```
Though obviously this scales poorly as the number of supported databases increases.

crates/ilp-node/src/redis_store.rs
bstrie: Yes, in this case the only reason this is in this module is to have as few cfgs as possible, given that node.rs already contains several cfg blocks (and will continue to gain more) and so IMO eliminating the need for additional cfg blocks will help node.rs remain easier to read and reason about.

crates/ilp-node/src/redis_store.rs
bstrie: This isn't used at all in the tests in this repo, but the settlement-engines repo does call this function from its tests. Interestingly, in those tests is the comment `// TODO insert the accounts via HTTP request` ( https://github.com/interledger-rs/settlement-engines/blob/d8995867751ee62b13bbc4a785588a1a30bff62a/crates/ethereum-engine/tests/eth_ledger_settlement.rs#L85 ), which implies that this function could be removed entirely. @gakonst , how difficult would that be? Does this code simply predate the ability to insert accounts via the HTTP API?

crates/ilp-node/tests/redis/accounts_api.rs
bstrie: An intersection of various other concerns:

1) The desire to specifically denote which tests are specific to redis (as it turns out, right now, all of them) and in the future which tests are specific to every other supported database in a way that doesn't turn the tests directory into a mess. To me this suggested having subdirectories for each database.

2) However, `cargo test` doesn't automatically recurse into subdirectories of `tests/` in order to look for tests there. In order to run a test that lives in a subdirectory, we need to add a four-line entry to Cargo.toml for each test. In order to keep Cargo.toml from having N*M such entries, I reasoned that we should have a central `foo_tests.rs` file for each supported database that imports all the other test files. (In addition, at RustFest we spoke about how we may be able to reduce test build times by consolidating tests into a single binary, which reduces time spent linking, though I haven't actually benchmarked whether this noticeably impacts our build times).

3) However, now that all the tests are contained within a single binary rather than in multiple binaries, trying to run the tests in parallel (as `cargo test` does by default) causes inevitable panics as each test tries to reserve the same ports. We could either require everyone to manually do `cargo test -- --test-threads=1`, but it seems like a better solution would be to just use different ports.

4) Instead of manually going through and assigning unique ports to every test, I just used the built in support for getting a random open port. This does mean that effectively there's no reason for `get_open_port` to accept an Option, unless we really want some tests to have specific ports for whatever reason.

crates/ilp-node/tests/redis/accounts_api.rs
bstrie: Heh, so first read the giant block of reasoning at https://github.com/interledger-rs/interledger-rs/pull/551#discussion_r353885241 , then:

5) Now that we're no longer using fixed port numbers, we can't have the settlement engine URL be a `const`, so we just make it a String formatted with the appropriate dynamically-determined port.

6) However, this now means that the closures that we construct to do the actual API tests have to close over the settlement engine URL string. The string is an owned value, so the closures become FnOnce (and we can't use a reference here, because async shenanigans). And in the actual test runner at the very end of the file, exactly one of these closures gets called twice, which is illegal for an FnOnce. So we reorder the API calls to make it so that it's no longer necessary to call that closure twice.

7) However, now one of the two accounts is now no longer being deleted and recreated from scratch, so when we check the balance in a later test we expect its balance to reflect a transaction that happened earlier in the test process. And at this point I was so tired of this rabbit hole that I just changed the test to expect the lower balance rather than try to restore the exact same behavior.

bstrie: And the coup de grace of this particular tragedy/comedy is that it appears that Georgios literally deleted this whole file 3 hours ago. 🤦‍♂

crates/ilp-node/src/node.rs
emschwartz: Got it. If the method is public, I think it would be slightly better to return an `impl Future` but make that just a `Box<dyn Future...>` internally for now. If it's not public, then it doesn't really matter

crates/ilp-node/src/redis_store.rs
emschwartz: I figured, can you just put a comment in here explaining that?

crates/ilp-node/src/redis_store.rs
emschwartz: > Does this code simply predate the ability to insert accounts via the HTTP API?

Yes. The tests should be changed to use the CLI or the HTTP API directly

bstrie: Since updating settlement-engines seems out of scope at the moment, for now I've just marked this function with the deprecated attribute so that it's less likely to be forgotten about.

bstrie:
Changes addressed and pushed. Note that this PR contains a breaking change to the command-line invocation of ilp-node, as the "redis_url" option has been renamed to "database_url". @dora-gt , do you know whether this might cause any CI probkems? Specifically, I imagine there could be a problem if a new script (using "database_url") somehow gets used to call an old binary (using "redis_url"). We could theoretically address this quite easily (search for "XXX" in the PR diff to see what would need to change in order to temporarily support old binaries with new scripts), but I'm not sure whether anybody else thinks this could be a problem.

bstrie:
I think the CI failures seen here may be what I was just suggesting in the prior comment. I'll push a version with the aliases and we'll see how CI likes that. @emschwartz , you mentioned that if this happened then you'd want to consider changes to how the CI works?

crates/ilp-node/src/redis_store.rs
gakonst: @bstrie Thanks, this is code which existed before we had integrated adding accounts only via the HTTP API. You can safely remove it, and I will add a PR to `settlement-engines` which converts the E2E tests to not require it.

crates/ilp-node/tests/redis/accounts_api.rs
gakonst: Bulletpoint 3. seems like a strong enough argument to me for this

crates/ilp-node/tests/redis/accounts_api.rs
gakonst: Ugh yeah, this file was not really needed in the first place and was replaced by dedicated unit tests

bstrie:
Rebased.

crates/ilp-node/src/redis_store.rs
```diff
@@ -0,0 +1,87 @@
...
+                .insert_account(account)
+                .map_err(|_| error!(target: "interledger-node", "Unable to create account"))
+                .and_then(|account| {
+                    debug!(target: "interledger-node", "Created account: {}", account.id());
+                    Ok(account.id())
+                })
+        })
+}
+
+/*fn deserialize_redis_connection<'de, D>(deserializer: D) -> Result<ConnectionInfo, D::Error>
```
gakonst: Can this commented piece of code be removed?

crates/ilp-node/src/node.rs
```diff
@@ -251,22 +250,65 @@ impl InterledgerNode {
...
-        let settlement_api_bind_address = self.settlement_api_bind_address;
+    fn serve_node(self) -> Box<dyn Future<Item = (), Error = ()> + Send + 'static> {
         let ilp_address = if let Some(address) = &self.ilp_address {
             address.clone()
         } else {
             DEFAULT_ILP_ADDRESS.clone()
         };
+
+        // TODO: store a Url directly in InterledgerNode rather than a String?
+        let database_url = Url::parse(&self.database_url).unwrap();
```
gakonst: Can it ever be expected that the database endpoint cannot be parsed as `Url`?

crates/ilp-node/src/node.rs
```diff
@@ -275,194 +317,184 @@ impl InterledgerNode {
...
+                    } else {
+                        debug!(target: "interledger-node", "Not using exchange rate provider. Rates must be set via the HTTP API");
+                    }
+
+                    Ok(())
+                },
+            )
         })
-        .in_current_span())
+.in_current_span()
```
gakonst: is the indentation here correct?

Issue #521, opened 2019-11-19
Should /ilp be moved to /accounts/:username/ilp ?


emschwartz:
Right now we use the authorization to determine which account the packets are being sent for. We could continue doing that or it might make more sense to put that path under accounts, since the packets are being sent on behalf of a specific account.

This could be a breaking change but we could do it in a backwards-compatible way by supporting both in the meantime and deprecating the `/ilp` (and `/ilp/btp`).

Thoughts?

cc @gakonst @dora-gt

matdehaast:
We initially had been using a singular `/ilp` but ended up changing towards an equivalent of `/accounts/:username/ilp`. The latter provides much more flexibility, especially when you are doing both authz and authn in two different steps.

dora-gt:
I think we should adopt this approach as mentioned here https://github.com/interledger-rs/interledger-rs/issues/500#issuecomment-555370312 because now our bearer tokens don't seem to conform to normative ones.

gakonst:
Fixed in https://github.com/interledger-rs/interledger-rs/pull/553

PR #552, opened 2019-12-04
Accounts and Node Settings API Tests Refactor

gakonst:
Previously, the tests for the accounts and node settings api were implemented as E2E tests. This was
1. unnecessarily complex
2. resulted in VERY slow feedback loops when trying to test the API's functionalities

This PR creates `test_helpers` under `interledger-api`, and proceeds to unit test each function of the API in order to ensure that permission control is applied properly. The E2E tests which duplicated the functionality are removed.

This is done in preparation for #521, since we will be moving some things around, so that we can be sure we have solid coverage of the endpoints.


crates/interledger-api/src/routes/accounts.rs
```diff
@@ -618,3 +618,138 @@ where
...
+    use crate::routes::test_helpers::*;
+
+    #[test]
+    fn only_admin_can_create_account() {
+        let api = test_accounts_api();
+        let resp = api_call(&api, "POST", "/accounts", "admin", DETAILS.clone());
+        assert_eq!(resp.status().as_u16(), 200);
+
+        let resp = api_call(&api, "POST", "/accounts", "wrong", DETAILS.clone());
+        assert_eq!(resp.status().as_u16(), 500);
```
emschwartz: This should be a 401 not a 500

crates/interledger-api/src/routes/mod.rs
```diff
@@ -3,3 +3,375 @@ mod node_settings;
 
 pub use accounts::accounts_api;
 pub use node_settings::node_settings_api;
+
+#[cfg(test)]
+pub mod test_helpers {
```
emschwartz: Any reason not to put this in a separate file?

crates/interledger-api/src/routes/accounts.rs
gakonst: I have a follow up PR coming later so I'll fix it there.

PR #553, opened 2019-12-04
Make HTTP and BTP outgoing tokens symmetric to incoming tokens

gakonst:
1. Mounts `/ilp` to `/accounts/:username/ilp`
1. Mounts `/ilp/btp` to `/accounts/:username/ilp/btp`
2. This allows moving the `username` from the `outgoing_token` to being specified at the URL when creating an account in BTP and HTTP
3. Removes the need for `AuthToken` in HTTP & BTP tokens.


crates/interledger-http/src/server.rs
```diff
@@ -30,65 +32,247 @@ where
...
+                parse_auth_clone(&path_username, &password)
             })
             .and(warp::body::content_length_limit(MAX_PACKET_SIZE))
             .and(warp::body::concat())
+            .and_then(move |account: S::Account, body: warp::body::FullBody| {
+                respond_clone(account, body)
+            })
+            .boxed();
+
+        let old_api = warp::post2()
```
emschwartz: I don't think it's worth doing this in a backwards-compatible way. We'll just make the next release a breaking one

crates/interledger-http/src/client.rs
```diff
@@ -68,22 +64,10 @@ where
...
-                        data: &[],
-                    }
-                    .build()))
-                }
-            };
             Box::new(
                 self.client
                     .post(url.as_ref())
-                    .header("authorization", auth.to_bearer())
+                    .header("authorization", &format!("Bearer {}", token))
```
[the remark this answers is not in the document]
gakonst: I just switched the return type of `get_http_auth_token` to be `SecretString`

.circleci/config.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       - run:
           name: Run run-md Test
           command: |
-            scripts/run-md-test.sh
+            scripts/run-md-test.sh '^.*$' 1
```
emschwartz: What does that do?

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -167,15 +170,12 @@ where
         .and(warp::header::<String>("authorization"))
```
emschwartz:
```suggestion
        .and(warp::header::<SecretString>("authorization"))
```

crates/interledger-api/src/routes/accounts.rs
```diff
@@ -458,7 +458,6 @@ where
         .or(put_account_settings)
         .or(incoming_payment_notifications)
         .or(post_payments)
-        .boxed()
```
emschwartz: Why not box this?

crates/interledger-api/src/routes/node_settings.rs
```diff
@@ -33,7 +33,7 @@ pub fn node_settings_api<S, A>(
     admin_api_token: String,
     node_version: Option<String>,
     store: S,
-) -> warp::filters::BoxedFilter<(impl warp::Reply,)>
+) -> impl warp::Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone
```
emschwartz: This can still return an `impl Filter` even if the underlying code returns a `Box<Filter>`

crates/interledger-http/src/client.rs
```diff
@@ -67,23 +64,17 @@ where
...
+                .to
+                .get_http_auth_token()
+                .unwrap_or_else(|| SecretString::new("".to_owned()));
             Box::new(
                 self.client
                     .post(url.as_ref())
-                    .header("authorization", auth.to_bearer())
+                    .header(
+                        "authorization",
+                        &format!("Bearer {}", token.expose_secret()),
```
emschwartz: This creates a new `String` that may not be zeroed in the way that `secrecy` does it. That's why I thought we might want to make the account method return a `SecretString` that contains the full authorization header value (including the `Bearer ` part)

crates/interledger-http/src/lib.rs
```diff
@@ -22,7 +23,7 @@ pub use self::server::HttpServer;
 
 pub trait HttpAccount: Account {
     fn get_http_url(&self) -> Option<&Url>;
-    fn get_http_auth_token(&self) -> Option<&str>;
+    fn get_http_auth_token(&self) -> Option<SecretString>;
```
emschwartz: Could this be `Option<&SecretString>` so it doesn't need to copy it each time?

crates/interledger-http/src/server.rs
```diff
@@ -38,19 +40,23 @@ where
...
-                store
-                    .get_account_from_http_auth(auth.username(), auth.password())
-                    .map_err(move |_| -> Rejection {
-                        error!(
-                            "Invalid authorization provided for user: {}",
-                            auth.username()
-                        );
-                        ApiError::unauthorized().into()
-                    })
+            .and(warp::header::<String>("authorization"))
```
emschwartz:
```suggestion
            .and(warp::header::<SecretString>("authorization"))
```

crates/interledger-store/src/account.rs
```diff
@@ -255,10 +254,14 @@ impl HttpAccount for Account {
         self.ilp_over_http_url.as_ref()
     }
 
-    fn get_http_auth_token(&self) -> Option<&str> {
-        self.ilp_over_http_outgoing_token
-            .as_ref()
-            .map(|s| str::from_utf8(s.expose_secret().as_ref()).unwrap_or_default())
+    fn get_http_auth_token(&self) -> Option<SecretString> {
+        self.ilp_over_http_outgoing_token.as_ref().map(|s| {
```
emschwartz: Seems like it should be fine to just return `&self.ilp_over_http_outgoing_token`

.circleci/config.yml
gakonst: It runs the Markdown tests in source mode. Otherwise, the tests try to download the binary from the `latest` release, and they would fail.

crates/interledger-api/src/routes/accounts.rs
gakonst: Not possible currently as we're missing a `FromStr` implementation for `SecretString`, opened a PR: https://github.com/iqlusioninc/crates/pull/309

crates/interledger-api/src/routes/accounts.rs
gakonst: Compiling interledger-api with boxed takes 2.72 sec, while without takes 3.39, so I'll re-enable it.

crates/interledger-http/src/server.rs
gakonst: Tracked in #556

crates/interledger-store/src/account.rs
gakonst: Tokens are stored as `SecretBytes` because we use the same variable for both unencrypted and encrypted tokens (which have to be bytes, and cannot be strings).

Hence, this conversion is necessary

crates/interledger-http/src/lib.rs
gakonst: This function creates a string which it owns with str::from_utf8 and `to_string()` subsequently so we cannot really make it return a reference, because it'd be returning a referecne to data it already owns.

https://github.com/interledger-rs/interledger-rs/blob/863c99112d6c6eaf646e3ade540c31154c0be788/crates/interledger-store/src/account.rs#L260

We cannot convert the `SecretBytes` to a `SecretString`, because tokens are stored encrypted as `SecretBytes` and we're storing both encrypted and cleartext values in the same variable.

PR #555, opened 2019-12-05
Make BTP outgoing tokens symmetric to incoming tokens

gakonst:
Based on #553 

BTP tokens now follow the same logic.

I did not like the way we were creating service and filter in 1 function, so I broke them in 2 different functionalities. The BTP token is now the only data required to be passed via the websocket, as the username is fetched from the path.


crates/interledger-btp/src/server.rs
```diff
@@ -167,11 +164,12 @@ where
 
 struct Auth {
     request_id: u32,
-    token: AuthToken,
+    token: String,
```
emschwartz:
```suggestion
    token: SecretString,
```

PR #559, opened 2019-12-06
docs: Add OpenAPIv3 specification and updated README/api docs

gakonst:
As title.

Swagger API published here: https://app.swaggerhub.com/apis/interledger-rs/Interledger/1.0

We should soon make a release and make Xpring update to that, since that's what the docs are based on.


docs/api.md
```diff
@@ -4,172 +4,24 @@ For instructions on running the ILP Node, see the [Readme](../README.md).
...
-Your HTTP requests would look like:
+Your HTTP requests should look like:
 
 ```
 GET /accounts HTTP/1.1
 Authorization: Bearer BEARER-TOKEN-HERE
 ```
 
-The format of the `Authorization` header is either of the following.
+For administrative functionalities, the value of the token must be the value of `admin_auth_token` when the node was launched. When authorizing as a user, it must be the `ilp_over_http_incoming_token` which was specified during that user's account creation.
```
bstrie: A nit, but if we're about to do a stable release, is `ilp_over_http_incoming_token` still the best name for it? Isn't it just a general user auth token now?

docs/api.yml
```diff
@@ -0,0 +1,702 @@
...
+          description: The node's information
+          content:
+            application/json:
+                schema:
+                  $ref: '#/components/schemas/NodeInformation'
+
+  # Default SPSP Account
+  /.well_known/pay:
+    get:
+      summary: The default SPSP account used on the node. This endpoint is only enabled if the node is run with the configuration option ILP_DEFAULT_SPSP_ACCOUNT.
```
bstrie: If this endpoint is a standardized feature of SPSP, should the summary also link to the appropriate section in the spec?

docs/api.yml
```diff
@@ -0,0 +1,702 @@
...
+          description: Bearer token with the account's or administrator's authorization
+      responses:
+        '200':
+          description: The requested account's information
+          content:
+            application/json:
+                schema:
+                  $ref: '#/components/schemas/Account'
+    put:
+      summary: Edit an account's information
```
bstrie: I'd like the summary to briefly note the differences between this endpoint and the settings endpoint.

docs/api.yml
```diff
@@ -0,0 +1,702 @@
...
+  /accounts/{username}/settings:
+    parameters:
+      - in: path
+        name: username
+        schema:
+          type: string
+        required: true
+        description: Username of the account whose information you are operating on
+    put:
+      summary: Edit an account's settings (e.g. tokens, URLs, settlement thresholds)
```
bstrie: Likewise this should mention the difference with the PUT /accounts/username endpoint.

docs/api.yml
```diff
@@ -0,0 +1,702 @@
...
+      - in: path
+        name: username
+        schema:
+          type: string
+        required: true
+        description: Username of the account whose information you are operating on
+    post:
+      summary: Send payment to an account
+      tags:
+        - users
```
bstrie: Should the summary of this endpoint note the caveat that, despite this endpoint being unavailable to admins, if the admin has actual access to the server itself then they can still do whatever they want, including sending payments from any account?

docs/api.yml
```diff
@@ -0,0 +1,702 @@
...
+          content:
+            text/plain:
+              schema:
+                type: string
+                example: "alice"
+
+  # Rates endpoints
+  /rates:
+    get:
+      summary: Get all of the node's exchange rates (currently open, but might be converted to admin/user-only in the future)
```
bstrie: This is something that needs to be decided before a stable release. Is there an issue filed for it?

docs/api.md
gakonst: We need 2 tokens per account, the incoming is the one which is used by the account's owner to make requests regarding that account (e.g. get/edit its info or send payments), while the outgoing is the one which contains the credentials of node operator's account on the account owner's node (as a node operator, there is no point in specifying an outgoing token for an account, if you do not have an account on the account owner's node)

docs/api.yml
gakonst: @emschwartz @dora-gt I think it's fine to leave it always open

docs/peering.md
```diff
@@ -0,0 +1,93 @@
...
+Depending on the type of relation used, the node will choose to broadcast and/or receive routes from that account over the [Connector Communication Protocol](../crates/interledger-ccp/README.md).
+
+These are:
+1. NonRoutingAccount: The node will not attempt to broadcast routes to that account, and will reject any broadcasted routes from it.
+1. Peer: The node will both try to broadcast and receive routes to/from that account.
+1. Child: The node will try to broadcast routes to that account, but will reject any broadcasted routes from it.
+1. Parent: The node will not try to broadcast routes from that account, but will accept any broadcasted routes from it.
+
+Assigned addresses also may depend on the network relation between nodes.
+Parent nodes are expected to be in higher Interledger Address namespaces (e.g. if Alice is the parent of Bob, then it can be expected that Bob's address will be `example.alice.bob`). If no ILP Address is specified during account creation, then the account's username is appended to the node's ILP Address at lower hierarchical level.
```
bstrie: > If no ILP Address is specified during account creation, then the account's username is appended to the node's ILP Address at lower hierarchical level.

I would re-word this as "If no ILP Address is specified during account creation, then the account's ILP Address will be automatically generated by appending the account's username to the ILP Address of its parent account."

docs/peering.md
```diff
@@ -0,0 +1,93 @@
...
+
+## URLs, Incoming and Outgoing Tokens
+
+1. It is assumed that the node operator knows the format of ILP-over-HTTP/BTP URLs the peer is using. It is expected that there is an out of band communication channel via whcih the peer communicates such information.
```
bstrie: Nit: "whcih" -> "which"

docs/peering.md
```diff
@@ -0,0 +1,93 @@
...
+1. The incoming token on a node MUST be the same as the outgoing token on the peer. Note that the incoming token is used either for HTTP authentication messages (HTTP-token only) such as sending an SPSP payment or editing account information, or a packet coming from a peer via ILP-over-HTTP/BTP.
```
bstrie: > HTTP-token

I'd put "HTTP Bearer token" for maximum clarity.

docs/peering.md
```diff
@@ -0,0 +1,93 @@
...
+1. If there is no expectation of Bob sending a packet to Alice via ILP-over-HTTP, the incoming token may not be specified for the peering process.
+
+## Actions which happen simultaneously to peering:
+
+When adding an account, the following happens before the account is inserted to the node's store:
+
+1. If a BTP URI is present, the node tries to establish a BTP websocket connection. The BTP outgoing token must also be set, so that the node can authorize against the peer (when the node shuts down these connections are also closed and reinitiated on launch)
+1. If they are a parent:
+    1. The node performs an ILDCP request to the parent's node to get the address assigned to it by the parent (this is expected to be a lower-level address, e.g. if the parent is `g.alice`, the ILDCP Response will assign the child address as `g.alice.bob`).
```
bstrie: > 1. If they are a parent:
>    1. The node performs an ILDCP request to the parent's node to get the address assigned to it by the parent

To me this reads like "if the new account is a parent, the node performs a request to the parent's node", which is confusing. Was "If they are a parent" supposed to be "If they are a child"?

docs/api.md
```diff
@@ -4,172 +4,24 @@ For instructions on running the ILP Node, see the [Readme](../README.md).
...
 
-1. The admin token
-    - `Bearer admin-token`
-    - `admin-token` is the token you specified as `admin_auth_token` when you spun up the node.
-1. User tokens
-    - `Bearer password`
-    - `password` is the token you specified as `ilp_over_http_incoming_token` when you created the account.
+## HTTP REST API
 
+### **By default, the API is available on port `7770` and it exposes endpoints as specified in [this OpenAPIv3 specification](https://app.swaggerhub.com/apis/interledger-rs/Interledger/1.0)  ([corresponding yml file](./api.yml)).**
```
dora-gt: This looks a bit weird like:
image
So I would suggest:


```suggestion
**By default, the API is available on port `7770` and it exposes endpoints as specified in [this OpenAPIv3 specification](https://app.swaggerhub.com/apis/interledger-rs/Interledger/1.0)  ([corresponding yml file](./api.yml)).**
```

docs/api.md
dora-gt: In my opinion, something that could be considered as in apposition should be listed because readers will be able to understand the items are in apposition and how many items are there at a glance. That's why I wrote so.

docs/peering.md
```diff
@@ -0,0 +1,93 @@
+# Connecting 2 Interledger Nodes
+
+In order for 2 nodes to connect, they must have added each other as an account with the appropriate settings set. In this document, we go over the exact settings and elaborate on the ways node operators can add accounts so that they route packets to and from them.
+
+## Minimum Viable Peering
+
+Assume we have Alice (node at `alice.com`) and Bob (`bob.com`), and they want to peer their nodes together. Without loss of generality, we will assume Alice initiates the peering process, and that they will use `alice` and `bob` as usernames respectively.
+
+1. Alice issues an API token which she will use to authenticate on Bob's node.
```
dora-gt: Are you using `authenticate` intentionally?
I think this is `authorize`. As my recognition, Authenticate = Identify, Authorize = Allow.

docs/api.md
gakonst: I think it's easy to understand here that it's either `Authentication: Bearer <token>`, where `token` is either the admin token or the `ilp_over_http_incoming_token`, without having to be overly overbose?

docs/api.md
```diff
@@ -4,172 +4,24 @@ For instructions on running the ILP Node, see the [Readme](../README.md).
 
 ## Authorization
 
-The ILP Node uses HTTP Bearer Token authorization. Most requests must either be authorized with the admin token configured on the node or the token configured for a particular account.
```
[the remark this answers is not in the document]
gakonst: The OpenAPI specification makes this explicit now

Issue #485, opened 2019-10-31
Config doesn't parse snake case ExchangeRateProvider


emschwartz:
The `ExchangeRateProvider` should be able to parse both values like `CoinCap` and `coin_cap` (see the serde aliases defined [here](https://github.com/interledger-rs/interledger-rs/blob/e79a12cdbc174e8e2c09efc39552cc68fb47da7b/crates/interledger-service-util/src/exchange_rates_service.rs#L209-L213)). I thought this would work because as far as I can tell the `config-rs` library uses serde to parse the config values. However, passing the value `coin_cap` either via YAML or as an environment variable currently results in the error: `enum ExchangeRateProvider does not have variant constructor coin_cap`.

I wonder if this could be related to the magic we've got for allowing arguments to be passed in as either command line args, env vars, or via config file.

dora-gt:
I just create a simple test for this:
https://github.com/interledger-rs/interledger-rs/blob/fix-%23485/crates/ilp-node/tests/cli.rs

It fails because of
```
2019-11-22T11:29:08.408185+00:00 DEBUG interledger-cli-test: config error: enum ExchangeRateProvider does not have variant constructor crypto_compare
```

So I think this is related to something Config-ish problem.

Maybe this line?
https://github.com/mehcode/config-rs/blob/e84979415a2b3a68734bc26c3228aa326dc33e3c/src/de.rs#L373

gakonst:
Cloudflare's wrangler moved away from config here: https://github.com/cloudflare/wrangler/issues/867

Seems like config devs do not want multiple types of env vars / config options: https://github.com/mehcode/config-rs/issues/111#issuecomment-544970441

In this specific case, I'd be more in favor of just removing all the aliases and just using all lowercase. Don't see why anybody would write `crypto_compare` instead of `cryptocompare`, or why they'd want to add a capital C

emschwartz:
FWIW using snake_case would be the most consistent with the rest of the config but picking one form instead of having multiple aliases sounds reasonable to me

PR #563, opened 2019-12-14
fix(exchange-rates-service): remove service provider aliases

gakonst:
Closes https://github.com/interledger-rs/interledger-rs/issues/485

IMO it makes more sense to keep things simple and allow only 1 intuitive format, rather than try to support multiple


crates/ilp-node/tests/redis/exchange_rates.rs
```diff
@@ -112,7 +112,7 @@ fn cryptocompare() {
         "exchange_rate": {
             "poll_interval": 60000,
             "provider": {
-                "crypto_compare": api_key
+                "cryptocompare": api_key
```
emschwartz: @austin-king Note that when the testnet node is updated, the config will need to be changed to be just lowercase

PR #567, opened 2019-12-17
ci(config.yml): set timeout to 30m

dora-gt:
It seems that building docker images requires over 15min so this PR increases it to 30min.
I'm not sure if this is sufficient but I'd like to see whether it goes well with this.


gakonst:
Merged, let's see if this fixes it.

PR #577, opened 2019-12-18
feat: SQLite database backend, foundation

bstrie:
This commit adds the ability to run a server backed by an SqliteStore as
opposed to a RedisStore, although currently all the necessary Store
trait methods are doing the least amount of work necessary and all the
_un_necessary Store trait methods are simply stubbed out with
`unimplemented!()`. The result is that the running node is able to
successfully respond to the status endpoint, "GET /".

---

The SQLite store is going to involve a lot of code, so as I get it all ready I'm splitting it out into individual PRs to make it tractable to review. This first PR involves a few small additional changes made to ilp-node in order to better accommodate the new database. It also introduces the `interledger-store::sqlite` module, in which will live all the code in subsequent PRs.

The contents of this PR are largely straightforward, but there are some important things to discuss. In particular we need to confront two things:

1. Our driver's query API is synchronous, not async. The driver we're using is the most popular and well-developed SQLite driver in Rust, and in any case there aren't any SQLite drivers that I can find with async support. This problem is hopefully mitigated by the fact that, for the time being, the database itself is stored in-memory within our process rather than on-disk. Therefore reading and writing involves no file I/O, although synchronization is still capable of blocking the thread if multiple connections try to access the same tables. Which brings us to...
2. Unlike Redis, SQLite (the database software itself, not just our driver) doesn't allow the same connection to be shared among multiple threads. To quote [the docs](https://www.sqlite.org/threadsafe.html), "In [the mode we want to be using], SQLite can be safely used by multiple threads provided that no single database connection is used simultaneously in two or more threads." In Rust terms, this means that an SQLite connection is `Send` but not `Sync`. This is the sort of scenario that a connection pool would solve, and indeed for our driver (rusqlite) there exists an adapter for the popular r2d2 connection pool library. However, r2d2 is itself synchronous, not async; there exist async connection pools in Rust (e.g. bb8, deadpool), but as of right now neither of those have rusqlite adapters.

For expedience, for the time being my approaches to these problems are:

1. Ignore the synchronous aspect of the query API and hope that queries are fast enough and that the number of simultaneous tasks is low. Depending on the roles that we see each supported database backend playing, it may suffice to tell users in high-contention scenarios to use the redis backend for the time being. The alternative to this approach would be to add async support to rusqlite ourselves.
2. Open a new connection in each context that a connection is called for. This may not be as bad as it sounds, since (as noted previously), we're currently using an in-memory database rather than an on-disk database, which means that opening a new connection involves no file I/O. Furthermore, by leveraging SQLite's support for sharing caches among connections, we reduce the amount of work that is required each time a new connection is created. The alternative to this approach would be to add rusqlite support to an async connection pool, or else to fork rusqlite and replace the internal `Rc` in each `Connection` with a `Sync` type like `RwLock` (and I'm far from 100% sure that this latter approach is safe, depending on the invariants that SQLite demands).

(And in case you're wondering about the ramifications of using an in-memory database, it should be possible to both write the in-memory database to disk and to load an in-memory database from disk, although I have yet to try doing so.)

Progress towards #474.

### DIFF
--- a/crates/interledger-store/src/sqlite/mod.rs
+++ b/crates/interledger-store/src/sqlite/mod.rs
@@ -1,0 +1,506 @@
+#![allow(unused_imports)]
+#![allow(unused_variables)]
+//! Data store for Interledger.rs using SQLite
+
+use super::account::{Account, AccountWithEncryptedTokens};
+use super::crypto::{encrypt_token, generate_keys, DecryptionKey, EncryptionKey};
+use bytes::Bytes;
+use futures::{
+    future::{err, ok, result, Either},
+    sync::mpsc::UnboundedSender,
+    Future, Stream,
+};
+use http::StatusCode;
+use interledger_api::{AccountDetails, AccountSettings, EncryptedAccountSettings, NodeStore};
+use interledger_btp::BtpStore;
+use interledger_ccp::{CcpRoutingAccount, RouteManagerStore, RoutingRelation};
+use interledger_http::HttpStore;
+use interledger_packet::Address;
+use interledger_router::RouterStore;
+use interledger_service::{Account as AccountTrait, AccountStore, AddressStore, Username};
+use interledger_service_util::{
+    BalanceStore, ExchangeRateStore, RateLimitError, RateLimitStore, DEFAULT_ROUND_TRIP_TIME,
+};
+use interledger_settlement::core::{
+    idempotency::{IdempotentData, IdempotentStore},
+    scale_with_precision_loss,
+    types::{Convert, ConvertDetails, LeftoversStore, SettlementStore},
+};
+use interledger_stream::{PaymentNotification, StreamNotificationsStore};
+use lazy_static::lazy_static;
+use log::{debug, error, trace, warn};
+use num_bigint::BigUint;
+use parking_lot::RwLock;
+use rusqlite::Connection;
+use secrecy::{ExposeSecret, Secret, SecretBytes};
+use serde::{Deserialize, Serialize};
+use serde_json;
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Display,
+};
+use std::{
+    iter::{self, FromIterator},
+    str,
+    str::FromStr,
+    sync::Arc,
+    time::{Duration, Instant},
+};
+use tokio_executor::spawn;
+use tokio_timer::Interval;
+use url::Url;
+use uuid::Uuid;
+use zeroize::Zeroize;
+
+/*const ACCOUNT_DETAILS_FIELDS: usize = 21;
+
+static PARENT_ILP_KEY: &str = "parent_node_account_address";
+static ROUTES_KEY: &str = "routes:current";
+static STATIC_ROUTES_KEY: &str = "routes:static";
+static DEFAULT_ROUTE_KEY: &str = "routes:default";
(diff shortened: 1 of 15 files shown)


bstrie:
As I feared, trying to compile with *both* the redis backend and sqlite backend has caused CI to time out. :\

crates/interledger-store/src/sqlite/mod.rs
```diff
@@ -0,0 +1,517 @@
...
+}*/
+
+lazy_static! {
+    static ref DEFAULT_ILP_ADDRESS: Address = Address::from_str("local.host").unwrap();
+}
+
+pub struct SqliteStoreBuilder {
+    sqlite_url: Url,
+    secret: [u8; 32],
+    poll_interval: u64,
```
emschwartz: Polling may not be necessary if SQLite can only be use in a single-node deployment (all the values would be altered via the API)

emschwartz:
> Open a new connection in each context that a connection is called for

Would this mean each connection would effectively be a different database?

bstrie:
> Would this mean each connection would effectively be a different database?

Nope. Normally opening multiple in-memory databases in SQLite will cause a different database to be created for each connection, but when opening shared-cache connections to an explicitly-declared (rather than implicitly-declared) in-memory database the two connections will be to the same database. (AIUI, when talking about in-memory databases, the "cache" simply *is* the database.)

bstrie:
Comments addressed. I've also added a commit that allows CI to ignore the sqlite feature since it's neither usable nor has any tests yet, so that we don't need to bump up the CI timeout.

bstrie:
> This seems fine so far, though it doesn't really seem fleshed out enough to merge into master

The only point at which this will be fleshed out is when it's all done, which is going to involve the addition of thousands of lines, so I'm not sure what the alternative is that doesn't involve dumping a giant unreviewable PR.

bstrie:
As an alternative to merging these commits to master, I'll start issuing new PRs against this branch. Fortunately there's almost no risk of merge conflicts, since the primary file is brand new. But we will eventually need to be careful to merge them in-order to keep from needing to rebase repeatedly.

crates/interledger-store/src/sqlite/mod.rs
```diff
@@ -0,0 +1,517 @@
...
+        unimplemented!()
+    }
+}
+
+impl RateLimitStore for SqliteStore {
+    type Account = Account;
+
+    /// Apply rate limits for number of packets per minute and amount of money per minute
+    ///
+    /// This uses https://github.com/brandur/redis-cell so the redis-cell module MUST be loaded into redis before this is run
```
gakonst: Outdated comment - is there a way to rate limit sqlite?

gakonst:
So far so good, given that the db is loaded in memory, should we go with the 2nd approach of opening multiple connections?

gakonst:
@bstrie ping on this ^

dora-gt:
Is there any other SQL database that supports async?

If the goal of introducing SQL to interledger-rs is to make the store have strong persistence, we could use other SQL databases that support the async ecosystem. We might not need to persist in SQLite itself.

matdehaast:
Also as a note, Redis does support full persistence. You just have to configure it that way

gakonst:
IMO tokio-postgres is what we should use here https://docs.rs/tokio-postgres

bstrie:
This branch needs to be rebased anyway, so I'll close this for now.

PR #582, opened 2020-01-02
Migration to 0.3 & Async/Await

### LINKED ISSUE #85: Switch to async/await syntax once stabilized
https://boats.gitlab.io/blog/post/await-decision-ii/

> I’m working actively on drafting a stabilization report to propose stabilizing a minimum viable version of async/await in the 1.37 release of Rust. 1.37 will be released in mid-August, and branches on July 4.

Once that's out, we should get a PR that replaces all the futures handling with async/await calls.

gakonst:
As title. Still WIP. Opening for initial feedback. This will be split in smaller PRs and squashed/split into more verbose commits.

- We can get rid of the `BoxedIlpFuture` and replace it everywhere with `IlpResult` (which is really a `Result<Fulfill, Reject>`.
- We can utilize async-trait to allow our traits to have async functions. So far so good, haven't hit any edge cases yet. Had to manually add `'async_trait` as a lifetime.
- Note that the reduction in generics, as well as the eventual removal of the huge closure-chains will improve our compilation times significantly. 
- Bytes-05 is now needed for all HTTP related crates, so we temporarily maintain both versions in the required crates. Before the migration is over, Bytes 0.5 should be the only version in the repository. This could be more complex than expected due to breaking changes: 
   - https://github.com/tokio-rs/bytes/issues/350
    - https://github.com/tokio-rs/bytes/pull/288
   I am considering getting this PR to a merge-able state, and then open a separate issue to port the whole crate to Bytes 0.5.
- Warp needs to be added from master for now
- Tokio-tungstenite got futures 0.3 support recently so we can expect it to be released soon

Will fix #85.

- [x] service
- [x] ildcp
- [x] http
- [x] router
- [x] btp: blocked on https://github.com/snapview/tokio-tungstenite/pull/68
- [x] ccp
- [x] service-util: how do timeouts with 0.3 futures work?
- [x] api: blocked on some Warp changes (need to document)
- [x] settlement: requires compat methods for tokio-retry
- [x] store
- [x] stream
- [x] spsp
- [x] ilp-node: blocked on BTP and Prometheus
- [x] ilp-cli

### DIFF
--- a/crates/interledger-ildcp/src/client.rs
+++ b/crates/interledger-ildcp/src/client.rs
@@ -4,7 +4,7 @@ use interledger_service::*;
 use log::{debug, error};
 use std::convert::TryFrom;
 
-/// Get the ILP address and asset details for a given account.
+/// Sends an ILDCP Request and receives the ILP address and asset details for a given account.
 pub async fn get_ildcp_info<S, A>(service: &mut S, account: A) -> Result<IldcpResponse, ()>
 where
     S: IncomingService<A>,
--- a/crates/interledger-ildcp/src/client.rs
+++ b/crates/interledger-ildcp/src/client.rs
@@ -1,34 +1,30 @@
 use super::packet::*;
-use futures::Future;
+use futures::future::TryFutureExt;
 use interledger_service::*;
 use log::{debug, error};
 use std::convert::TryFrom;
 
 /// Get the ILP address and asset details for a given account.
-pub fn get_ildcp_info<S, A>(
-    service: &mut S,
-    account: A,
-) -> impl Future<Item = IldcpResponse, Error = ()>
+pub async fn get_ildcp_info<S, A>(service: &mut S, account: A) -> Result<IldcpResponse, ()>
 where
     S: IncomingService<A>,
     A: Account,
 {
     let prepare = IldcpRequest {}.to_prepare();
-    service
+    let fulfill = service
         .handle_request(IncomingRequest {
             from: account,
             prepare,
         })
         .map_err(|err| error!("Error getting ILDCP info: {:?}", err))
-        .and_then(|fulfill| {
-            let response =
-                IldcpResponse::try_from(fulfill.into_data().freeze()).map_err(|err| {
-                    error!(
-                        "Unable to parse ILDCP response from fulfill packet: {:?}",
-                        err
-                    );
-                })?;
-            debug!("Got ILDCP response: {:?}", response);
-            Ok(response)
-        })
+        .await?;
+
+    let response = IldcpResponse::try_from(fulfill.into_data().freeze()).map_err(|err| {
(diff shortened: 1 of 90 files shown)


crates/interledger-service/src/lib.rs
```diff
@@ -303,29 +305,36 @@ where
...
 pub trait AddressStore: Clone {
-    /// Saves the ILP Address in the store's memory and database
+    /// Saves the ILP Address in the database AND in the store's memory so that it can
+    /// be read without read overhead
     fn set_ilp_address(
         &self,
+        // The new ILP Address of the node
         ilp_address: Address,
-    ) -> Box<dyn Future<Item = (), Error = ()> + Send>;
+    ) -> Box<dyn Future<Output = Result<(), ()>> + Send>;
```
bstrie: Why not use async_trait on this trait rather than returning an explicitly boxed future?

crates/interledger-service/src/lib.rs
```diff
@@ -39,8 +40,8 @@ mod username;
 pub use username::Username;
 #[cfg(feature = "trace")]
 mod trace;
-#[cfg(feature = "trace")]
-pub use trace::*;
+
+pub type IlpResult = Result<Fulfill, Reject>;
```
bstrie: 👍

crates/interledger-ildcp/src/client.rs
```diff
@@ -1,34 +1,30 @@
...
-            Ok(response)
-        })
+        .await?;
+
+    let response = IldcpResponse::try_from(fulfill.into_data().freeze()).map_err(|err| {
+        error!(
+            "Unable to parse ILDCP response from fulfill packet: {:?}",
+            err
+        );
+    })?;
```
bstrie: As I mention elsewhere, I'd love to introduce proper error return types so that we would no longer have to deal with so much wasted space due to `map_err`s, like those in this function. Doesn't need to be part of this effort, but async/await makes our code so much cleaner that instances like these really really stand out.

crates/interledger-ildcp/src/server.rs
```diff
@@ -44,10 +43,72 @@ where
...
-            Box::new(ok(fulfill))
+            Ok(Fulfill::from(response))
         } else {
-            Box::new(self.next.handle_request(request))
+            self.next.handle_request(request).await
+        }
+    }
+}
+
+#[cfg(test)]
```
bstrie: Why add these tests as part of this PR?

crates/interledger-service/src/lib.rs
gakonst: Ended up doing it on a separate branch https://github.com/interledger-rs/interledger-rs/blob/gakonst/futures03-http/crates/interledger-service/src/lib.rs#L324

PR #592, opened 2020-01-23
docs: correct docker images in README

kincaidoneil:
Replaces `interledgerrs/node` with `interledgerrs/ilp-node`, since the former is listed as deprecated on Docker Hub.


gakonst:
I'll cherrypick this over to my open PR about docs which has the clippy issues fixed

PR #596, opened 2020-01-23
[Part 2/N] Interledger Service: Futures 0.3 Transition

gakonst:


crates/interledger-service/src/lib.rs
```diff
@@ -269,17 +277,16 @@ where
     }
 }
 
+#[async_trait]
 impl<F, IO, A, R> IncomingService<A> for WrappedService<F, IO, A>
 where
-    F: Fn(IncomingRequest<A>, IO) -> R,
-    IO: IncomingService<A> + Clone,
+    F: Fn(IncomingRequest<A>, IO) -> R + Send + Sync,
```
bstrie: Out of curiosity, are these extra `Send + Sync` bounds something that you explicitly wanted to add, or were you just adding them to pacify the compiler? Because if it's the latter, I suspect that would arise because of the bounds that `async_trait` inserts by default on the return types of async methods; you can tell `async_trait` *not* to insert `Send + Sync` by writing the attribute as `#[async_trait(?Send)]`.

gakonst: I thought of doing that, however in other places of the codebase we `tokio::spawn` futures, e.g. `send_route_update` in `interledger_ccp` which makes a subsequent [`send_request`](https://github.com/interledger-rs/interledger-rs/blob/c4cd6f3177b942e80a35d52668517083b796ffc8/crates/interledger-ccp/src/server.rs#L891). Same applies for [`handle_request` here](https://github.com/interledger-rs/interledger-rs/blob/d9cc785f73a98cd860b8d232eaa913a4f1e49728/crates/interledger-btp/src/service.rs#L274). Since these 2 calls are spawned, the future has to be Send/Sync, so we cannot remove that requirement unfortunately

PR #601, opened 2020-01-23
[Part 8/N] Interledger Stream: Futures 0.3 Transition 

gakonst:
Moving the discussion about the `pending_requests` removal here: https://github.com/interledger-rs/interledger-rs/pull/585#discussion_r368953339


crates/interledger-stream/src/server.rs
```diff
@@ -84,25 +85,34 @@ impl ConnectionGenerator {
...
 pub struct PaymentNotification {
+    /// The username of the receiver of the payment notification
     pub to_username: Username,
+    /// The username of the sender of the payment notification
     pub from_username: Username,
+    /// The ILP Address of the receiver of the payment notification
     pub destination: Address,
+    /// The amount received
     pub amount: u64,
+    /// The time this payment notification was fired
```
kincaidoneil: In what format?

crates/interledger-stream/src/server.rs
```diff
@@ -84,25 +85,34 @@ impl ConnectionGenerator {
     }
 }
 
+/// A payment notification data type to be used by Pubsub API consumers
 #[derive(Debug, Deserialize, Serialize)]
 pub struct PaymentNotification {
+    /// The username of the receiver of the payment notification
```
kincaidoneil:
```suggestion
    /// The username of the account that received the Interledger payment
```

crates/interledger-stream/src/server.rs
```diff
@@ -84,25 +85,34 @@ impl ConnectionGenerator {
     }
 }
 
+/// A payment notification data type to be used by Pubsub API consumers
```
kincaidoneil:
```suggestion
/// Notification that STREAM fulfilled a packet and received a single Interledger payment, used by Pubsub API consumers
```

crates/interledger-stream/src/server.rs
```diff
@@ -84,25 +85,34 @@ impl ConnectionGenerator {
...
 #[derive(Debug, Deserialize, Serialize)]
 pub struct PaymentNotification {
+    /// The username of the receiver of the payment notification
     pub to_username: Username,
+    /// The username of the sender of the payment notification
```
kincaidoneil: Is this more accurate?

```suggestion
    /// The username of the account that routed the Interledger payment to this node
```

crates/interledger-stream/src/packet.rs
```diff
@@ -105,23 +112,39 @@ impl<'a> StreamPacketBuilder<'a> {
...
+/// A Stream Packet as specified in its [ASN.1 definition](https://interledger.org/rfcs/asn1/Stream.asn)
 #[derive(PartialEq, Clone)]
 pub struct StreamPacket {
+    /// The cleartext serialized packet
     pub(crate) buffer_unencrypted: BytesMut,
+    /// The packet's sequence number
     sequence: u64,
+    /// The [ILP Packet Type](../interledger_packet/enum.PacketType.html)
     ilp_packet_type: IlpPacketType,
+    /// The amount inside the stream packet
```
kincaidoneil: That's inaccurate

```suggestion
    /// Destination amount of the ILP Prepare, used for enforcing minimum exchange rates and congestion control.
    /// Within an ILP Prepare, represents the minimum amount the recipient needs to receive in order to fulfill the packet.
    /// Within an ILP Fulfill or ILP Reject, represents the amount received by the recipient.
```

crates/interledger-stream/src/packet.rs
```diff
@@ -161,22 +184,28 @@ impl StreamPacket {
...
     pub fn sequence(&self) -> u64 {
         self.sequence
     }
 
+    /// The packet's [type](../interledger_packet/enum.PacketType.html)
     pub fn ilp_packet_type(&self) -> IlpPacketType {
         self.ilp_packet_type
     }
 
+    /// The packet's prepare amount
```
kincaidoneil:
```suggestion
    /// Destination amount of the ILP Prepare, used for enforcing minimum exchange rates and congestion control.
```

crates/interledger-stream/src/packet.rs
```diff
@@ -10,16 +10,23 @@ use lazy_static::lazy_static;
...
+/// The Stream Protocol's version
 const STREAM_VERSION: u8 = 1;
 
+/// Builder for [Stream Packets](https://interledger.org/rfcs/0029-stream/#52-stream-packet)
 pub struct StreamPacketBuilder<'a> {
+    /// The stream packet's sequence number
     pub sequence: u64,
+    /// The [ILP Packet Type](../interledger_packet/enum.PacketType.html)
     pub ilp_packet_type: IlpPacketType,
+    /// The amount inside the stream packet
```
kincaidoneil:
```suggestion
    /// Destination amount of the ILP Prepare, used for enforcing minimum exchange rates and congestion control.
    /// Within an ILP Prepare, represents the minimum amount the recipient needs to receive in order to fulfill the packet.
    /// Within an ILP Fulfill or ILP Reject, represents the amount received by the recipient.
```

crates/interledger-stream/src/crypto.rs
```diff
@@ -96,7 +123,17 @@ fn encrypt_with_nonce(
...
+///
+/// The secret key is generated deterministically by HMAC-256'ing the `shared_secret`
+/// and the hardcoded string "ilp_stream_encryption"
+///
+/// The `additional_data` field is left empty.
+///
+/// The nonce and auth tag are extracted from the first 12 and 16 bytes
+/// of the ciphertext.
 pub fn decrypt(shared_secret: &[u8], mut ciphertext: BytesMut) -> Result<BytesMut, ()> {
+    println!("Got ciphertext with length {:?}", ciphertext.len());
```
kincaidoneil: Why was this added? Should this use a logger?

crates/interledger-stream/src/client.rs
```diff
@@ -48,170 +58,219 @@ impl StreamDelivery {
...
-                rejected_packets: 0,
-                error: None,
-                last_fulfill_time: Instant::now(),
+                // Connection is finally closed, we can now return the receipt and the next service
+                return Ok((sender.receipt, service));
             }
-        })
+        // b. We still need to send more packets!
+        } else {
+            sender.try_send_money().await?
```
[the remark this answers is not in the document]
kincaidoneil: Related Rust question. The congestion controller will limit the amount in flight, and `try_send_money` will return if the amount is 0. But is there a lot of overhead to this loop constantly calling and checking if there's more money available to send? (Seems inefficient?)

crates/interledger-stream/src/crypto.rs
gakonst: Oops this was a debug log, will remove ty

crates/interledger-stream/src/client.rs
```diff
@@ -48,170 +58,219 @@ impl StreamDelivery {
...
     from_account: A,
+    /// The ILP Address of the account sending the payment
     source_account: Address,
+    /// The ILP Address of the account receiving the payment
     destination_account: Address,
+    /// The shared secret generated by the sender and the receiver
     shared_secret: Bytes,
+    /// The amount sent by the sender
     source_amount: u64,
+    /// The [congestion controller](./../congestion/struct.CongestionController.html) for this stream
```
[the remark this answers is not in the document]
gakonst: Expanded on the congestion controller docs

crates/interledger-stream/src/client.rs
```diff
@@ -48,119 +65,142 @@ impl StreamDelivery {
...
     destination_account: Address,
     shared_secret: &[u8],
     source_amount: u64,
-) -> impl Future<Item = (StreamDelivery, S), Error = Error>
+) -> Result<(StreamDelivery, S), Error>
 where
-    S: IncomingService<A> + Clone,
-    A: Account,
+    S: IncomingService<A> + Clone + 'static,
+    A: Account + 'static,
```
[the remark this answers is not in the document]
gakonst: This has to do with us pushing the `next.handle_request` future to the `PendingRequests` vector.  Removing the statics results in:

```
   --> crates/interledger-stream/src/client.rs:243:29
    |
176 | impl<S, A> SendMoneyFuture<S, A>
    |      - help: consider adding an explicit lifetime bound `S: 'static`...
...
243 |                     future: request,
    |                             ^^^^^^^
    |
note: ...so that the type `impl core::future::future::Future` will meet its required lifetime bounds
   --> crates/interledger-stream/src/client.rs:243:29
    |
243 |                     future: request,
    |                             ^^^^^^^

error[E0310]: the parameter type `A` may not live long enough
   --> crates/interledger-stream/src/client.rs:243:29
    |
176 | impl<S, A> SendMoneyFuture<S, A>
    |         - help: consider adding an explicit lifetime bound `A: 'static`...
...
243 |                     future: request,
    |                             ^^^^^^^
    |
note: ...so that the type `impl core::future::future::Future` will meet its required lifetime bounds
   --> crates/interledger-stream/src/client.rs:243:29
    |
243 |                     future: request,
    |                             ^^^^^^^
```

(Note that I've tried rewriting this without `pin_project` (ie replace `self.project()` with `unsafe { self.get_unchecked_mut() }`, and there's no lifetime requirement introduced by it.)

My intuition about this is:

1. `SendMoneyFuture` has a vector of pending requests.
1. Each pending request has a future associated with it, which has static lifetime
1. As a result the `SendMoneyFuture` struct must also live for static.
1. This means that the associated accounts and service also must live for static.

It is not clear to me why this was not required on master before though.

gakonst:
Alternatively to what I wrote above, you could do the following:

```patch
diff --git i/crates/interledger-stream/src/client.rs w/crates/interledger-stream/src/client.rs
index 641e067e..89eafb65 100644
--- i/crates/interledger-stream/src/client.rs
+++ w/crates/interledger-stream/src/client.rs
@@ -73,8 +73,8 @@ pub async fn send_money<S, A>(
     source_amount: u64,
 ) -> Result<(StreamDelivery, S), Error>
 where
-    S: IncomingService<A> + Clone + 'static,
-    A: Account + 'static,
+    S: IncomingService<A> + Clone,
+    A: Account,
 
@@ -233,10 +231,10 @@ where
 
             // Send it!
             self.congestion_controller.prepare(amount);
-            if let Some(ref next) = self.next {
-                let mut next = next.clone();
+            if let Some(ref mut next) = self.next {
                 let from = self.from_account.clone();
                 let request = Box::pin(async move {
                     next.handle_request(IncomingRequest { from, prepare }).await

```
error[E0495]: cannot infer an appropriate lifetime due to conflicting requirements
   --> crates/interledger-stream/src/client.rs:234:25
    |
234 |             if let Some(ref mut next) = self.next {
    |                         ^^^^^^^^^^^^
    |
note: first, the lifetime cannot outlive the anonymous lifetime #1 defined on the method body at 183:5...
   --> crates/interledger-stream/src/client.rs:183:5
    |
183 | /     fn try_send_money(&mut self) -> Result<bool, Error> {
184 | |         let mut sent_packets = false;
185 | |         loop {
186 | |             let amount = min(
...   |
251 | |         Ok(sent_packets)
252 | |     }
    | |_____^
note: ...so that reference does not outlive borrowed content
   --> crates/interledger-stream/src/client.rs:234:25
    |
234 |             if let Some(ref mut next) = self.next {
    |                         ^^^^^^^^^^^^
    = note: but, the lifetime must be valid for the static lifetime...
    = note: ...so that the expression is assignable:
            expected std::pin::Pin<std::boxed::Box<(dyn core::future::future::Future<Output = std::result::Result<interledger_packet::packet::Fulfill, interledger_packet::packet::Reject>> + 'static)>>
               found std::pin::Pin<std::boxed::Box<dyn core::future::future::Future<Output = std::result::Result<interledger_packet::packet::Fulfill, interledger_packet::packet::Reject>>>>
```

But this results in the following error, which makes sense! The `Pin<Box<T>>` must live for static, and we cannot guarantee that. Or can we? (note: we _must_ make it a pinned future to poll the pending requests)

crates/interledger-stream/src/client.rs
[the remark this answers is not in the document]
bstrie: Does this or any of the other 'static bounds cause any problems anywhere else? If not then let's not worry about it, but if so then we could try a few things. I just know that 'static has a tendency to become "infectious" and I have a lot of code yet to review, so I'm not sure if this is going to cause problems downstream.

PR #604, opened 2020-01-23
[Part 11/N] Interledger Service Util: Futures 0.3 Transition 

gakonst:


crates/interledger-service-util/src/validator_service.rs
```diff
@@ -101,68 +101,66 @@ where
...
-                                data: &[],
-                            }
-                            .build())
-                        }
-                    }),
-            )
+            let fulfill = self
+                .next
+                .send_request(request)
+                // TODO: Re-enable timeout!
```
bstrie: Why is this now commented out?

gakonst: Implemented in https://github.com/interledger-rs/interledger-rs/pull/604/commits/f2ac63a53af0b14991eec2e2b3ec74ae246a7222

PR #606, opened 2020-01-23
[Part 13/N] Interledger Store: Futures 0.3 Transition

gakonst:


crates/interledger-store/src/redis/mod.rs
```diff
@@ -1387,163 +1333,103 @@ impl AddressStore for RedisStore {
...
 
-    fn get_local_and_configured_routes(
+    async fn get_local_and_configured_routes(
         &self,
-    ) -> Box<dyn Future<Item = (RoutingTable<Account>, RoutingTable<Account>), Error = ()> + Send>
-    {
-        let get_static_routes = cmd("HGETALL")
+    ) -> Result<(RoutingTable<Account>, RoutingTable<Account>), ()> {
+        let mut connection = self.connection.clone();
+        let static_routes: Vec<(String, RedisAccountId)> = cmd("HGETALL")
```
bstrie: Any reason this wasn't converted to something like `connection.hgetall`, like has been done elsewhere for redis commands in this file?

PR #608, opened 2020-01-23
[Part 15/15] ILP Node: Futures 0.3 Transition

gakonst:
Note that we feature gate monitoring features and disable them in this PR so that we can limit the type length limit bug's surface area


crates/ilp-node/src/node.rs
```diff
@@ -327,257 +314,212 @@ impl InterledgerNode {
...
+        spawn(warp::serve(settlement_api).bind(settlement_api_bind_address));
+
+        // Exchange Rate Polling
+        if let Some(provider) = exchange_rate_provider {
+            let exchange_rate_fetcher = ExchangeRateFetcher::new(
+                provider,
+                exchange_rate_poll_failure_tolerance,
+                store.clone(),
+            );
+            // This function does not compile on 1.39 for some reason.
```
bstrie: Which function, the spawn_interval? Does it compile on 1.40? Can you elaborate?

gakonst: Oops outdated comment, good catch

Issue #454, opened 2019-10-23
Document Prometheus


dora-gt:
It would be better to document `Prometheus`.

https://github.com/interledger-rs/interledger-rs/pull/444#issuecomment-545124494

gakonst:
This is blocked on the node not compiling fast with metrics enabled. Will revisit once we have figured out that problem in https://github.com/interledger-rs/interledger-rs/pull/609

PR #609, opened 2020-01-23
Revive all features and investigate type length limit error

gakonst:
This PR should fail. Exponential type blowup each time we're wrapping our services.

All the [wrapping done in `node.rs` ](https://github.com/interledger-rs/interledger-rs/blob/gakonst/15-futures03-node/crates/ilp-node/src/node.rs#L365-L366)is controlled via the feature-flags `monitoring` and `google-pubsub`.

Related: https://github.com/rust-lang/rust/issues/68508

### DIFF
--- a/crates/ilp-node/src/lib.rs
+++ b/crates/ilp-node/src/lib.rs
@@ -1,4 +1,4 @@
-#![type_length_limit= "5000000"]
+#![type_length_limit = "5000000"]
 mod instrumentation;
 mod node;
 
--- a/crates/ilp-node/src/lib.rs
+++ b/crates/ilp-node/src/lib.rs
@@ -1,7 +1,3 @@
-// #![type_length_limit = "25000000"]
-// #![type_length_limit = "1500000"] // needed to cargo build --bin ilp-node --feature "monitoring"
-#![type_length_limit = "80000000"] // needed to cargo build --all-features --all-targets
-
 mod instrumentation;
 mod node;
 
--- a/crates/ilp-node/src/lib.rs
+++ b/crates/ilp-node/src/lib.rs
@@ -1,3 +1,4 @@
+#![type_length_limit= "5000000"]
 mod instrumentation;
 mod node;
 
--- a/crates/interledger-service/src/lib.rs
+++ b/crates/interledger-service/src/lib.rs
@@ -140,8 +140,8 @@ pub trait IncomingService<A: Account> {
     /// and/or handle the result of calling the inner service.
     fn wrap<F, R>(self, f: F) -> WrappedService<F, Self, A>
     where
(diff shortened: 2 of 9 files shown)


gakonst:
Once this PR is closed, prometheus usage should be documented: https://github.com/interledger-rs/interledger-rs/issues/454

crates/ilp-node/src/lib.rs
```diff
@@ -1,7 +1,4 @@
-// #![type_length_limit = "25000000"]
-// #![type_length_limit = "1500000"] // needed to cargo build --bin ilp-node --feature "monitoring"
-#![type_length_limit = "80000000"] // needed to cargo build --all-features --all-targets
-
+#![type_length_limit = "5000000"]
```
bstrie: Did you experiment to see if these changes remove the need for this attribute at all?

crates/interledger-service/src/lib.rs
```diff
@@ -345,3 +345,209 @@ pub trait AddressStore: Clone {
...
+// the tests still build nearly instantly. The trick is to make the wrapping function
+// take a trait object instead of the concrete type
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use lazy_static::lazy_static;
+    use std::str::FromStr;
+
+    #[test]
+    fn incoming_service_no_exponential_blowup_when_wrapping() {
```
bstrie: It's not a bad idea to try to test for this, but it is kind of weird that this regression test can't ever actually fail... not sure what to do about that.

crates/interledger-service/src/lib.rs
```diff
@@ -345,3 +345,209 @@ pub trait AddressStore: Clone {
     /// (the value is stored in memory because it is read often by all services)
     fn get_ilp_address(&self) -> Address;
 }
+
+// Even though we wrap the types _a lot_ of times in multiple configurations
+// the tests still build nearly instantly. The trick is to make the wrapping function
+// take a trait object instead of the concrete type
```
bstrie: Did you try running these tests before you implemented the boxing in this PR, to make sure that these test cases ever actually triggered the degenerate wrapping behavior?

crates/interledger-service/src/lib.rs
gakonst: Yeah I am using this as guidance for being able to reproduce the issue / ensure that it's not reproducible from just within the interledger-service crate. Otherwise, the issue only manifests at the ilp-node crate.

crates/interledger-service/src/lib.rs
gakonst: Yes. Rust-lang issue here: https://github.com/rust-lang/rust/issues/68508

Issue #356, opened 2019-09-30
Disconnect BTP connections if an account's password changes


emschwartz:
If an account has an open connection but the password is changed by the user or an admin, we should disconnect the BTP connections.

PR #611, opened 2020-01-27
Close BTP connection if a BTP token is provided

gakonst:
If a BTP token is provided, we should just remove the existing connection from the WebSockets HashMap, since a new one will be created by the `connect_to_external_services` call. 

I've added a timeout on the send_request call to not hang forever if the server closes the connection and leaves us with a broken websocket.

Closes https://github.com/interledger-rs/interledger-rs/issues/356


crates/interledger-btp/src/service.rs
```diff
@@ -137,6 +137,12 @@ where
         }
     }
 
+    /// Deletes the websocket associated with the provided `account_id`
+    pub fn close_connection(&self, account_id: &Uuid) {
+        println!("account remvoed {:?}", account_id);
```
gakonst: todo -> remove this

Issue #583, opened 2020-01-07
Implement From<Reject> for common types


gakonst:
Would be useful to find which types get put in map_err's a lot of times, and implement From<Reject>  for them. Related to #58 and #317

Issue #364, opened 2019-09-30
Improve account creation API and errors


emschwartz:
- If you try to create an account with fields that conflict with another account (e.g. same username), you get a 500 error instead of a 400 error
- The `PUT /accounts/:username` cannot be used to insert accounts, only modify existing ones. This is somewhat annoying if you are working on inserting and modifying a given account because you first need to call `POST /accounts` and then `PUT /accounts/:username` instead of being able to call `PUT /accounts/:username` from the start

gakonst:
> If you try to create an account with fields that conflict with another account (e.g. same username), you get a 500 error instead of a 400 error

This has to do with our store traits not returning any meaningful error.

> The PUT /accounts/:username cannot be used to insert accounts, only modify existing ones. This is somewhat annoying if you are working on inserting and modifying a given account because you first need to call POST /accounts and then PUT /accounts/:username instead of being able to call PUT /accounts/:username from the start

This does not make much sense to me, what does insert & modify an account mean? You'd just insert it with the correct info

Issue #317, opened 2019-09-23
Improve Error types throughout the codebase


emschwartz:
There are many places that we return `()` as the Error type from Futures, mostly because I was lazy and didn't want to have to figure out how to do it properly. We should go through the codebase and define proper error types as structs/enums.

As far as I understand, the core Error type has improved somewhat since when this project was started so I don't think we need to use an external crate like `failure` or `quick_error` anymore, though I could be wrong about this.

dora-gt:
We really need this to give more precise error information.
If it means something of security matter, we could utilize server modes like `dev` or `production` or something. 

- `dev` = much information
- `production` -> less information

emschwartz:
For this issues, I am mainly talking about returning better errors within the codebase. Each time we return an `Error = ()`, we should replace that with a more meaningful error. What we return via the API is slightly separate from that and I agree that we should be a little careful with how much detail we return to the caller

dora-gt:
I meant, for example:
https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-service/src/lib.rs#L117-L123
This is used from APIs, and I could not return something meaningful because () tells nothing. It ends up with just returning `Internal Server Error`s which is sometimes not so good for investigating what's happening.

Issue #58, opened 2019-03-27
Error Handling Improvements


ekrenzke:
Hi everybody, I started some work on improving error handling in another WIP branch [here](https://github.com/emschwartz/interledger-rs/tree/ek-error-handling).

# Motivation
I began working on some tests, but found it hard to test some error cases. In an attempt to improve the creation of tests, I took it upon myself to improve error handling in the process.

# Implementation
Due to the layered nature of ILP, I thought it would be useful to formally define the types of errors that could occur within each layer for each crate. The error handling and propagation scheme looks roughly like this:

![error_handling](https://user-images.githubusercontent.com/11590210/55112683-7db6a280-50b3-11e9-8b2e-117e9da3f30d.png)
_edit: I meant `InvalidPacket` and `UnexpectedPacket` in the diagram above..._

Furthermore, each layer of the protocol is restricted to only emitting errors to the layer directly below it. Since no ILP packet handling is done explicitly at the BTP layer, I thought this made sense. Does anybody have different opinions on this?

Since the protocol stack is roughly ranked from bottom to top as: BTP, ILP, STREAM, SPSP, I thought it was appropriate to only allow the layers to communicate errors to the layer immediately beneath. (e.g. STREAM can propagate errors to ILP, which can be propagated to BTP, and handled at that layer.)

For now, I am purposefully omitting BTP Error packets and ILP Error packets, because the builder interface seems more appropriate for naming purposes, but it _does_ seem to lead to some slight inconsistency. Should these cases be handled in the module-level error handling case described above?

Higher-level (sub)protocols and interfaces are still able to utilize error types found in these crates by importing the appropriate error. (e.g. `use ilp::Error as IlpError`)

# Notes
This is also intended to reduce error handling dependencies. I think this is a more reasonable approach since each layer of the protocol is split into crates, therefore error types for each layer should be explicitly defined.

____
Let me know if you have any thoughts, concerns, or ideas, before I proceed.

emschwartz:
The changes in that branch look good so far!

> Furthermore, each layer of the protocol is restricted to only emitting errors to the layer directly below it. 

I'm not totally sure what you mean by this.

> For now, I am purposefully omitting BTP Error packets and ILP Error packets, because the builder interface seems more appropriate for naming purposes, but it does seem to lead to some slight inconsistency.

[...]

PR #613, opened 2020-01-28
Add errors to the store

gakonst:
The objective of this PR is to solidify the errors across the codebase (while now they're just `()`). In addition, we'd like to reduce the usage of `map_err`, since it can get annoying to have to do that everywhere. 

In addition, when consuming store errors at higher level crates such as the API we'd lose information, e.g. we couldn't know if the store call errored because an account was not found, if the account's id was not found, or if the socket was dropped.

To do that, the following plan is executed:
- [x] We create a new `interledger-errors` crate which will house all of these errors. 
     - This is consumed by most crates and is very low in the dependency chain. I decided to create this due to wanting to implement `From<RedisError> for T`, `From<T> for warp::Reject` and `From<T> for ApiError`, where `T` is each error type. 
    - Had I not introduced a separate crate, I'd have to add a new module for errors in each crate and add `warp` as a direct dependency to them which I preferred to avoid, and keep it as an implicit dependency via the errors crate.
- [x] Each error has variants for all possible cases it may encounter
- [x] Each error matches against internal error `Other` types and casts it to its own, if it know which one it is, e.g. passing a `AccountStoreError::AccountNotFound` to a `BtpStoreError` would normally go to the `Other` variant, while in this case we'd like it to propagate to the `BtpStoreError::AccountNotFound` variant
- [x] Each of these errors utilizes `thiserror`. It contains either specific errors for the errors that type cares about, or an `Other` variant which takes a `Box<dyn StdError>`, to bubble up errors which come directly from the store, redis for example
- [x] Remove lots of `map_err` calls from API store calls. 
- [x] Remove lots of `map_err` calls from Redis store calls.

As a driveby change, I decided to simplify a few traits, specifically:
1. `BalanceStore`: We had a longstanding TODO which required provided an Account as associated type, while all that's needed is the account's id. I've removed that associated type and dependency on the `AccountStore` trait and replaced the argument with `Uuid`, for the account's id. 
1. `NodeStore`: Previously depended on `AddressStore`, that was not required

Closes: #58 
Closes: #583 
Closes: #317 
Closes https://github.com/interledger-rs/interledger-rs/issues/364

### DIFF
--- a/crates/interledger-errors/src/address_store_error.rs
+++ b/crates/interledger-errors/src/address_store_error.rs
@@ -3,8 +3,9 @@ use interledger_packet::Address;
 use std::error::Error as StdError;
 use thiserror::Error;
 
-/// Errors for the AccountStore
+/// Errors for the AddressStore
 #[derive(Error, Debug)]
+#[non_exhaustive]
 pub enum AddressStoreError {
     #[error("{0}")]
     Other(#[from] Box<dyn StdError + Send>),
@@ -36,6 +37,7 @@ impl From<AddressStoreError> for ApiError {
     }
 }
 
+#[cfg(feature = "warp_errors")]
 impl From<AddressStoreError> for warp::Rejection {
     fn from(src: AddressStoreError) -> Self {
         ApiError::from(src).into()
--- a/crates/interledger-errors/src/address_store_error.rs
+++ b/crates/interledger-errors/src/address_store_error.rs
@@ -1,0 +1,43 @@
+use super::{ApiError, NodeStoreError};
+use interledger_packet::Address;
+use std::error::Error as StdError;
+use thiserror::Error;
+
+/// Errors for the AccountStore
+#[derive(Error, Debug)]
+pub enum AddressStoreError {
+    #[error("{0}")]
+    Other(#[from] Box<dyn StdError + Send>),
+    #[error("Could not save address: {0}")]
+    SetAddress(Address),
+    #[error("Could not save address: {0}")]
+    ClearAddress(Address),
+}
+
+impl From<NodeStoreError> for AddressStoreError {
+    fn from(src: NodeStoreError) -> Self {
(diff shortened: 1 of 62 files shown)


bstrie:
I like the choice of thiserror, but I'm a little confused at the goal here and the focus on the store specifically. For error-handling improvements in general I would proceed like this:

1. Go one crate at a time, working my way up from the bottom of our crate hierarchy
2. For every public item in `interledger-foo` that returns a `Result<_, ()>`, I would change the `()` to `interledger_foo::Error`.

In particular I wouldn't worry about `From<Reject>` for the moment, since we have so many places that just uselessly return a `()` which seems a lot worse (and which also means a ton of extra map_errs to paper over the lack of proper errors).

gakonst:
Interledger-CCP is one of the implementations which has multiple `Result<.*, ()>` functions, however most of these are not exposed in the public api. The functionality which is exposed everywhere is the traits for the Stores. As a result I think that revamping the stores first is a good course of action.

bstrie:
> As a result I think that revamping the stores first is a good course of action. One a sidenote, this is going to be another fairly large PR

Will you be making this your next highest priority? Not sure what else is in the pipe that would compete with it for attention (other than the SQL stuff, which I'll be tackling).

gakonst:
Yes this is my highest priority at the moment

crates/interledger-errors/src/address_store_error.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl From<AddressStoreError> for ApiError {
+    fn from(src: AddressStoreError) -> Self {
+        // AddressStore erroring is always an internal server error
+        ApiError::internal_server_error().detail(src.to_string())
+    }
+}
+
+impl From<AddressStoreError> for warp::Rejection {
```
bstrie: Should this be feature-gated?

PR #617, opened 2020-02-03
Expand Tests and remove some redundant clone's

### LINKED ISSUE #610: Expand Service Util and Store Tests
The RateLimitService, BalanceService, MaxPacketAmountService are all missing tests.

Same applies to the store implementations, as they just check for errors, but not specifically for which errors.

depends on #613 

- [x] RateLimitService
- [x] BalanceService
- [x] MaxPacketAmountService

gakonst:
- [x] Expands Service-Util tests
- [x] Expands Redis store tests
- [x] Improves redis efficiency by using references in many places instead of passing by value

Closes #610

### DIFF
--- a/crates/interledger-store/src/redis/mod.rs
+++ b/crates/interledger-store/src/redis/mod.rs
@@ -83,7 +83,7 @@ fn uncredited_amount_key(account_id: impl ToString) -> String {
 }
 
 /// Domain separator for idempotency keys
-fn prefixed_idempotency_key(idempotency_key: String) -> String {
+fn prefixed_idempotency_key(idempotency_key: &str) -> String {
     format!("idempotency-key:{}", idempotency_key)
 }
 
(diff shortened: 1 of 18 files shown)


crates/interledger-store/tests/redis/settlement_test.rs
```diff
@@ -142,23 +131,20 @@ async fn idempotent_settlement_calls() {
...
         .await
         .unwrap();
-    let (balance, prepaid_amount): (i64, i64) = cmd("HMGET")
-        .arg(format!("accounts:{}", id))
-        .arg("balance")
-        .arg("prepaid_amount")
-        .query_async(&mut conn)
+
+    let (balance, prepaid_amount): (i64, i64) = connection
+        .hget(format!("accounts:{}", id), &["balance", "prepaid_amount"])
```
bstrie: I don't suppose using .hget rather than .hget_multiple here matters?

crates/interledger-store/src/redis/mod.rs
```diff
@@ -397,11 +388,14 @@ impl RedisStore {
...
         pipe.hset(
             ROUTES_KEY,
-            account.ilp_address.to_bytes().to_vec(),
+            account.ilp_address.as_bytes(),
             RedisAccountId(account.id),
         )
         .ignore();
 
+        // Set account details
+        pipe.cmd("HMSET").arg(&id).arg(encrypted).ignore();
```
bstrie: Is it ok that this has been moved after the hset_multiple on line 370?

crates/interledger-store/src/redis/mod.rs
```diff
@@ -1819,7 +1796,7 @@ async fn update_routes(
 // As a result, we wrap Uuid in a local data type, and implement the necessary
 // traits for that.
 #[derive(Eq, PartialEq, Hash, Debug, Default, Serialize, Deserialize, Copy, Clone)]
-struct RedisAccountId(Uuid);
+pub struct RedisAccountId(pub Uuid);
```
bstrie: Does this need to become public? Can it just be pub(crate)?

crates/interledger-store/tests/redis/settlement_test.rs
gakonst: It chooses the command under the hood: https://github.com/mitsuhiko/redis-rs/blob/master/src/commands.rs#L385

crates/interledger-store/src/redis/mod.rs
gakonst: Yes, it shouldn't matter. They are setting distinct fields under the same key.

Issue #586, opened 2020-01-17
Add endpoint to adjust tracing level


gakonst:
Currently we initialize `tracing` with whatever log level is given via the `RUST_LOG` env variable. It'd be nice if we had one more [admin API endpoint](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-api/src/routes/node_settings.rs), at `/tracing-level` which would allow the node operator to modify it. 

- https://docs.rs/tracing-subscriber/0.2.0-alpha.4/tracing_subscriber/reload/index.html method to change the tracing level
- probably outdated since it's commented out, example from docs https://github.com/tokio-rs/tracing/blob/master/examples/examples/tower-load.rs#L40-L41

matdehaast:
This would be really handy as usually I don't want to bounce the process to debug an issue

dora-gt:
Hmmm I’m not sure if we want to set the level after launching nodes. Isn’t it sufficient to modify `RUST_LOG`? Do you want to **after launching**? @matdehaast

gakonst:
Yes, this is for modifyign the log levels after the node has been launched, without having to restart the node.

PR #621, opened 2020-02-04
Adjust tracing at runtime

gakonst:
Closes https://github.com/interledger-rs/interledger-rs/issues/586

I kinda wish there was a better way to do multi-line conditional compilation blocks


crates/ilp-node/src/node.rs
```diff
@@ -488,7 +499,65 @@ impl InterledgerNode {
...
+        #[cfg(feature = "monitoring")]
+        let adjust_tracing = warp::put()
+            .and(warp::path("tracing-level"))
+            .and(warp::path::end())
+            .and(warp::header::<SecretString>("authorization"))
+            .and(warp::body::bytes())
+            .and_then(
+                move |auth_header: SecretString, new_level: bytes05::Bytes| {
+                    let handle = handle.clone();
+                    let admin_auth_token = admin_auth_token.clone();
```
bstrie: I'm not sure how this line could be doing anything, `admin_auth_token` has already been moved into this closure, and making a clone that shadows it by using the exact same name must be a no-op.

PR #624, opened 2020-02-06
Update dependencies

bstrie:
This PR updates our dependencies to their newest major version, fixes any breakage that results, and regenerates our lockfile using the new lockfile format (which will be understood by any toolchain >= 1.38, so this doesn't change our minimum supported Rust).


gakonst:
Part of https://github.com/interledger-rs/interledger-rs/issues/140

PR #625, opened 2020-02-06
refactor: [WIP] migrate interledger-packet from bytes 0.4 to 0.5

bstrie:
This PR migrates interledger-packet from bytes 0.4 to bytes 0.5, to the extent that `cargo build --package interledger-packet` and `cargo test --package interledger-packet` pass. However it doesn't update any other crates, so the rest of the repo will still fail to build. I'll be opening a tracking issue to separately track the migration, for now this PR just exists to get the code somewhere visible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,7 +189,6 @@ dependencies = [
  "byteorder",
  "either",
  "iovec",
- "serde",
 ]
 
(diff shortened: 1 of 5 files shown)


bstrie:
Closing for now, but tracked in #625.

PR #635, opened 2020-02-28
STREAM sender reliability improvements

### LINKED ISSUE #467: STREAM: Retry R00 Errors
In [this test](https://circleci.com/gh/interledger-rs/interledger-rs/2641) run, the STREAM client stopped because of an R00 error:

```
[DEBUG interledger_stream::client] Prepare 1 with amount 2000 was rejected with code: R00 (2000 left to send)
[ERROR interledger_stream::client] Send money stopped because of error: SendMoneyError("Packet was rejected with error: R00 ")
[ERROR interledger_spsp::client] Error sending payment: SendMoneyError("Packet was rejected with error: R00 ")
```

I'm not sure why it timed out, but in any case the STREAM sender should retry on `RXX` errors.

### LINKED ISSUE #513: The congestion controller's minimum value is 1, but if the sender sends a prepare packet with a value of 1, it MAY get a T04 error.
If the sender sends a prepare packet with a value of 1, there's a good chance that it will be rounded down (when the path exchange rates are applied) to 0. So the sender will be sending lots of packets but effectively not delivering anything. The following code for error T04.
https://github.com/interledger-rs/interledger-rs/blob/bd7ab9a9e86a158c61e4b9494d03bfbbbb700d55/crates/interledger-stream/src/congestion.rs#L114-L128

Potential solution 1: The reject packet can have a separate error code for such a situation (T05: minimum acceptable amount?), which would mean that the sender should increase/adjust the amount in the packet accordingly. 

A) This reject packet MAY indicate the details that the sender can use to compute the path exchange rate. It’s more in the vein of path exchange rate discovery [here](https://interledger.org/rfcs/0029-stream/#34-exchange-rates) and the STREAM packets already have that capability. The sender can then adjust the minimum value for the next prepare packet based on:
a) calculated exchange rate from reject packet, and 
b) minimum acceptable value for a prepare packet to be fulfilled.

https://github.com/interledger-rs/interledger-rs/blob/bd7ab9a9e86a158c61e4b9494d03bfbbbb700d55/crates/interledger-stream/src/client.rs#L84-L93

B) If the reject packet does not explicitly indicate the necessary details to compute the path exchange rate information (for some reason), then the sender can slowly start increasing the amount on the packets so as to see when it starts to get the fulfill packets again.
@emschwartz

### LINKED ISSUE #633: STREAM fail-fast should ignore temporary errors
The fail-fast behavior I implemented of 200+ packets and 99% failure rate doesn't work under liquidity errors/waiting for settlements. If there are insufficient liquidity errors that take a second or more to resolve, it'll hit the 99%+ error rate since the Rust implementation retries packets so quickly.

T04s (and maybe other temporary errors) should _not_ affect the fail-fast behavior.

### LINKED ISSUE #634: How to handle payment "dust"?
When enforcing minimum exchange rates, it's possible that--by chance--at the end of the payment, the remaining amount to send becomes too small to safely deliver it the recipient. i.e., the remaining is less the minimum source amount required to deliver money to the recipient without the exchange rate failing due to rounding.

In effect, this is payment "dust." How should this be handled? Should the STREAM sender anticipate the dust and distribute the remaining amount among more, slightly smaller packets?

kincaidoneil:
- Fix #513: Compute minimum source amount necessary to deliver money, corresponding to the minimum destination amount. This enables STREAM payments across peers in which a credit limit is hit and settlement occurs during the payment. Previously, T04s would reduce the source amount to 1, which will always fail due to rounding or not meeting the minimum exchange rate.
- Fix #634: Distribute dust amounts across remaining packets. By chance, the last packet in a STREAM payment may be too small that it won't meet the minimum destination amount due to rounding errors. This heuristic approximates in advance if this may happen and increases each packet amount slightly so there's no dust amount.
- Fix #633: Only apply T00, T01, F99 to fail-fast behavior. e.g. T04s should not trigger the fail-fast behavior, since they take time to resolve, not simply a number of packets -- this was a bug.
- Fix #467: Continue sending on R01 errors, which are triggered when the router receives a packet that rounds to 0. For other relative errors, e.g. R00, R02, the behavior stays the same, since they're likely terminal.
- Increase default slippage to 1.5%. This is so if operators set a 1% spread, the sender works by default.

### DIFF
--- a/crates/interledger-api/src/routes/accounts.rs
+++ b/crates/interledger-api/src/routes/accounts.rs
@@ -29,7 +29,7 @@ use warp::{self, reply::Json, Filter, Rejection};
 pub const BEARER_TOKEN_START: usize = 7;
 
 const fn get_default_max_slippage() -> f64 {
-    0.01
+    0.015
 }
 
 #[derive(Deserialize, Debug)]
--- a/crates/interledger-stream/src/client.rs
+++ b/crates/interledger-stream/src/client.rs
@@ -14,8 +14,10 @@ use interledger_service::*;
 use log::{debug, error, warn};
 use num::rational::BigRational;
 use num::traits::cast::{FromPrimitive, ToPrimitive};
-use num::traits::identities::One;
+use num::traits::identities::{One, Zero};
+use num::traits::ops::checked::CheckedDiv;
 use num::traits::pow::pow;
+use num::BigInt;
 use serde::{Deserialize, Serialize};
 use tokio::sync::Mutex;
 use tokio::time::timeout_at;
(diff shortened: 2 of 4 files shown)


gakonst:
LGTM! These last few changes on STREAM really improved it

koivunej:
Tagging this with `breaking` for  `crates/interledger-stream/src/congestion.rs` changes, which I understand to be public.

PR #637, opened 2020-04-08
Replace log dependency with tracing dependency

KristianWahlroos:
Previously Interledger used partially logs and partially tracing. This PR switches it to use only tracing. Later this change will allow easy enhancing or adding of the tracing spans, which wasn't done on this PR.


KristianWahlroos:
There is deprecated code in Interledger. Did separate pull request to fix this #637

aphelionz:
Might somebody be able to re-run CI on this?

PR #639, opened 2020-05-10
Fix/indefinitely growing logs

KristianWahlroos:
Fix a problem with indefinitely growing logs. The previous versions with enters didn't exit spans in async context. This one exits.

### DIFF
--- a/crates/ilp-node/src/instrumentation/google_pubsub.rs
+++ b/crates/ilp-node/src/instrumentation/google_pubsub.rs
@@ -1,5 +1,3 @@
-#[cfg(feature = "google_pubsub")]
-use base64;
 use chrono::Utc;
 use futures::{compat::Future01CompatExt, Future, TryFutureExt};
 use interledger::{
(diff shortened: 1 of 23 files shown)


matdehaast:
@kincaidoneil

crates/ilp-node/src/instrumentation/google_pubsub.rs
```diff
@@ -1,5 +1,3 @@
-#[cfg(feature = "google_pubsub")]
```
kincaidoneil, line 1: What's the effect of removing this? Will this file always be compiled?

KristianWahlroos: This ([attributes](https://doc.rust-lang.org/reference/attributes.html)) only affects the next thing in code, which was the now removed `use base64;`

koivunej:
Tagging as breaking for the change in `crates/interledger-stream/src/packet.rs`.

Issue #615, opened 2020-01-30
Don't clone the entire service chain on every packet


kincaidoneil:
Since calling `handle_request` on the service interface requires a mutable borrow, we're cloning the entire service chain on every packet to get around this.

For ILP-over-HTTP requests, it happens _twice_:

https://github.com/interledger-rs/interledger-rs/blob/4b491ab76ca4b73ef53e101fc2cb1a71e32f53d2/crates/interledger-http/src/server.rs#L124-L135

https://github.com/interledger-rs/interledger-rs/blob/4b491ab76ca4b73ef53e101fc2cb1a71e32f53d2/crates/interledger-http/src/server.rs#L63-L74

Here's where it happens for BTP packets:

https://github.com/interledger-rs/interledger-rs/blob/4b491ab76ca4b73ef53e101fc2cb1a71e32f53d2/crates/interledger-btp/src/service.rs#L258-L259

This probably isn't great for memory usage or performance (and is probably much worse than #469 or #554). We should consider redesigning the service interface so it doesn't require a mutable borrow, since it appears very few (if any?) services require mutation.

gakonst:
It's not the mutable borrow that makes us require the clone, it's the requirement of the closure to be `Fn`. If you don't clone and pass by value it'll obviously be `FnOnce`, which doesn't work. Alternatively, you could try adding references to work around this, but I haven't figured out how to make lifetimes work in this case. There might be a way to get it working, but do note that even the Warp docs capture the state with a clone.

PR #642, opened 2020-05-15
perf: remove a few clones of the incoming service chain

ljedrz:
Address a bit of #615 by removing 2 unnecessary allocations of the incoming service.


aphelionz:
@ljedrz same here - if you want to rebase this I will merge

PR #644, opened 2020-05-18
Add time-based settlement to complement the threshold-based

koivunej:
Currently interledger-rs supports settling peer accounts whenever their balance goes over `settle_threshold`, settling for `balance - settle_to` and thus resetting the `balance` to `settle_to`. This PR adds a time-based method which works in addition to the existing threshold-based settlement. This feature allows driving the balance to `settle_to` later. The feature works with  `settle_threshold`: if the balance goes over the threshold, a settlement is still fired off right away and any existing time-based settlement is cancelled.

The feature works by adding a delay on each prepare for every peering account id which did not already have a delay. The delay duration is set through the `--settle_every` configuration option. When the delay expires, a path similar to threshold settlement is taken to settle the balances, but now the balance's relation to the threshold is ignored.

The hot path of the prepare/fulfill processing in balance service is changed by addition of a non-blocking non-awaiting call to [Sender::try_send](https://docs.rs/tokio/0.2.21/tokio/sync/mpsc/struct.Sender.html#method.try_send), and on failing the send there is protection against flooding the logging by a non-awaited try to lock the mutex to log at most every minute. This guarding is because we don't anticipate running into issues with processing of the queue – the situtation should only arise when the node is badly overloaded, and in that case the logging should try not to make the bad situtation worse.

In balance_service there's additionally minor refactoring to share the `settle_or_rollback` for threshold and time based approaches which involved a few lines moved.

Note: In a cluster configuration where multiple nodes share a single database and database accounts, using this can result in many settlements.

### DIFF
--- a/crates/ilp-node/src/main.rs
+++ b/crates/ilp-node/src/main.rs
@@ -133,8 +133,12 @@ async fn main() {
             .long("settle_every")
             .takes_value(true)
             .help("Settlement delay, in seconds, the peering accounts will be settled after \
-                this many seconds after the first fulfill packet unless the balance exceeds \
-                settlement threshold before."),
+                this many seconds after the first fulfill packet unless the balance had \
+                exceeded the settlement threshold.\n\n\
+                \
+                Note: In a cluster configuration where multiple nodes share a \
+                single database and database accounts, using this can result in \
+                many settlements."),
         ]);
 
     let mut config = get_env_config("ilp");
--- a/crates/ilp-node/src/main.rs
+++ b/crates/ilp-node/src/main.rs
@@ -129,6 +129,12 @@ async fn main() {
                 old data. For example, a value of 1000ms (1 second) would mean that the \
                 node forgets the oldest 1 second of histogram data points every second. \
                 Defaults to 10000ms (10 seconds)."),
+        Arg::with_name("settle_every")
+            .long("settle_every")
+            .takes_value(true)
+            .help("Settlement delay, in seconds, the peering accounts will be settled after \
+                this many seconds after the first fulfill packet unless the balance exceeds \
+                settlement threshold before."),
         ]);
 
     let mut config = get_env_config("ilp");
--- a/crates/ilp-node/src/main.rs
+++ b/crates/ilp-node/src/main.rs
@@ -132,7 +132,7 @@ async fn main() {
         Arg::with_name("settle_every")
             .long("settle_every")
             .takes_value(true)
-            .help("Settlement delay, in seconds, the peering accounts will be settled after \
+            .help("Settlement delay, in seconds; the peering accounts will be settled after \
                 this many seconds after the first fulfill packet unless the balance had \
                 exceeded the settlement threshold.\n\n\
                 \
--- a/crates/interledger-service-util/src/balance_service.rs
+++ b/crates/interledger-service-util/src/balance_service.rs
@@ -8,7 +8,9 @@ use interledger_settlement::core::{
     SettlementClient,
 };
 use std::marker::PhantomData;
-use std::{fmt, time::Duration};
+use std::sync::{Arc, Mutex};
+use std::{fmt, time::Duration, time::Instant};
+use tokio::sync::mpsc::error::TrySendError;
 use tracing::{debug, error, info, trace, warn};
 use uuid::Uuid;
 
@@ -65,6 +67,7 @@ pub struct BalanceService<S, O, A> {
     settlement_client: SettlementClient,
     policy: Policy,
     account_type: PhantomData<A>,
+    channel_last_fail: Arc<Mutex<Instant>>,
 }
 
 impl<S, O, A> BalanceService<S, O, A>
@@ -75,7 +78,7 @@ where
 {
     pub fn new(
         store: S,
-        sender: Option<tokio::sync::mpsc::UnboundedSender<ManageTimeout>>,
+        sender: Option<tokio::sync::mpsc::Sender<ManageTimeout>>,
         next: O,
     ) -> Self {
         BalanceService {
@@ -87,6 +90,7 @@ where
                 None => Policy::ThresholdOnly,
             },
             account_type: PhantomData,
+            channel_last_fail: Arc::new(Mutex::new(Instant::now())),
         }
     }
 }
@@ -171,6 +175,7 @@ where
                         to,
                         settlement_client,
                         self.policy.clone(),
+                        self.channel_last_fail.clone(),
                     );
                 }
 
@@ -211,6 +216,7 @@ fn settle_or_rollback_later<Acct, Store>(
     to: Acct,
     settlement_client: SettlementClient,
     policy: Policy,
+    channel_last_fail: Arc<Mutex<Instant>>,
 ) where
     Acct: SettlementAccount + Send + Sync + 'static,
     Store: BalanceStore + SettlementStore<Account = Acct> + Send + Sync + 'static,
@@ -223,6 +229,7 @@ fn settle_or_rollback_later<Acct, Store>(
         to,
         settlement_client,
         policy,
+        channel_last_fail,
     ));
 }
 
@@ -233,7 +240,8 @@ async fn settle_or_rollback_now<Acct, Store>(
     from_id: Uuid,
     to: Acct,
     settlement_client: SettlementClient,
-    policy: Policy,
+    mut policy: Policy,
+    channel_last_fail: Arc<Mutex<Instant>>,
 ) -> Result<(), ()>
 where
     Acct: SettlementAccount + Send + Sync + 'static,
@@ -267,13 +275,13 @@ where
             //
             // the new settlement thing will need to settle any balance x to settle_to,
             // where x >= settle_to
-            policy.settle_later(to.id());
+            policy.settle_later(to.id(), channel_last_fail);
         }
         return Ok(());
     }
 
     // cancel a pending settlement always before trying it
-    policy.clear_later(to.id());
+    policy.clear_later(to.id(), channel_last_fail);
 
     settle_or_rollback(store, to, amount_to_settle, settlement_client).await
 }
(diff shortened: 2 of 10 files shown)


kincaidoneil:
Implementation looked fine at first glance, but a couple questions:

- If multiple node instances are running and share the same Redis database, couldn't an account be settled more frequently than the configured interval? Would that be an issue?
- More broadly, what specific use cases do you have in mind for this feature? Could it be better served by logic within a settlement engine to only execute settlements at a particular frequency?

darentuzi:
Its useful for some use cases where the connector has some business logic where it needs to settle its debts every day/week/month even if it hasn't reached a specific threshold.
Since currently all the logic is on the node accounts and not on the settlement engine we thought it would be more appropriate for this one to be there too.

aphelionz:
Just a friendly ping to see if we can get an updated review :)

aphelionz:
@kincaidoneil Checking in. Is this still something you'd like in ILP? If so we can rebase to master again and continue the review.

koivunej:
The test failure is transient setup issue (had to login to circleci to see this):

```
Updating crates.io index
Downloading crates ...
warning: spurious network error (2 tries remaining): [28] Timeout was reached (failed to download any data for `cargo-audit v0.12.0` within 30s)
Downloaded cargo-audit v0.12.0
Installing cargo-audit v0.12.0
Downloading crates ...
warning: spurious network error (2 tries remaining): failed to get 200 response from `https://crates.io/api/v1/crates/unicode-normalization/0.1.13/download`, got 504
Downloaded fnv v1.0.7
Downloaded glob v0.3.0
warning: spurious network error (1 tries remaining): failed to get 200 response from `https://crates.io/api/v1/crates/unicode-normalization/0.1.13/download`, got 504
Downloaded itoa v0.4.6
error: failed to compile `cargo-audit v0.12.0`, intermediate artifacts can be found at `/tmp/cargo-installYutqjG`
Caused by:
failed to download from `https://crates.io/api/v1/crates/unicode-normalization/0.1.13/download`
Caused by:
failed to get 200 response from `https://crates.io/api/v1/crates/unicode-normalization/0.1.13/download`, got 504
Exited with code exit status 101
CircleCI received exit code 101
```

aphelionz:
@koivunej @darentuzi and @kincaidoneil do we still want this feature?

kincaidoneil:
> - If multiple node instances are running and share the same Redis database, couldn't an account be settled more frequently than the configured interval? Would that be an issue?

@aphelionz I'm not opposed to merging, but could this comment be addressed?

aphelionz:
Of course, not trying to push this in too fast :) Just wanted to make sure it was something we still want.

darentuzi:
@kincaidoneil Can you elaborate on why do you think multiple nodes using the same Redis would trigger it more frequently?

kincaidoneil:
@darentuzi I was thinking that in a deployment of multiple nodes using same Redis database for accounts, you'd have an issue where one node wouldn't know if another had already performed a settlement within the time period. The nodes wouldn't settle more than they should, since that's still enforced by `settle_to`; they might just settle more frequently than they should.

On second thought, one answer to this is only one node should have `--settle_every` configured, and other nodes should have a prohibitively high `--settle_every` configured so they never try to settle. This seems a little awkward though.

darentuzi:
@kincaidoneil You are right that it would settle more frequently if multiple nodes are using same db. 
Is it common practice for multiple nodes to use the same db? I guess horizontal scaling scenarios. 
If we made it extra clear in field description/docs that anyone who turns this feature on should know that IF they have multiple nodes on the same db, they have to be aware that time based settlement will occur more frequently. Do you think that would be enough to move this forward or should we find a solution for that case too?

kincaidoneil:
Adding a note in the description sounds good 👍

koivunej:
Rebased with minor changes:
 - Cargo.lock changes were conflicting, outdated

Added a note about clustering to `ilp-node --help` where it now reads:

```
--settle_every <settle_every>
    Settlement delay, in seconds, the peering accounts will be settled after this many seconds after the first
    fulfill packet unless the balance had exceeded the settlement threshold.

    Note: In a cluster configuration where multiple nodes share a single database and database accounts, using
    this can result in surprisingly many settlements.
```

After looking at the code, recorded a FIXME of a one way to handle the too many settlements in clustered environments, which would be to have a property/hash key on account requiring that the settlements happen a minimum time period apart. 

Lets see if there's anything in the CI...

koivunej:
Build failed because of new clippy lints, all seemed unrelated..  I'll do another PR.

crates/ilp-node/src/main.rs
```diff
@@ -129,6 +129,16 @@ async fn main() {
                 old data. For example, a value of 1000ms (1 second) would mean that the \
                 node forgets the oldest 1 second of histogram data points every second. \
                 Defaults to 10000ms (10 seconds)."),
+        Arg::with_name("settle_every")
+            .long("settle_every")
+            .takes_value(true)
+            .help("Settlement delay, in seconds, the peering accounts will be settled after \
```
ljedrz, line 135:
```suggestion
            .help("Settlement delay, in seconds; the peering accounts will be settled after \
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -211,6 +207,429 @@ where
...
+    Acct: SettlementAccount + Send + Sync + 'static,
+    Store: BalanceStore + SettlementStore<Account = Acct> + Send + Sync + 'static,
+{
+    let (balance, amount_to_settle) = store
+        .update_balances_for_fulfill(to.id(), outgoing_amount)
+        .map_err(|err| error!("Error applying balance changes for fulfill from account: {} to account: {}. Incoming amount was: {}, outgoing amount was: {}. Error: {}", from_id, to.id(), incoming_amount, outgoing_amount, err))
+        .await?;
+
+    // this message is really important, if you want to recover the balance after a crash; all of
+    // the "amount that need to be settled" must be summed and added to accounts "balance".
```
ljedrz, line 258:
```suggestion
    // the "amount that need to be settled" must be summed and added to the account's "balance".
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -211,6 +207,429 @@ where
...
+                .await?;
+        } else {
+            info!(
+                "Settlement for account {} for {} succeeded",
+                to.id(),
+                amount
+            );
+        }
+    } else {
+        debug!("Settlement for account {} for {} failed as the account has no settlement-engine details",
```
ljedrz, line 347:
```suggestion
        debug!("Settlement for account {} for {} failed as the account has no settlement engine details",
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -211,6 +207,429 @@ where
...
+    Other(tokio::time::Error),
+}
+
+impl fmt::Display for ExitReason {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        use ExitReason::*;
+        match *self {
+            InputClosed => write!(fmt, "Input was closed and ran out of timed settlements"),
+            Shutdown => write!(fmt, "Tokio is shutting down"),
+            Capacity => write!(fmt, "Timer service is over capacity"),
```
ljedrz, line 472:
```suggestion
            Capacity => write!(fmt, "Timer service capacity exceeded"),
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -211,6 +207,429 @@ where
...
+            Other(ref e) => write!(fmt, "Other: {}", e),
+        }
+    }
+}
+
+/// Start a background task for time based settlement. If time based settlement is configured but
+/// this task is never started the time based settlement does not happen and a warning is logged
```
ljedrz, line 479:
```suggestion
/// this task is never started, the time based settlement does not happen and a warning is logged
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -211,6 +207,429 @@ where
...
+        + Clone
+        + Send
+        + Sync
+        + 'static,
+    Acct: SettlementAccount + Send + Sync + 'static,
+{
+    let client = SettlementClient::default();
+    tokio::spawn(async move {
+        info!(
+            "Starting to run delayed settlements with timeout of {:?}",
```
ljedrz, line 500:
```suggestion
            "Starting to run delayed settlements with a timeout of {:?}",
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -211,6 +207,429 @@ where
...
+                                    );
+                                    Err(())
+                                }
+                            }?;
+
+                            let (balance, amount_to_settle) = store.update_balances_for_delayed_settlement(id).await
+                                .map_err(|e| warn!("Time based settlement failed for {}: {}", id, e))?;
+
+                            debug!(
+                                "Account {} balance at time based settlement: {}, Amount that needs to be settled: {}",
```
ljedrz, line 603:
```suggestion
                                "Account {} balance at time based settlement: {}, amount that needs to be settled: {}",
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -211,6 +207,429 @@ where
...
+                                    warn!(
+                                        "Failed to load account {} for time based settlement: {}",
+                                        id, e
+                                    );
+                                    Err(())
+                                }
+                            }?;
+
+                            let (balance, amount_to_settle) = store.update_balances_for_delayed_settlement(id).await
+                                .map_err(|e| warn!("Time based settlement failed for {}: {}", id, e))?;
```
ljedrz, line 600:
```suggestion
                                .map_err(|e| warn!("Time-based settlement failed for {}: {}", id, e))?;
```

crates/interledger-service-util/src/balance_service.rs
```diff
@@ -211,6 +207,429 @@ where
...
+                                Ok(accounts) => {
+                                    error!(
+                                        "Asked for account {} for delayed settlement got back {} accounts: {:?}",
+                                        id, accounts.len(), accounts
+                                    );
+                                    Err(())
+                                }
+                                Err(e) => {
+                                    warn!(
+                                        "Failed to load account {} for time based settlement: {}",
```
ljedrz, line 592:
```suggestion
                                        "Failed to load account {} for time-based settlement: {}",
```

koivunej:
Checked how this behaves with `cargo test --no-default-features`: there are a few existing unused warnings and this adds one for `use std::num::NonZeroU32;`. I'll add a feature gate to that import as well, but not doing anything for the others, they might need a bit more work.

Rebasing on top of master as soon as I handle that `cargo audit` which failed the last build: https://rustsec.org/advisories/RUSTSEC-2020-0091 (not published yet at least).

koivunej:
And another forced push as I forgot the sign-offs from the last two commits.

ljedrz (approved):
LGTM :+1:

pradovic (approved):
LGTM!

aphelionz:
Great! Let's merge on Monday unless @kincaidoneil or @emschwartz have objections

koivunej:
A reminder to anyone susbscribed: we will be merging this in today, please let any of your objections known.

PR #647, opened 2020-05-28
Enhanced tracing setup

ljedrz:
This solution improves the current logging/tracing setup by setting it up earlier, fixing the runtime log level control (it no longer resets the non-default settings) and allowing for multiple log subscribers (currently just `stdout`).

This change can be later extended with additional log subscribers including channels, websockets, log files etc.


aphelionz:
Friendly ping for a review :) Also @ljedrz looks like it needs a rebase.

matdehaast:
@kincaidoneil

koivunej:
breaking for the changes in crates/ilp-node/src/node.rs at least.

PR #654, opened 2020-10-01
Fix build on Rust 1.45.0 and up 

niklaslong:
This fixes the f64 to u64 conversion and associated bounds for the outgoing amount in the exchange rate calculation. 

Included:
- updated deps (`cargo update`)
- set type length limit so that api crate would compile 
- fixed f64 to u64 undefined behaviour as [Rust 1.45.0 defines out of range float to int conversions using `as` as a saturating conversion](https://github.com/rust-lang/rust/blob/master/RELEASES.md#language-1).
- updates CI to use Node `12.18.4` (the JS ilp-settlement-xrp package which used the ripple-binary-codec lib is no longer compatible with Node `11x`)
- CLI documentation for `account update` and `account update-settings` 
- clippy fixes
- cargo audit fixes: see `FIXME` in CI config. All offending deps have been ignored (beyond the scope of this PR). `net2` has been replace with `socket2` in `interledger` crates, though is still present in `tokio`: https://github.com/tokio-rs/tokio/issues/2677


crates/ilp-cli/src/parser.rs
```diff
@@ -151,7 +151,7 @@ fn accounts_create<'a, 'b>() -> App<'a, 'b> {
 
 fn accounts_update<'a, 'b>() -> App<'a, 'b> {
     AuthorizedSubCommand::with_name("update")
-        .about("Creates a new account on this node")
+        .about("Overwrite the account on this node")
```
ljedrz, line 154:
```suggestion
        .about("Update an account on this node")
```

crates/ilp-node/tests/redis/redis_helpers.rs
```diff
@@ -22,20 +24,35 @@ pub fn connection_info_to_string(info: ConnectionInfo) -> String {
...
+            let listener = socket.into_tcp_listener();
+            return listener.local_addr().unwrap().port();
         }
     }
 
     for _i in 0..1000 {
-        let listener = net2::TcpBuilder::new_v4().unwrap();
-        listener.reuse_address(true).unwrap();
-        if let Ok(listener) = listener.bind("127.0.0.1:0") {
-            return listener.listen(1).unwrap().local_addr().unwrap().port();
```
ljedrz, line 36: I see that `listen(1)` isn't there anymore, but since we're after the port here, I don't think it was ever needed :+1:

ljedrz (approved):
LGTM

crates/ilp-node/tests/redis/redis_helpers.rs
niklaslong: Interestingly the tests pass if I remove them all—I've added this one back in as discussed.

niklaslong:
@ljedrz CI fails (no test is actually failing, might be because circleci is having some issues?) though it's green locally. I don't have the permissions to re-run the build on this. 

In addition to the fixes discussed, I've also fixed a new 1.47 release related clippy lint.

PR #656, opened 2020-10-28
Consume messages in websockets to unlock pong responses

pradovic:
There is a potential problem with websocket paths in accounts API:
`/accounts/:username/payments/incoming` &
`/payments/incoming`.

The server is not consuming any messages from the client, and in that case it is not returning pong responses to the client. It appears that `warp` is not sending appropriate pong responses if the messages are not consumed. 

The fix proposed in this PR is adding a loop that consumes all received messages in to order enable server to return appropriate pong responses. This helps with a potential `keep-alive` from the clients that send periodic ping requests and expect an appropriate pong response.

I have used `tokio` tasks, and moved tokio as a dep because I have seen it used across the repo. If that is a problem, maybe we could consider a solution based on `future::task`?

### DIFF
--- a/crates/interledger-api/src/routes/accounts.rs
+++ b/crates/interledger-api/src/routes/accounts.rs
@@ -324,7 +324,9 @@ where
         .and(with_store.clone())
         .map(|id: Uuid, ws: warp::ws::Ws, store: S| {
             ws.on_upgrade(move |ws: warp::ws::WebSocket| {
-                notify_user(ws, id, store).map(|result| result.unwrap())
+                let (ws_tx, ws_rx) = ws.split();
+                tokio::task::spawn(notify_user(ws_tx, id, store).map(|result| result.unwrap()));
+                consume_msg_drain(ws_rx)
             })
         });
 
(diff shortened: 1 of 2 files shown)


crates/interledger-api/src/routes/accounts.rs
```diff
@@ -447,8 +451,20 @@ where
         .or(post_payments)
 }
 
+async fn consume_msg_drain(mut ws_rx: futures::stream::SplitStream<warp::ws::WebSocket>){
+    while let Some(result) = ws_rx.next().await {
+        let _ = match result {
```
ljedrz, line 456:
```suggestion
        if let Err(e) = result {
```
and the related small refactoring below

pradovic: Fixed.

aphelionz (approved):
Good enough for @ljedrz good enough for me!

aphelionz:
Waiting 24 hours for @kincaidoneil and @emschwartz et al

kincaidoneil (approved):
Concept ACK

Issue #40, opened 2019-03-08
BTP: Allow multiple connections per account


emschwartz:
The BTP Server should store a Vec of connections per account instead of just one

dora-gt:
Should also consider https://github.com/interledger-rs/interledger-rs/issues/297

Issue #636, opened 2020-03-12
BTP Closing Connection Issue


matdehaast:
If you run the node with the following `RUST_LOG="interledger=TRACE,warp" ` and connect to an account using BTP and then disconnect the client. The logging with spam indefinitely as far as I can tell.

```
2020-03-11T13:44:30.679783+00:00  DEBUG warp::filters::ws: websocket poll error: Trying to work with closed connection    
2020-03-11T13:44:30.679828+00:00  DEBUG warp::filters::ws: websocket poll error: Trying to work with closed connection    
2020-03-11T13:44:30.679873+00:00  DEBUG warp::filters::ws: websocket poll error: Trying to work with closed connection    
2020-03-11T13:44:30.679921+00:00  DEBUG warp::filters::ws: websocket poll error: Trying to work with closed connection    
2020-03-11T13:44:30.679965+00:00  DEBUG warp::filters::ws: websocket poll error: Trying to work with closed connection    
2020-03-11T13:44:30.680009+00:00  DEBUG warp::filters::ws: websocket poll error: Trying to work with closed connection    
2020-03-11T13:44:30.680054+00:00  DEBUG warp::filters::ws: websocket poll error: Trying to work with closed connection    
2020-03-11T13:44:30.680098+00:00  DEBUG warp::filters::ws: websocket poll error: Trying to work with closed connection    
2020-03-11T13:44:30.680141+00:00  DEBUG warp::filters::ws: websocket poll error: Trying to work with closed connection    
```

Some preliminary investigation seems to point to this
https://github.com/seanmonstar/warp/issues/446
https://github.com/seanmonstar/warp/issues/257

koivunej:
This probably has been fixed for some time now, as the current 629f8b52809e58141401b37548cad39152626d4c lockfile has 0.2.5 and the linked issue was [fixed in 0.2.2](https://github.com/seanmonstar/warp/commit/6c4283a1c4ff111ae4dcddf26c4b040344602b56). Commit b1fcbebd2dbbf27988091f5f3c0916847a949ff5 updated from 0.2.1 to 0.2.2.

Just out of interest, what btp client/lib/code/tool did you use to connect and disconnect @matdehaast? Would be good to have repro instructions just to document.

PR #657, opened 2020-10-29
Allow multple consumers per accountID in websocket payment subcription

pradovic:
In the current code only the last consumer that subscribes to `/accounts/:username/payments/incoming` will get notifications. The quick fix is store a vec of `UnboundedSender` per accountID instead of a single one.

The goal was to support a several clients per accountID. If we expect much more clients per accountID in the future,  more scalable solution could be considered.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -576,9 +576,9 @@ checksum = "1b980f2816d6ee8673b6517b52cb0e808a180efc92e5c19d02cdda79066703ef"
 
 [[package]]
 name = "futures"
-version = "0.3.7"
+version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95314d38584ffbfda215621d723e0a3906f032e03ae5551e650058dac83d4797"
+checksum = "9b3b0c040a1fe6529d30b3c5944b280c7f0dcb2930d2c3062bca967b602583d0"
 dependencies = [
  "futures-channel",
  "futures-core",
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
LGTM

aphelionz:
Waiting 24 hours for @kincaidoneil @emschwartz et al

kincaidoneil (approved):

pradovic:
Fix for the CircleCI problem: https://github.com/interledger-rs/interledger-rs/pull/658

aphelionz:
@kincaidoneil @emschwartz @matdehaast My understanding is that this may fix #40 and #636, is that correct?

kincaidoneil:
@aphelionz I don't think so. I believe there's a separate WS server for BTP (sending/receiving incoming ILP packets) vs this payment notification API.

Here's some of the relevant code that would need to be changed for multiple BTP clients, per account:
https://github.com/interledger-rs/interledger-rs/blob/96a699f6d042608198ac476663634a1e02b9a515/crates/interledger-btp/src/server.rs#L33-L51
https://github.com/interledger-rs/interledger-rs/blob/96a699f6d042608198ac476663634a1e02b9a515/crates/interledger-btp/src/service.rs#L49

kincaidoneil:
(Don't think that's necessary to merge this, just wanted to point it out in case you also wanted to support multiple BTP clients per account)

aphelionz:
Agree, we can tackle that in another PR or potentially upstream in `warp` / `tungstenite`

koivunej (approved):

PR #659, opened 2020-11-27
Fix interledger-stream build and test in isolation

frnsskk:
Cargo.toml in the `interledger-stream` package was missing the feature `"std"` for `base64`. This caused it to report an error when building or testing the package in isolation. This pull request introduces the minor fix to the problem.


aphelionz (approved):
Great, thanks @frnsskk!

PR #660, opened 2020-12-01
Add fields to WebSocket notifications

frnsskk:
For improved payment tracking and debugging, we'd like to add two additional fields to the WebSocket payment notifications: the Boolean value `connection_closed` and the packet sequence number `sequence`.

The `connection_closed` field will be `true` only when the last packet, which contains the `ConnectionClose` frame, has been received. As a result, in addition to the normal payment notifications reported in the WebSocket, an extra notification will be sent at the end, containing `{ "amount": 0, "connection_closed": true, ... }`.

The `sequence` field reports the sequence number of each received payment-carrying packet.

### DIFF
--- a/crates/ilp-node/tests/redis/payments_incoming.rs
+++ b/crates/ilp-node/tests/redis/payments_incoming.rs
@@ -137,7 +137,9 @@ async fn payments_incoming() {
     }
 
     // create a cross-thread collection of payment notifications
-    let pmt_notifications = Arc::new(Mutex::new(Vec::<PmtNotification>::with_capacity(3)));
+    let pmt_notifications = Arc::new(Mutex::new(
+        Vec::<(PmtNotification, PmtNotification)>::with_capacity(3),
+    ));
 
     // spawn a thread that listens in on node-wide payment notifications for node b
     let notifications = pmt_notifications.clone();
@@ -153,9 +155,12 @@ async fn payments_incoming() {
         // loop as many times as there are expected payment notifications
         for _ in 0..3 {
             let msg = payments_ws.read_message().unwrap();
-            let notification: PmtNotificationWrapper =
+            let payment: PmtNotificationWrapper =
                 serde_json::from_str(&msg.into_text().unwrap()).unwrap();
-            notifications.lock().unwrap().push(notification.Ok);
+            let msg = payments_ws.read_message().unwrap();
+            let close: PmtNotificationWrapper =
+                serde_json::from_str(&msg.into_text().unwrap()).unwrap();
+            notifications.lock().unwrap().push((payment.Ok, close.Ok));
         }
 
         payments_ws.close(None).unwrap();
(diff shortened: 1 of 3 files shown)


frnsskk:
There seems to be some problem with the `payments_incoming` test. It sometimes goes through, but usually fails. When it still used to send 3 payment notifications, all of them came through to the reader and the WebSocket connection was then closed successfully. Now that there are 6 notifications to send, the connection is most of the time reset before the last message can be read. I am currently unaware of the reason for this and how to fix it.

frnsskk:
@koivunej has introduced a fix, so that all tests now pass successfully and consistently.

crates/interledger-stream/src/server.rs
```diff
@@ -272,6 +311,10 @@ fn receive_money(
                 source_asset_scale: asset_scale,
             }));
         }
+
+        if let Frame::ConnectionClose(_) = frame {
+            connection_closed = true;
+        }
```
koivunej, line 317: This could use some comments like the above cases have. For me it sounds like we should make sure that this is the last frame in the packet, maybe? I am not up to date on the specs around this however. If you already know that more frames can come after this it could be interesting to read in the comment.

crates/ilp-node/tests/redis/payments_incoming.rs
```diff
@@ -197,23 +194,33 @@ async fn payments_incoming() {
     .await
     .unwrap();
 
-    // a small delay to make sure the notifications are intercepted in the other thread
-    thread::sleep(Duration::from_secs(1));
+    // FIXME: this should be doable with rx.collect::<????>().await.unwrap()
```
koivunej, line 197: I couldn't find a way around this but must've made some silly mistake in the process.

koivunej (commented):
I think this is looking good but I think we should add commentary on why/how the connectionclose can be handled as it is currently handled.

crates/interledger-stream/src/server.rs
frnsskk: Added clarifying comment.

crates/ilp-node/tests/redis/payments_incoming.rs
```diff
@@ -137,11 +132,8 @@ async fn payments_incoming() {
     }
```
koivunej, line 133: Just realized as we've been discussing this internally at least related to #663. The tests should have a sleep in the beginning similar to https://github.com/interledger-rs/interledger-rs/blob/98eca405495aa3b78c6eb45fd4d4dd31f9e2d1e3/crates/ilp-node/tests/redis/exchange_rates.rs#L34-L35

Otherwise as we've seen with #663 there will be there could be the rare failure as one of the nodes isn't yet ready to serve the ws connection, or the nodes haven't yet exchanged routing information.

```suggestion
    // FIXME: similar to exchange_rates::coincap and others; need to await for the nodes to be ready.
    tokio::time::delay_for(Duration::from_millis(1000)).await;
```

koivunej (approved):
I think this is starting to look good! #667 will need to go first though. I can't see any breaking changes here, just the two added fields in the payment notification json payload and the `receive_money` changes are internal to the module.

The json additions it might break in some serde_json configs, but in general fields should be able to be added.

koivunej:
The 24h grace period (starting from the latest rebase) will pass in a few hours and this will be merged, unless there are objections.

PR #663, opened 2020-12-04
Multiple payments benchmark

KristianWahlroos:
Adds two benchmarks which tracks, how long it takes for an ilp-node to receive payment through websocket, process it and respond back. One benchmark with high packet count and one benchmark with lower packet count.

```bash #
# This runs the process payment benchmarks
cargo bench -- process_payment
```

### DIFF
--- a/crates/ilp-node/benches/multiple_payments.rs
+++ b/crates/ilp-node/benches/multiple_payments.rs
@@ -1,0 +1,388 @@
+//! Multiple payments benchmark
+//! Adds four bench functions which tracks, how long it takes for an ilp-node to receive payment through websocket, process it and respond back.
+//! Half of the benchmarks are with http and half are with btp.
+//! Half of the benchmarks are with high packet count (100) and half are with single packets.
+//! Sometimes it might take a while to start the benchmarks as the route propagation start up is sometimes slow.
+//! When a request is sent it will receive message for every packet + one connection closed packet.
+//! This is why 101 and 2 are used instead of 100 and 1 for [`RECEIVED_MESSAGE_COUNT_HIGH`] and [`RECEIVED_MESSAGE_COUNT_LOW`].
+use criterion::{criterion_group, criterion_main, Criterion};
+use ilp_node::InterledgerNode;
+use serde_json::{self, json};
+use tokio::runtime::Runtime;
+use tokio::sync::mpsc::channel;
+use tungstenite::{client, handshake::client::Request};
+
+mod redis_helpers;
+mod test_helpers;
+
+use redis_helpers::*;
+use test_helpers::*;
+
+const RECEIVED_MESSAGE_COUNT_HIGH: usize = 101;
+const RECEIVED_MESSAGE_COUNT_LOW: usize = 2;
+
+/// There will be executions where routes seem to be propagated right away during init, but on most
+/// runs they require this propagation delay. It will create additional noise to results.
+/// FIXME: this is a workaround to difficult startup
+const ROUTE_BROADCAST_INTERVAL: u64 = 500;
+
+fn multiple_payments_btp(c: &mut Criterion) {
+    let mut rt = Runtime::new().unwrap();
+
+    let node_a_http = get_open_port(None);
+    let node_a_settlement = get_open_port(None);
+    let node_b_http = get_open_port(None);
+    let node_b_settlement = get_open_port(None);
+    let context = TestContext::new();
+
+    let mut connection_info1 = context.get_client_connection_info();
+    connection_info1.db = 1;
+    let mut connection_info2 = context.get_client_connection_info();
+    connection_info2.db = 2;
+
+    // accounts to be created on node a
+    let alice_on_a = json!({
+        "username": "alice_on_a",
+        "asset_code": "XYZ",
+        "asset_scale": 9,
+        "ilp_over_http_incoming_token" : "default account holder",
+        "max_packet_amount": 100,
+    });
+    let b_on_a = json!({
+        "username": "b_on_a",
+        "asset_code": "XYZ",
+        "asset_scale": 9,
+        "ilp_over_btp_url": format!("ws://localhost:{}/accounts/{}/ilp/btp", node_b_http, "a_on_b"),
+        "ilp_over_btp_outgoing_token" : "token",
+        "routing_relation": "Parent",
+    });
+
+    // accounts to be created on node b
+    let a_on_b = json!({
+        "username": "a_on_b",
+        "asset_code": "XYZ",
+        "asset_scale": 9,
+        "ilp_over_btp_incoming_token" : "token",
+        "routing_relation": "Child",
+    });
+    let bob_on_b = json!({
+        "username": "bob_on_b",
+        "asset_code": "XYZ",
+        "asset_scale": 9,
+        "ilp_over_http_incoming_token" : "default account holder",
+    });
+
+    // node a config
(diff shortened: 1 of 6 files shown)


aphelionz:
I would personally love a nice benchmarking foundation and would love to hear others thoughts on it :)

GitHub suggested @dora-gt and @emschwartz, let it be so

KristianWahlroos:
`Benchmarking process_payment (Low packet count): Warming up for 3.0000 sJan 07 00:26:45.942 ERROR interledger_router::router: No route found for request : IncomingRequest { prepare: Prepare { destination: Address("example.parent.bob_on_b.V1kI5fLTy-alvd6LFsFOUqDT"), amount: 1, expires_at: "2021-01-06T22:27:14.895+00:00", execution_condition: "cd8e7a6fd001e5c5d493a317937ec3424bafe3bd472237628669c2445cb2dcb9", data_length: 77 }, from: Account { id: 3a931422-a39d-402b-9fa1-d7f20dc259df, username: Username("a_on_b"), ilp_address: Address("example.parent.a_on_b"), asset_code: "XYZ", asset_scale: 9, max_packet_amount: 18446744073709551615, min_balance: None, ilp_over_http_url: None, ilp_over_http_incoming_token: None, ilp_over_http_outgoing_token: None, ilp_over_btp_url: None, ilp_over_btp_incoming_token: Some(SecretBytesMut([REDACTED])), ilp_over_btp_outgoing_token: None, settle_threshold: None, settle_to: None, routing_relation: Child, round_trip_time: 500, packets_per_minute_limit: None, amount_per_minute_limit: None, settlement_engine_url: None } }
Jan 07 00:26:45.942 ERROR interledger_stream::client: Send money stopped because of error: SendMoneyError("Packet was rejected with error: F02 ")
Jan 07 00:26:45.942 ERROR interledger_spsp::client: Error sending payment: SendMoneyError("Packet was rejected with error: F02 ")
Jan 07 00:26:45.942 ERROR interledger_api::routes::accounts: Error sending SPSP payment: Error sending money: 1`

Got an error message, that I am fixing

koivunej (approved):
I think this is good to go.

This will add to the duplication with the `{redis,test}_helpers.rs`. I did look into this and we can most likely split it into two unpublished dev-dependency crates in the future since the redis support is needed from a few crates, but only the ilp-node related http functions are only needed in `ilp-node`.

koivunej:
This needs to be rebased.

KristianWahlroos:
I have tried to fix `interledger_stream::crypto: Error decrypting Unspecified` error in my ilp over http benchmarks. However I noticed that this error happens in tests `three_nodes.rs` and `prometheus.rs` as well. My benchmark ilp over http version and these two tests seems to be the only occurrences that uses ilp over http instead of ilp over btp, so I am guessing that there might be something fishy going on. 🍣 The decrypting error will be fixed in an another pull request as it is not a benchmark only problem.

crates/ilp-node/benches/multiple_payments.rs
```diff
@@ -0,0 +1,379 @@
...
+    rt.block_on(async {
+        loop {
+            let response = request.try_clone().unwrap().send().await.unwrap();
+            if response.status().is_success() {
+                break;
+            }
+            tokio::time::delay_for(std::time::Duration::from_millis(100)).await;
+        }
+        for _ in 0..received_message_count {
+            receiver.recv().await.unwrap().into_text().unwrap();
```
koivunej, line 357: Why does this need to do the string conversion?

crates/ilp-node/benches/multiple_payments.rs
```diff
@@ -0,0 +1,379 @@
...
+    wait_for_propagation(&mut rt, &req_low, received_message_count_low, &mut receiver);
+    c.bench_function("process_payment_http_single_packet", |b| {
+        b.iter(|| bench_fn(&mut rt, &req_low, received_message_count_low, &mut receiver));
+    });
+    wait_for_propagation(
+        &mut rt,
+        &req_high,
+        received_message_count_high,
+        &mut receiver,
+    );
```
koivunej, line 324: Why is this needed here instead of just doing it once after node creation?

crates/ilp-node/benches/multiple_payments.rs
```diff
@@ -0,0 +1,379 @@
...
+            "slippage": 0.025 // allow up to 2.5% slippage
+        }));
+    let handle = std::thread::spawn(move || {
+        let mut payments_ws = client::connect(ws_request).unwrap().0;
+        while let Ok(message) = payments_ws.read_message() {
+            sender.try_send(message).unwrap();
+        }
+        payments_ws.close(None).unwrap();
+    });
+    wait_for_propagation(&mut rt, &req_low, received_message_count_low, &mut receiver);
```
koivunej, line 315: As there's no `route_broadcast_interval` for this cases Nodes I think this can sometimes time out (didn't test but that happened in my own tests).

This waiting should be guarded by a timeout not to get a hang.

crates/ilp-node/benches/multiple_payments.rs
```diff
@@ -0,0 +1,379 @@
...
+    // node b config
+    let node_b: InterledgerNode = serde_json::from_value(json!({
+        "ilp_address": "example.parent",
+        "default_spsp_account": "bob_on_b",
+        "admin_auth_token": "admin",
+        "database_url": connection_info_to_string(connection_info2),
+        "http_bind_address": format!("127.0.0.1:{}", node_b_http),
+        "settlement_api_bind_address": format!("127.0.0.1:{}", node_b_settlement),
+        "secret_seed": random_secret(),
+        "route_broadcast_interval": Some(200),
```
koivunej, line 92: I'd make the right-hand side a constant u64, use it consistently on all of these node configurations. It'd also be a great place to document why it is needed and what issues does it cause (noise to the measurements).

crates/ilp-node/benches/multiple_payments.rs
```diff
@@ -0,0 +1,379 @@
...
+            bench_fn(
+                &mut rt,
+                &req_high,
+                received_message_count_high,
+                &mut receiver,
+            )
+        });
+    });
+
+    drop(rt);
```
koivunej, line 185: The redis testcontext nags during exit, most likely because this drops the redis connection(s). Have you tested if dropping testcontext before this would silence it? Could be that there'd just be another nag from the ilp-node side.

koivunej (changes requested):

crates/ilp-node/benches/multiple_payments.rs
```diff
@@ -0,0 +1,386 @@
...
+    handle.join().unwrap().unwrap();
+}
+
+/// Forwards the payment notifications from the websocket into the channel
+/// When runtime (rt) is dropped, error will happen ending the loop and then the websocket
+fn payment_notification_forwarder(
+    ws_request: tungstenite::http::Request<()>,
+    mut sender: tokio::sync::mpsc::Sender<tungstenite::Message>,
+) -> Result<(), tungstenite::Error> {
+    let mut payments_ws = client::connect(ws_request).unwrap().0;
```
koivunej, line 332:
```suggestion
    let mut payments_ws = client::connect(ws_request)?.0;
```

I think this is a leftover `unwrap`.

koivunej (approved):
Minor left-over unwrap, but I think this is looking good to go.

koivunej (approved):
Oki, lets merge this after the CI finishes. This doesn't really change any behaviour but it helps with further development, so even if 24h rule applied I'm sure that has already passed.

PR #665, opened 2020-12-17
Enable multiple nodes to use same redis db

pradovic:
The main use case to cover was to enable nodes to share the same redis db instance without braking the existing functionality. The proposed solution was to use different prefix for each node that want to share the db instance. This PR is enabling use of prefix, if provided, for redis in all db calls so that this functionality can be achieved. 

I have tested the functionality by running the examples from the `/examples` folder, and by  subscribing to various nodes with ws client to check the redis pub/sub as well.
With 2 nodes started, 1st one (node_a) without a prefix and a second one (node_b) with `test` prefix like: 
```
cargo run --bin ilp-node -- --ilp_address example.node_b \                                                                
  --secret_seed 1604966725982139900555208458637022875563691455429373719368053354 \
  --admin_auth_token admin-b \
  --redis_url redis://127.0.0.1:6379/ \
  --http_bind_address 127.0.0.1:8770 \
  --settlement_api_bind_address 127.0.0.1:8771 \
  --database_prefix test
```

I get expected functionality and the db looks like this:
```
127.0.0.1:6379> keys *
 1) "test:accounts:ad2e87ba-323a-42f2-8b5a-178550366f49"
 2) "routes:current"
 3) "usernames"
 4) "test:usernames"
 5) "test:accounts:7c9ad66e-80cd-4f0c-9cfd-d0584dfd832c"
 6) "accounts:41884c6c-a63e-4c85-8dd3-df04f36cee9e"
 7) "accounts"
 8) "test:routes:current"
 9) "accounts:ae910fb7-b31a-4e8b-97ab-d2d6d81af0d4"
```

So, in order to spin up different nodes that share the same db redis instance, they should be run with a `--database_prefix` argument provided, similar to how the node_b was spawn in the example. Each node should use a unique prefix for itself in order not to intervene with a data from other nodes.

We had this use case and we tried it on our fork, seems to work fine. Thought it would be useful for the main implementation as well. Let us know what you think :)

### DIFF
--- a/crates/ilp-node/src/node.rs
+++ b/crates/ilp-node/src/node.rs
@@ -92,9 +92,6 @@ fn default_settlement_api_bind_address() -> SocketAddr {
 fn default_http_bind_address() -> SocketAddr {
     SocketAddr::from(([127, 0, 0, 1], 7770))
 }
-fn default_database_prefix() -> String {
-    "".to_string()
-}
 // We allow unreachable code on the below function because there must always be exactly one default
 // regardless of how many data sources the crate is compiled to support,
 // but we don't know which will be enabled or in which quantities or configurations.
@@ -204,7 +201,7 @@ pub struct InterledgerNode {
     )]
     pub database_url: String,
     /// Database prefix which can be used in case a db instance is shared by multiple nodes
-    #[serde(default = "default_database_prefix")]
+    #[serde(default)]
     pub database_prefix: String,
     /// IP address and port to listen for HTTP connections
     /// This is used for both the API and ILP over HTTP packets
--- a/crates/ilp-node/src/node.rs
+++ b/crates/ilp-node/src/node.rs
@@ -92,6 +92,7 @@ fn default_settlement_api_bind_address() -> SocketAddr {
 fn default_http_bind_address() -> SocketAddr {
     SocketAddr::from(([127, 0, 0, 1], 7770))
 }
+fn default_database_prefix() -> String { "".to_string() }
 // We allow unreachable code on the below function because there must always be exactly one default
 // regardless of how many data sources the crate is compiled to support,
 // but we don't know which will be enabled or in which quantities or configurations.
@@ -200,6 +201,9 @@ pub struct InterledgerNode {
         alias = "redis_url"
     )]
     pub database_url: String,
+    /// Database prefix which can be used in case a db instance is shared by multiple nodes
+    #[serde(default = "default_database_prefix")]
+    pub database_prefix: String,
     /// IP address and port to listen for HTTP connections
     /// This is used for both the API and ILP over HTTP packets
     #[serde(default = "default_http_bind_address")]
--- a/crates/ilp-node/src/node.rs
+++ b/crates/ilp-node/src/node.rs
@@ -92,7 +92,9 @@ fn default_settlement_api_bind_address() -> SocketAddr {
 fn default_http_bind_address() -> SocketAddr {
     SocketAddr::from(([127, 0, 0, 1], 7770))
 }
-fn default_database_prefix() -> String { "".to_string() }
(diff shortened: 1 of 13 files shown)


kincaidoneil:
As a note, another way to allow multiple nodes to share the same Redis instance is to use different database indices (a single Redis instance supports up to 16). You can specify this at the end of the URL passed to the `--redis_url` parameter: e.g. `redis://127.0.0.1:6379/0`, `redis://127.0.0.1:6379/1`, etc.

(Also, to partition data, [this post](https://www.mikeperham.com/2015/09/24/storing-data-with-redis/) suggests using indices or multiple Redis instances if you can instead of key namespacing.)

crates/ilp-node/src/node.rs
```diff
@@ -200,6 +203,9 @@ pub struct InterledgerNode {
         alias = "redis_url"
     )]
     pub database_url: String,
+    /// Database prefix which can be used in case a db instance is shared by multiple nodes
+    #[serde(default = "default_database_prefix")]
```
koivunej, line 207: this could be simplified into:
```suggestion
    #[serde(default)]
```
which will call `<String as std::default::Default>::default()` with one less fn, if this was not yet considered.

koivunej (commented):
Only a small nit. Was thinking of the same with the `DEFAULT_DB_PREFIX` but I guess keeping it in redis builder is more in line with the existing code.

I think we should consider parametrizing existing integration tests to run on one and multiple dbs, perhaps landing that refactoring in a preceding PR after the current PRs are in, which add test (add_tbs) or modify the existing?

pradovic:
> As a note, another way to allow multiple nodes to share the same Redis instance is to use different database indices (a single Redis instance supports up to 16). You can specify this at the end of the URL passed to the `--redis_url` parameter: e.g. `redis://127.0.0.1:6379/0`, `redis://127.0.0.1:6379/1`, etc.
> 
> (Also, to partition data, [this post](https://www.mikeperham.com/2015/09/24/storing-data-with-redis/) suggests using indices or multiple Redis instances if you can instead of key namespacing.)

Thanks. Yes, I was aware of this functionality, however we ended up needing more then 16, but do not expect a high traffic from single interleadger instance, that's why this seemed like a good approach for this particular use-case. 

I would also argue that it might be useful for testing/simulation scenarios, as it would be easy to spawn large number of instances on a single redis locally, for example. 

Anyway, if we end up accepting this change, it might be good  to change the name to `db_namespace` instead of `db_prefix` on an cli argument level :)

crates/ilp-node/src/node.rs
pradovic: I agree about the parametrizing existing integration tests. I think it would make sense to do it in a separate PR, just to have more freedom if more good testing idea apears so it can be part of the same testing-related PR. Just 2 c

pradovic: Serde thingy is done :)

koivunej:
@pradovic Just reminding, you noticed this earlier, now that #644 is finally merged, this needs to be rebased and the added script needs to be fixed to be prefixed as well.

pradovic:
> @pradovic Just reminding, you noticed this earlier, now that #644 is finally merged, this needs to be rebased and the added script needs to be fixed to be prefixed as well.

I'm on it! Thanks for tge reminder!

pradovic:
@koivunej Done, pls take a quick look just to double check if it is correct 🙏

koivunej (approved):
Looking good

koivunej:
@aphelionz should we schedule a merge of this PR? While there are in-redis alternatives to this (different databases) this as a feature is very much unintrusive, the different databases continue to be supported and there is no migration need for existing databases as long as this setting is not enabled. I think the 24h delay has already passed, and this being a non-breaking change, this should be ready to be merged already.

PR #667, opened 2021-01-14
Fix clippy and audit warnings

pradovic:
`cargo clippy` and `cargo audit` have some new warnings since the new version. This is an attempt to simply fix those.

### DIFF
--- a/crates/interledger-btp/src/packet.rs
+++ b/crates/interledger-btp/src/packet.rs
@@ -96,9 +96,9 @@ where
     let mut protocol_data = Vec::new();
 
     let num_entries = reader.read_var_uint()?;
-    let mut i = BigUint::from(0 as u32);
+    let mut i = BigUint::from(0u32);
     while i < num_entries {
-        i = i.add(BigUint::from(1 as u8)); // this is probably slow
+        i = i.add(BigUint::from(1u8)); // this is probably slow
         let protocol_name = String::from_utf8(reader.read_var_octet_string()?)?;
         let content_type = ContentType::from(reader.read_u8()?);
         let data = reader.read_var_octet_string()?;
--- a/crates/interledger-packet/src/address.rs
+++ b/crates/interledger-packet/src/address.rs
@@ -288,12 +288,8 @@ mod test_address {
         let addr1 = Address::from_str("test.alice.1234.5789").unwrap();
         let addr2 = Address::from_str("test.bob").unwrap();
         assert_ne!(addr1, addr2);
-        assert_eq!(addr1, addr1);
-        assert_eq!(addr2, addr2);
-        assert!(addr1 == addr1.clone());
-        assert!(addr1 != addr2);
+        assert_eq!(addr1, addr1.clone());
         assert!(addr1.eq(&addr1));
-        assert!(addr1.ne(&addr2));
     }
 
     #[test]
--- a/crates/interledger-settlement/src/core/types.rs
+++ b/crates/interledger-settlement/src/core/types.rs
@@ -8,6 +8,7 @@ use interledger_service::Account;
 use num_bigint::BigUint;
 use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
+use std::fmt;
 use std::ops::{Div, Mul};
 use std::str::FromStr;
 use url::Url;
@@ -192,18 +193,29 @@ pub struct ConvertDetails {
     pub to: u8,
 }
 
+#[derive(Debug)]
+pub struct ConversionError;
+
+impl fmt::Display for ConversionError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Conversion error")
+    }
+}
+
+impl std::error::Error for ConversionError {}
+
 /// Helper trait for u64 and f64 asset code conversions for amounts and rates
 pub trait Convert {
     type Item: Sized;
 
     /// Returns the scaled result, or an error if there was an overflow
-    fn normalize_scale(&self, details: ConvertDetails) -> Result<Self::Item, ()>;
+    fn normalize_scale(&self, details: ConvertDetails) -> Result<Self::Item, ConversionError>;
 }
 
 impl Convert for u64 {
     type Item = u64;
 
-    fn normalize_scale(&self, details: ConvertDetails) -> Result<Self::Item, ()> {
+    fn normalize_scale(&self, details: ConvertDetails) -> Result<Self::Item, ConversionError> {
         let scale_diff = (details.from as i8 - details.to as i8).abs() as u8;
         let scale = 10u64.pow(scale_diff.into());
         let (res, overflow) = if details.to >= details.from {
@@ -212,7 +224,7 @@ impl Convert for u64 {
             self.overflowing_div(scale)
         };
         if overflow {
-            Err(())
+            Err(ConversionError)
         } else {
             Ok(res)
         }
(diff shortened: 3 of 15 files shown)


crates/interledger-settlement/src/core/types.rs
```diff
@@ -197,13 +197,19 @@ pub trait Convert {
     type Item: Sized;
 
     /// Returns the scaled result, or an error if there was an overflow
-    fn normalize_scale(&self, details: ConvertDetails) -> Result<Self::Item, ()>;
+    fn normalize_scale(
+        &self,
+        details: ConvertDetails,
+    ) -> Result<Self::Item, Box<dyn std::error::Error + 'static>>;
```
koivunej, line 203: Yeah it looks like a public trait, but `interledger-settlement` is only 0.3 so this is much easier breaking change.

As this will be a public api going forward and regardless of anyone using it, I'd collect the common issues to an `enum ConversionError` if there are more than just the `Overflow` case? I cannot at least see one; strange that there is no underflow or the normalization returning zero.

Then again, it'd seem that this is used to `unwrap`..

koivunej (commented):
Some comments we already discussed, before your latest push

crates/interledger-packet/src/address.rs
```diff
@@ -287,9 +287,6 @@ mod test_address {
     fn test_eq() {
         let addr1 = Address::from_str("test.alice.1234.5789").unwrap();
         let addr2 = Address::from_str("test.bob").unwrap();
-        assert_ne!(addr1, addr2);
-        assert_eq!(addr1, addr1);
-        assert_eq!(addr2, addr2);
```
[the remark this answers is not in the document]
pradovic: `assert_eq!(addr1, addr1);` these are the ones that clippy is complaining about. I will move back to these then, and the allow clause to suppress warnings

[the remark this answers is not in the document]
pradovic: I changed to what you suggested but left `assert!(addr1.eq(&addr1))` just for completeness of the test and to avoid having to add `allow` clause for clippy

crates/interledger-settlement/src/core/types.rs
pradovic: I agree, let's go for en enum then. I will see if there is a few possible errors and address map them in enum. If not, I will just make a struct for an `ConversionError`

crates/interledger-btp/src/packet.rs
```diff
@@ -96,9 +96,9 @@ where
     let mut protocol_data = Vec::new();
 
     let num_entries = reader.read_var_uint()?;
-    let mut i = BigUint::from(0 as u32);
+    let mut i = BigUint::from(0_u32);
```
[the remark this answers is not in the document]
pradovic: Changed to what was suggested 👍

crates/interledger-settlement/src/core/types.rs
pradovic: There is no more specific conversion errors at the moment. In the other file `crypto.rs` there has been a case of a two different errors being handled by the same unit error `()`. I have replaced those to one single error similar to here as well, because it was also used just to unwrap, and there were no specific branches of different errors (because there was not any). I would omit adding different errors in this iterations, probably better in some future PR when/if we change these functions as well.

crates/interledger-store/src/crypto.rs
```diff
@@ -138,9 +149,9 @@ pub fn encrypt_token(encryption_key: &aead::LessSafeKey, token: &[u8]) -> BytesM
 pub fn decrypt_token(
     decryption_key: &aead::LessSafeKey,
     encrypted: &[u8],
-) -> Result<SecretBytesMut, ()> {
+) -> Result<SecretBytesMut, DecryptError> {
```
koivunej, line 152: This is a breaking change which would require the next released version to be `interledger-store` to become `2.0.0`.

As we do not have a way to keep track of breaking changes at the moment (perhaps we should create an issue), I am inclined to keep the old api, going the `#[allow(clippy::...)]`.

crates/interledger-stream/src/server.rs
```diff
@@ -75,12 +75,11 @@ impl ConnectionGenerator {
...
     /// This method returns a Result in case we want to change the internal
     /// logic in the future.
-    pub fn rederive_secret(&self, destination_account: &Address) -> Result<[u8; 32], ()> {
+    pub fn rederive_secret(&self, destination_account: &Address) -> [u8; 32] {
         let local_part = destination_account.segments().rev().next().unwrap();
         // Note this computes the HMAC with the token _encoded as UTF8_,
         // rather than decoding the base64 first.
-        let shared_secret = hmac_sha256(&self.secret_generator[..], &local_part.as_bytes()[..]);
-        Ok(shared_secret)
+        hmac_sha256(&self.secret_generator[..], &local_part.as_bytes()[..])
```
koivunej, line 82: [sadly this is another exposed public method](https://docs.rs/interledger-stream/0.4.0/interledger_stream/struct.ConnectionGenerator.html#method.rederive_secret), and `interledger-stream` is already 1.0.0. My inclination is the same `#[allow(clippy::...)]` as for the `-store` crate.

koivunej: Hmm interesting, there is no published version of interledger-stream 1.0.0 but 0.4.0; I missed that originally. I need to check more carefully than looking at the `Cargo.toml`.

koivunej: Oki, I guess we could just revert the versions back, and let this become interledger-store 0.5.0 the next time around! I had assumed since the versions were bumped to 1.0.0 that they had also been pushed to crates.io but apparently not. Well, this makes things easier.

I'll add a breaking changes tracking issue, link these three cases we know of there.

crates/interledger-store/src/crypto.rs
koivunej: See my other thread on breaking: https://github.com/interledger-rs/interledger-rs/pull/667/files#r558195143

crates/interledger-stream/src/server.rs
koivunej: Commenting on all of the breaking changes: I think we should take care to *reword* and possibly split the breaking changes. At the moment the commits are listed in #668 but it would be best if the commit messages said something about the breaking change. I usually follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) which defines using a `BREAKING CHANGE: what` in the footer of the commit message in addition to having a `!` in the commit subject. I am not trying to enforce conventional commits, just that we need a way to find these.

Of course we cannot go back in time and we need to go inspect the PRs we have put in for other breaking changes, but hopefully we can avoid it becoming a huge chore in the future.

crates/interledger-store/src/crypto.rs
koivunej: Whoops, this was outdated, see my later findings in: https://github.com/interledger-rs/interledger-rs/pull/667/files#r558208981

crates/interledger-stream/src/server.rs
pradovic: Sounds good. I think that in https://github.com/interledger-rs/interledger-rs/blob/master/docs/CONTRIBUTING.md there is a mention of that https://github.com/commitizen/cz-cli should be used, which is following conventional commits. so what you described should fit perfectly

koivunej (approved):
Changes look good, and commits look good! I haven't kept up with the circleci errors, but it looks like we must go ahead with local checks only. I checked that at least the `cargo test --workspace --all-features` works.

pradovic:
@koivunej Thanks for checking! I checked `cargo audit` and `cargo clippy` ones as well, and they appear to be correct.

koivunej:
Oki, I am thinking we should just merge this ahead, or have the 24h grace period as usual starting from now.

koivunej:
@emschwartz could you relax the merging requirements as the circleci is now defunct? Please see #669 for the next steps.

koivunej:
An alternative to getting this in would be to make disable other than the clippy for example in circleci for now and go with a lighter, fix the build by migrating to github actions in the next PR.

pradovic:
Can I merge this one without 24h warning, so we can unblock other PRs?

koivunej:
Yeah lets go ahead and merge, the 24h grace period already passed!

Issue #669, opened 2021-01-19
CircleCI plan does not cover the instance needed to run build and tests


pradovic:
CircleCI plan does not cover the instance needed to run build and tests, which is currently stopping a couple of PRs which are ready to merge.

One possible solution would be to switch to github actions. I am already on it, and will open the PR soon, so this could be a solution if accepted.

matdehaast:
@pradovic lets try get it moved to GH Actions. If it's really an issue moving then I will reach out and see if we can upgrade the account.

pradovic:
Sounds good! I already have it set up on a fork 90% and it seems to be working, so looks promising for now.

pradovic:
@matdehaast Reducing the instance size on circleCI seems to have solved the problem short-term. However, since circleCI decided to reduce the instances for a free account, I still think it still makes sense to move to github actions. I don't see any benefit in using the circleCI if we are not in the mood to upgrade the account, as github actions seem to offer enough functionality for now. Anywayz, I will open the PR with GH actions, and we can decide then if everything appears to be working good.

pradovic:
Well, I tried reducing the instance for docker updates and this is what I got:
```
Upgrade your plan to unblock build

This job has been blocked because Docker Layer Caching is not available on your current plan. Please upgrade to continue building.
Upgrade Plan
```
So, proceeding try with gh actions def :)

koivunej:
I think we could drop all of the .circleci code now, since it no longer seems to run on PRs, then close this issue?

pradovic:
Sunds good to me. I can do it today/tomorrow. Ok?

PR #670, opened 2021-01-20
Switch from circleCI to github actions

pradovic:
This PR is switching `build` and `test-md` steps from circleci to github actions. These steps are run on pushes and PRs.
The release part is deprecated since it is not suppored by the current plan, and it will be switched to gh actions in the separate PR.

Part of #669.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,46 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    env:
+      RUST_LOG: "interledger=trace"
+      RUST_BACKTRACE: "full"
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y redis-server redis-tools libssl-dev
+      - name: Install rust toolchain
+        uses: hecrj/setup-rust-action@v1.3.4
+        with:
+          rust-version: stable
+          components: clippy, rustfmt
+
+      - name: Build
+        run: cargo build --all-features --all-targets
+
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
+name: CI
+
+on: push
```
koivunej, line 3:
```suggestion
on:
  push:
    branches:
      - master
  pull_request:
    branches: '*'
```

As discussed this might help to make sure we build on PRs and post-merge.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+#
+#      - name: Audit
+#        # FIXME: Disabled:
+#        # 1. spin: is no longer actively maintained
+#        # 2. sized-chunks: no safe upgrade.
+#        # 3. net2: has been removed from crates, still present as a dep to tokio
+#        run: cargo audit --ignore RUSTSEC-2019-0031 --ignore RUSTSEC-2020-0041 --ignore RUSTSEC-2020-0016
+
+      - name: Test
+        run: timeout 10m cargo test --all --all-features
```
koivunej, line 47: Note the missing newline, might give yaml issues and good practice to enable 2newline at the end" in your editor. Some say, that a text file which doesn't end in a newline is not a text file :)

.github/workflows/ci.yml
pradovic: I will change it to suggested, makes sense! However, the build for master will probably be moved in a separate file to be along the releases, as this will all happen on push to master.

pradovic:
I was reverting some unnecessary changes that were covered by the different PRs and accidentally closed the PR. I am reoping it now as draft to continue working on github actions.

koivunej (approved):
This is looking good. As usual, I don't think we should do the 24h grace period for CI stuff.

PR #671, opened 2021-01-20
Reduce instance size and disable clippy, audit to unblock pending PRs

pradovic:
First step of  #669


koivunej:
This passing makes me wonder if the task disabling were really needed? I would not expect clippy and esp. audit to be that costly.

koivunej (approved):

koivunej:
I'd say we can merge this without any grace period to unblock other PRs.

PR #672, opened 2021-01-20
Reduce resourse class for all actions as well

pradovic:
This is an addition to #667. I did not know that `medium+` instances are also going to break, and it happened on release related tasks.


koivunej (approved):
Lets try this! Note, might need a rebase.

koivunej:
I don't think these CI changes in general need a grace period, so go ahead merging and lets see if it works.

PR #673, opened 2021-01-20
Revert "Fix clippy and audit warnings"

pradovic:


pradovic:
We will not need to do this afterall.

PR #674, opened 2021-01-25
Refactor build and test-md workflows into single file

pradovic:
Refactor `build` and `test-md` to be in a same workflow but a different jobs. No need for separate files.


.github/workflows/ci.yml
```diff
@@ -42,3 +42,31 @@ jobs:
...
+      - name: Test
+        run: |
+          scripts/run-md-test.sh '^.*$' 1
+
+      - name: 'Store artifacts'
+        uses: actions/upload-artifact@v2
+        with:
+          name: run-md-test
+          path: /tmp/run-md-test
+
```
koivunej, line 72: Do I understand correctly that this will still build the sources twice?

koivunej: I got a bit confused with this `test-md` job (same name as the now deleted workflow), but looking at the checks it'd seem that the two were executed concurrently, so there's the needless rebuilding.

pradovic: Correct, there were executed in parallel + the `test-md.sh` script is executing other scripts that are doing the `cargo build` at some point.

pradovic: As we discussed, I have moved this into a same job to reuse compiled binaries.

koivunej (approved):
For posterity, I initially suggested this to avoid the re-building in the `test-md` phase. It was attempted but since the `--all-features` is not used in examples there is still some recompilation. Given my past experiences with github actions I'd prefer to have a single efficient job than multiple as each of the jobs can always get spurious platform issues, but with luck this is faster as it is.

I don't think this ci+refactoring needs a 24h grace period, merging this right away.

pradovic:
> For posterity, I initially suggested this to avoid the re-building in the `test-md` phase. It was attempted but since the `--all-features` is not used in examples there is still some recompilation. Given my past experiences with github actions I'd prefer to have a single efficient job than multiple as each of the jobs can always get spurious platform issues, but with luck this is faster as is.
> 
> I don't think this ci+refactoring needs a 24h grace period, merging this right away.

Thanks. Ok, I will think about the best way to re-use the build step and if I find a solution I like, I will open the  PR

PR #675, opened 2021-02-01
Docker releases on gh actions

pradovic:
Docker releases that mimic the same behavior as before. Docker releases are triggered on:
- push to master
- tag on any branch with `ilp-node-|ilp-cli-` prefix

Tag is derived same as before, using the same scripts.
This is only the dry-run and will not push images to dockerhub.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -37,9 +37,14 @@ jobs:
           ILP_NODE_CLI_IMAGE_TAG=$(./.circleci/release/get_docker_image_tag.sh ilp-cli ${TAG})
           echo ::set-output name=ilp_node_cli_image_tag::${ILP_NODE_CLI_IMAGE_TAG}
 
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
 #   Build and push ilp-node in the case of push to master or tag with 'ilp-node-*'
 #   todo: this is a dry-run now, change to push = true when tested
-#   todo: change to appropriate repo
       - name: Build ilp-node
         if: startsWith(steps.get_tag.outputs.tag, 'ilp-cli-') != true
         uses: docker/build-push-action@v2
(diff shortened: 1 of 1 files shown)


koivunej:
Test failure at https://github.com/interledger-rs/interledger-rs/pull/675/checks?check_run_id=1810927961#step:5:1107

.github/workflows/docker.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Install node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 'v12.18.4'
+
+      - name: Get tags
+        id: tags
+        run: |
+          TAG=${GITHUB_REF/refs\/tags\//}
+          echo ::set-output name=tag::${TAG}
```
koivunej, line 32: I think this was deprecated already with the last round of security issues in github...?

pradovic:
> 
> 
> Test failure at https://github.com/interledger-rs/interledger-rs/pull/675/checks?check_run_id=1810927961#step:5:1107

I noticed as well, which was weird as I did not change any code in the repo. Also, some of other tests failed on a fork, so it might be that gh was slow last night. I will re-run to test what happens

pradovic:
> 
> 
> Test failure at https://github.com/interledger-rs/interledger-rs/pull/675/checks?check_run_id=1810927961#step:5:1107

Yup, now all failed tests passed here, and also on my private repos/forks. It seems to be that there were some connection errors on gh instrances last night.

.github/workflows/testdocker.yml
```diff
@@ -0,0 +1,15 @@
...
+
+jobs:
+  test-docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
```
koivunej, line 14: As we found out, part of the normal protective measures is that PRs cannot access secrets in any way. This was from  https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow which is referred from https://github.com/imjohnbo/ok-to-test#about.

koivunej:
As this PR cannot really be tested as a PR (see https://github.com/interledger-rs/interledger-rs/pull/675#discussion_r568640204), so only following the merge we can see if these work or not. I am thinking this doesn't really require a 24h period since either the docker push will work or it'll fail, but it doesn't really concern the code quality as there are no software changes. The previous `latest` image is already quite old, so in the best case this will refresh it.

koivunej (approved):

pradovic:
> As this PR cannot really be tested as a PR (see [#675 (comment)](https://github.com/interledger-rs/interledger-rs/pull/675#discussion_r568640204)), so only following the merge we can see if these work or not. I am thinking this doesn't really require a 24h period since either the docker push will work or it'll fail, but it doesn't really concern the code quality as there are no software changes. The previous `latest` image is already quite old, so in the best case this will refresh it.

Just to add a bit more clarification. This is going to be a dry-run and mergint this will test the credentials and building process. If everything goes well, another PR with just changing `push=true` would make it push to a dockerhub.

PR #676, opened 2021-02-02
Fix image tag parsing in gh actions docker script

pradovic:
Fix a image tag parsing in the case of a merge to `master` branch


koivunej (approved):
Looks good to me for another test run with `push: false`. Same as with #675 and other CI changes, no 24h period required.

PR #679, opened 2021-02-03
Ci housekepping, remove todos, etc..

pradovic:
Remove todos, and fix a minor bug with the if statement field.


koivunej (approved):
Looking good but yes the conditions are complicated. I am quite certain there are no changes after this one needed.

Same as previous CI things: I don't think this needs a 24h period, so I guess a new tag is in order?

PR #682, opened 2021-02-08
Cargo update to please audit

pradovic:

PR #683, opened 2021-02-10
Update dependencies following RUSTSEC-2021-0020

koivunej:
Take 2 of #682. Closes #682. This updates `yup-oauth2 = "4"`, which removes the need for hyper 0.12 in the dependency tree, while also updating the hyper 0.13 version. Sadly there are doesn't seem to be any tests for the google pubsub functionality so we can only hope it continues to work. Last commit updates rest of the dependencies.


pradovic (approved):
LGTM!

koivunej:
#682 already had a period of being approved before merge, so assuming this is good to merge.

PR #686, opened 2021-02-23
Fix ilp-node --version

### LINKED ISSUE #680: ilp-node --version contains the same string twice
Steps to reproduce:
 1. `cargo run -p ilp-node -- --version` outputs `ilp-node 1.0.0ilp-node 1.0.0\n` on 243c9797d

I think the string should appear only once there: `ilp-node 1.0.0`.

For debugging any issues from binary releases it would also be good if this included the commit it was built from: `ilp-node 1.0.0 (243c9797d)`.

 I wonder if this could had been broken in some `clap` update? Couldn't see it after a brief skimming over. There is also [a pending TODO](https://github.com/interledger-rs/interledger-rs/blob/243c9797d6325dede07e4f3574c90a9a605f290c/crates/ilp-node/src/main.rs#L51-L52) related to clap bug which seemed to be fixed already.

koivunej:
Fixes #680, adds tests for the existing behaviour, drops assumed defunct toml support.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1035,6 +1035,7 @@ dependencies = [
  "serde",
  "serde_json",
  "socket2",
+ "tempfile",
  "tokio 0.2.25",
  "tokio-retry",
  "tracing",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -290,7 +290,6 @@ dependencies = [
  "nom",
  "serde",
  "serde_json",
- "toml",
  "yaml-rust",
 ]
 
@@ -3301,15 +3300,6 @@ dependencies = [
  "log",
  "pin-project-lite 0.1.11",
  "tokio 0.2.25",
-]
-
-[[package]]
-name = "toml"
-version = "0.5.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a31142970826733df8241ef35dc040ef98c679ab14d7c3e54d827099b3acecaa"
-dependencies = [
- "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


koivunej:
What I gathered from the argument merging is that the clap is asked to parse args twice:

0. load environment variables (only `ILP_ADDRESS` in whatever casing) as config
1. get the config positional (last) argument, nested if it was given:
    1. if stdin is not a tty: merge from `stdin` to the config (env vars)
    2. merge config file into the config (env vars + stdin)
2. parse arguments again and merge them to config (env vars + stdin + config_file)

This does seem like quite strange:
- I'd expect stdin parsing only if the config was `-` (dash), certainly not every time a config file is presented and stdin just doesn't happen to be a tty
- perhaps it would be simpler to just make it explicit that the values are required with a single parse...?
  - clap does however make for a easier to get better reporting than the serde conversion used by `config` called via `TryFrom`

The merge operation is at `merge_args` which seems to fill in previously missing values, so I need to at minimum fix my comment about this.

koivunej:
While looking at this I felt it'd be best to write any tests for the configurability, ended up with simple cases which do test cmdline, cmdline+stdin, cmdline+file, cmdline+file+stdin. In the last commit I drop the support for toml as I couldn't figure out how to get it working in a test case (at least when a single value was given in toml).

If anyone knows how this is supposed to work I'll be happy to drop it and add the toml test cases.

pradovic (approved):
LGTM

koivunej:
Oki in the latest commits I added tests for the precedence and non-obvious bad value persisting throughout the different levels of config. While there are a lot of changes I do hope the current large set of commits would be readable/small enough to prove that no functionality was changed during them.

I don't think the current configuration way makes much sense but these should be haggled on a separate issue:
- I would have expected overwriting merge behaviour
- I was surprised of the stdin config reading

 The tests should also start to fail when defaults are changed but not all defaults are changed at the same time.

pradovic (approved):

koivunej:
As we are now few hours past the 24h, I started squashing the commits. If you'd like to find the previous unsquashed but a bit cleaned up commits, see https://github.com/koivunej/interledger-rs/tree/fix-680-pre-squash.

koivunej:
Merging once checks complete.

PR #688, opened 2021-03-12
Custom (de)serialization code cleanup

koivunej:
Assorted cleanup commits while looking into the bytes04, bytes05 issues. This is mostly idle work while thinking for the path forwards towards tokio-1.0 but publishing the PR for review. 

Among fixes is the 0x80 case is currently a panic on the packet/btp as the `0` is forwarded to `byteorder::ReadBytesExt::read_uint` which requires `1 <= n <= 8`. This was found while fuzzing which I didn't include at least yet, as for some reason the longer length cases weren't found with the naive fuzz target.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1305,7 +1305,6 @@ dependencies = [
  "bytes 0.5.6",
  "chrono",
  "criterion",
- "hex",
  "once_cell",
  "regex",
  "serde",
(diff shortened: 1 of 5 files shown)


koivunej:
Quite sure the test-md failure is transient.

koivunej:
Closing this in favor of #690.

PR #690, opened 2021-03-13
Bytes 05 upgrade

### LINKED ISSUE #626: Fully migrate the whole repo from bytes 0.4 to bytes 0.5
Currently the biggest subtask of #140 is the removal of our duplicated `bytes` dependency. `bytes` is a library for zero-copy network protocols. Most of our crates depend on it directly, and all of them depend on it indirectly via its use in interledger-packet, which is the root of our workspace dependencies. interledger-packet uses bytes for zero-copy handling of ILP packets, so it's fundamental to the library. However, when Rust stabilized async/await, the Rust async ecosystem updated to `bytes` v0.5, which has some big changes compared to `bytes` v0.4, which is what interledger-rs was originally written to use. As part of our massive async/await migration, v0.5 was used only where necessary and shims were put into place for interoperation with code that used v0.4 but that otherwise didn't need to be touched. These shims are not entirely zero-copy, so to reclaim the original intent of the design of interledger-packet it would be nice to fully migrate to v0.5.

As of this writing here are the libraries that explicitly depend on v0.4:

* ilp-node
* btp
* ccp
* ildcp
* packet
* stream

Here is a non-comprehensive overview of the incompatible changes between v0.4 and v0.5: https://github.com/tokio-rs/bytes/releases/tag/v0.5.0

Here's a WIP PR that ports interledger-packet alone to v0.5 that may be used as a base for a full migration (however, note that it's still not entirely zero-copy; see the TODO comments): https://github.com/interledger-rs/interledger-rs/pull/625/files

koivunej:
Fixes #626 as far as I can see. This builds upon #688. The OER stuff is a lot of pain and confusion but I managed to remove `btp::oer` module altogether, meaning that `packet::oer` extensions will now be used all over the workplace. There are quite the number of fuzzing failures in packet and btp this PR does not fix the next will. I did quite a bit off fuzzing guided testing to make sure not a lot of differences in errors were introduced, and panics were fixed.

This is a breaking change on the whole as public api of `interledger-packet` is modified at least by `bytes` version. Removals in `interledger-btp` didn't touch public items.

After this and the follow-up PR(s) it will be easier but not easy to continue with the `bytes = "1.0"` upgrade towards `tokio-1.0`.


koivunej:
The whole change in this PR is a breaking change but there may have been others. I could mark the commits if they are commented on in review?

koivunej:
Turned this into a draft as there are likely some serialization breaking changes already, would probably need to handle the cleanups and removals first in a separate PR and add more tests.

koivunej:
There's a bit of semi-avoidable churn in the commits as I didn't see it form beginning that `interledger-btp::oer` could be removed altogether.

This only adds the test cases found with fuzzing, not the fuzzing support itself. I've been fuzzing using roundtrip tests, which parse some bytes as a type (ilp or btp), if the parsing succeeds, it will serialize them back to bytes and make sure the bytes match original input. Some parts of this work are already included, like btp contenttype becoming roundtrippable in 5985cf4. The remaining non-roundtripping allowing features are related to reading varuints which use excessive leading bytes; those would always be written out with the minimal form but the reader side does not care if excess bytes are being used.

crates/interledger-packet/src/packet.rs
```diff
@@ -54,9 +49,9 @@ impl TryFrom<u8> for PacketType {
             12 => Ok(PacketType::Prepare),
             13 => Ok(PacketType::Fulfill),
             14 => Ok(PacketType::Reject),
-            _ => Err(ParseError::InvalidPacket(format!(
+            byte => Err(ParseError::InvalidPacket(format!(
                 "Unknown packet type: {:?}",
-                byte,
+                Some(byte),
```
koivunej, line 54: this looks weird but it retains the error message for `TryFrom<BytesMut> for Packet`

koivunej:
Audit failure awaits for #689.

crates/interledger-btp/src/packet.rs
```diff
@@ -47,6 +44,12 @@ pub enum BtpPacket {
 
 impl Serializable<BtpPacket> for BtpPacket {
     fn from_bytes(bytes: &[u8]) -> Result<BtpPacket, ParseError> {
+        if bytes.is_empty() {
```
pradovic, line 47: 👏

pradovic (commented):
Looks good to me. Great job at covering some possible panic places!

koivunej:
@pradovic could you give a new review pass? I did cut some stuff out of the tail, but also head lost some rework. This no longer has any of the fuzzing related work, but I'll PR them in right after this.

crates/interledger-stream/src/crypto.rs
```diff
@@ -133,11 +133,17 @@ fn encrypt_with_nonce(
...
+
+    // FIXME: note the next comment which includes nonce and tag but only makes sure that one of
+    // AUTH_TAG_LENGTH is present. This was implemented with bytes04 originally which didn't error
+    // for trying to split_to contents which didn't exist. When upgrading to bytes05 this became
+    // obvious because split_to now panics.
+
     // ciphertext must include at least a nonce and tag
     if ciphertext.len() < AUTH_TAG_LENGTH {
         return Err(());
     }
```
koivunej, line 146: This will also be fixed in the next PR; there is some parsing related bug which makes a lot of the test cases fail if the length check is fixed here.

koivunej: Next PR will fix this length check, but my previous comment no longer holds.

crates/interledger-btp/src/packet.rs
```diff
@@ -89,19 +92,16 @@ pub struct ProtocolData {
     pub content_type: ContentType,
     pub data: Vec<u8>,
 }
-fn read_protocol_data<T>(reader: &mut T) -> Result<Vec<ProtocolData>, ParseError>
-where
-    T: ReadOerExt,
-{
+
+fn read_protocol_data(reader: &mut &[u8]) -> Result<Vec<ProtocolData>, ParseError> {
+    // TODO: using bytes here might make sense
```
pradovic, line 97: is this `todo` on purpose for the future?

crates/interledger-packet/src/address.rs
```diff
@@ -38,34 +42,53 @@ impl FromStr for Address {
...
-        if bytes.len() > MAX_ADDRESS_LENGTH {
+        if buf.remaining() > MAX_ADDRESS_LENGTH {
             return Err(ParseError::InvalidAddress(AddressError::InvalidLength(
-                bytes.len(),
+                buf.remaining(),
             )));
         }
 
-        if ADDRESS_PATTERN.is_match(str::from_utf8(&bytes)?) {
+        // TODO: bytes1 changes bytes() to be chunk()
```
pradovic, line 58: another todo

pradovic (approved):
LGTM. Left a few comments about todo, just to remind you if it was not intentional.

crates/interledger-btp/src/packet.rs
koivunej: yes... Well, I could clean this up in the next PR, good point.

crates/interledger-packet/src/address.rs
koivunej: Yeah this is for the "nth step" ahead. I actually wrote this piece of code originally with `bytes = "1.0"` so it's here downgraded for `bytes = "0.5"`.

koivunej:
Thanks! For me this is good to merge on monday. I'll open the follow-up PR with fuzzing next.

koivunej:
Assuming enough time has passed, merging.

PR #693, opened 2021-03-19
Fuzzing and hardening

koivunej:
This builds upon #690. Some found fixes have already landed with #695. It adds on top of any fixes in earlier already:

 - feature `strict` to interledger-packet for accepting only varuints which have no leading zeroes
 - use of that `interledger-packet/strict` in btp, stream at least
 - `var_octet_bytes` can now only be usize wide, which should probably be specced lower to match reality
     - handling u64::MAX length payloads would really need a streaming approach
 - the u64::MAX case became apparent in ccp which preallocated the amount of items thus in fuzzing created huge allocations; now ccp and others mostly `Vec::new` which does have a chance of perf hit
 - fuzzing in 4 crates

`interledger-packet/strict` is useful for roundtrip fuzzing where input is first parsed and then converted back into bytes. When `interledger-packet/strict` is not enabled any additional leading zeros are accepted for varuints. Accepting them but not outputting them when converting the packet back to bytes creates a lot of false panics when fuzzing. `strict` shouldn't be enabled in any production usage because it did look like the java implementation for example didn't have such strict checks on parsed `BigInteger`s.

Most of this work came out of #680 where at updating `interledger-stream` I ran into the [bad length check](https://github.com/interledger-rs/interledger-rs/pull/690/commits/4803bc322b96916d69a21506545a77dec5223041). Fixing the length check right away failed a lot of existing test cases, which got me to do differential fuzzing between two versions. Differential fuzzing of course runs into all of the previously found issues with roundtrip fuzzing. ~I will probably drop the differential fuzzing before making this ready for review.~ Dropped already.

Per crate fuzzing targets and status:
- packet: address is good, packet roundtrip added late, immediate roundtrip failure(s), `#[ignore]` tests
- btp: most fuzzed, looking quite good
- stream: roundtrip added late, immediate roundtrip failure(s), `#[ignore]` tests
- ccp: roundtripping, might be good

~85% of added lines are caused by 4 added Cargo.lock files.~


koivunej:
Build failed with the latest checkpoint-y commit but that is ok as work continues. I've removed the btp "differential fuzzing" in 68260fe. The timestamp parsing was found quite lenient while fuzzing the btp, so that'll be the next thing but I'm ready to call this good to go with the ccp huge preallocations fixed. I'll probably complete this work by creating a few issues on:
 - more fuzzing
 - fixing the preallocation somehow in ccp and others which currently call `Vec::new()`

koivunej:
Hardening doesn't seem to great with test-md failures. Lets see if I can decipher these.

crates/interledger-btp/fuzz/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+
+prev = { package = "interledger-btp", git = "https://github.com/koivunej/interledger-rs.git", rev = "bd5386756caab72f53f3fda73219bf7362d00bdf" }
+bytes04 = { package = "bytes", version = "0.4" }
+bytes = { package = "bytes", version = "0.5" }
+hex = { version = "*" }
```
koivunej, line 17: Must've dropped some wrong commit, there shouldn't be even bytes.

crates/interledger-btp/src/packet.rs
```diff
@@ -71,18 +68,29 @@ impl Serializable<BtpPacket> for BtpPacket {
...
         match type_int {
             0 => ContentType::ApplicationOctetStream,
             1 => ContentType::TextPlainUtf8,
-            _ => ContentType::Unknown,
+            x => ContentType::Unknown(x),
+        }
+    }
+}
+
+impl From<ContentType> for u8 {
```
koivunej, line 88: Could add a test case to explain that this needs to roundtrip

crates/interledger-btp/src/packet.rs
```diff
@@ -272,6 +300,171 @@ impl Serializable<BtpError> for BtpError {
...
+        }
+
+        #[test]
+        fn fuzz_1() {
+            fails_to_parse(&[6, 0, 0, 1, 0, 1, 45]);
+        }
+
+        #[test]
+        fn fuzz_2() {
+            fails_to_parse(&[1, 1, 0, 0, 4, 4, 0]);
```
ljedrz, line 321: is this one, `fuzz_1` and the other cases w/o comments ones that were found to panic when fuzzing? it would be a good idea to write a short top-level comment indicating this so it's clear these cases aren't random

crates/interledger-btp/src/packet.rs
```diff
@@ -272,6 +300,171 @@ impl Serializable<BtpError> for BtpError {
...
+                let out = parsed.unwrap().to_bytes();
+                assert_eq!(out, lenient_output);
+            }
+        }
+
+        fn fails_to_parse(data: &[u8]) {
+            BtpPacket::from_bytes(data).unwrap_err();
+        }
+
+        #[allow(unused)]
```
koivunej, line 460: This could be removed

crates/interledger-ccp/src/packet.rs
```diff
@@ -116,14 +116,27 @@ impl RouteControlRequest {
...
+    fn try_from_data(mut data: &[u8]) -> Result<Self, ParseError> {
         let mode = Mode::try_from(data.read_u8()?)?;
         let mut last_known_routing_table_id: [u8; 16] = [0; 16];
         data.read_exact(&mut last_known_routing_table_id)?;
         let last_known_epoch = data.read_u32::<BigEndian>()?;
+
+        // TODO: see discussion for Route::try_from(&mut &[u8])
         let num_features = data.read_var_uint()?;
-        let mut features: Vec<String> = Vec::with_capacity(num_features as usize);
+        let mut features: Vec<String> = Vec::new();
```
ljedrz, line 139: shouldn't `read_var_uint` impose some protocol-level boundary? if the value is very high, filling that vector will also cause an OOM, just more slowly than with a pre-allocation

koivunej:
It would appear that the "xrp-settlement" failed. From the artifact it'd appear that everything went smooth for alice but in `node-bob-settlement-engine.log`:

```
$ less node-bob-settlement-engine.log
2021-01-29T09:58:02.213Z settlement-xrp Generated new XRP testnet account: address=rnGBkZRibgCZ6p3rVHHNVj2VmudXe3nedk secret=snscPxz9TQoDXMSN3fBekxaMk1Xt8
2021-01-29T09:58:02.723Z settlement-core Started settlement engine server
2021-01-29T09:58:12.944Z settlement-xrp Received incoming XRP payment: xrp=0.0005 account=5f3024b6-38d7-46aa-b32a-79cda329e30a txHash=19A0316F6C859266EDE0E2E4981A2AFEEB3DEE30622FF75EBF8F5AB4E1464182
2021-01-29T09:58:12.947Z settlement-core Notifying connector to credit settlement: amountToCredit=0.0005 account=5f3024b6-38d7-46aa-b32a-79cda329e30a settlementId=19A0316F6C859266EDE0E2E4981A2AFEEB3DEE30622FF75EBF8F5AB4E1464182
2021-01-29T09:58:12.955Z settlement-core Error: Connector failed to process settlement: amountToCredit=0.0005 account=5f3024b6-38d7-46aa-b32a-79cda329e30a settlementId=19A0316F6C859266EDE0E2E4981A2AFEEB3DEE30622FF75EBF8F5AB4E1464182
2021-01-29T09:58:12.955Z settlement-core Connector credited incoming settlement: leftover=0.0005 credited=0 amountToCredit=0.0005 account=5f3024b6-38d7-46aa-b32a-79cda329e30a settlementId=19A0316F6C859266EDE0E2E4981A2AFEEB3DEE30622FF75EBF8F5AB4E1464182
[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6382
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16)
[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6382
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16)
[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6382
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16)
```

Need to get these running locally.

crates/interledger-btp/src/packet.rs
koivunej: yeah that would had been a great idea in the first place. I guess I can at the moment decipher these packets mentally so I could change the names.

koivunej: Done, almost thought I had found a mistake but turned out to be ok.

crates/interledger-ccp/src/packet.rs
koivunej: Noted in the PR description.

crates/interledger-btp/src/packet.rs
koivunej: Added

koivunej:
Trying to bisect this with another branch.

koivunej:
Ok I guess the test is just flaky?

crates/interledger-btp/src/packet.rs
```diff
@@ -506,6 +506,16 @@ mod tests {
         }
     }
 
+    #[test]
+    fn content_type_roundtrips() {
+        // this is an important property for any of the datatypes, otherwise fuzzer will find the
+        // [^01] examples, which may not have any examples above.
+        for x in 0..255u8 {
```
koivunej, line 513: Well, this is obviously missing a case :)

crates/interledger-packet/Cargo.toml
```diff
@@ -7,6 +7,10 @@ license = "Apache-2.0"
 edition = "2018"
 repository = "https://github.com/interledger-rs/interledger-rs"
 
+[features]
+# used when fuzzing; accepts only roundtripping input
+strict = []
+
```
koivunej, line 13: After debugging some docker builds today it occured to me that those builds are done with `--all-features`. Need to check if that applies to all crates in the workspace, or does building `ilp-node` with `--all-features` just toggle on features of the `ilp-node`.

koivunej: This is not enabled with the default binary build at least in rust 1.51, did not check others.

Enabling this would require adding `strict` feature to `ilp-node`, which would need to enable `interledger/strict`, `strict` for `interledger`, which would again need to enable `interledger-packet/strict` and others.

koivunej:
Did the hopefully the final rebasing, made sure that the `strict` is not enabled for the default `cargo build -p ilp-node --all-features` builds -- enabling it would be in the spriti of OER notes (rfc 0030) but if someone is trying it against other implementations those might not be compatible.

crates/interledger-stream/src/lib.rs
```diff
@@ -23,6 +23,29 @@ pub use server::{
...
+
+        let other = packet::StreamPacketBuilder {
+            sequence: pkt.sequence(),
+            ilp_packet_type: pkt.ilp_packet_type(),
+            prepare_amount: pkt.prepare_amount(),
+            frames: &pkt.frames().collect::<Vec<_>>(),
+        }
+        .build();
+
+        // TODO: currently this should find a crash with two values having the same Debug string
```
pradovic, line 43: TODO

crates/interledger-btp/src/packet.rs
```diff
@@ -255,12 +288,7 @@ impl Serializable<BtpError> for BtpError {
         // TODO check that the code is only 3 chars
```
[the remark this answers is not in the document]
koivunej: this is sort of valid still, and a good question. the issue is that 3 bytes of UTF-8 is accepted but they are treated as `[u8; 3]` which would hint that you could only use ascii, so it doesn't really make sense to accept 3 bytes of utf-8. There were some fuzzing failures for this, as the ErrorCode would not roundtrip, I think.

So at least leave this, but definitely could investigate this more.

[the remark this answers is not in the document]
pradovic: It was strange for me as well.. FIXME comment indicates that just checking for the length would be enough, but that is trivial. We should check if the ascii is enough and maybe switch to ascii in the future, but I think it is out of scope for this PR

crates/interledger-stream/src/lib.rs
koivunej: Done in the bbf4b04

koivunej:
`test` failure which #701 would fix. Lets see what's the `test-md` failure.

koivunej:
I'll merge this today evening or tomorrow after re-running it to see if there's some time-specific issue which could reproduce.

koivunej:
Dropped the `Cargo.lock` files in the latest push, I cannot see any point in keeping them.

koivunej:
Test failure is "normal flaky", lets hope for a `test-md` failure instead. "Normal flaky" error was:

```
thread 'three_nodes::three_nodes' panicked at 'error binding to 127.0.0.1:40675: error creating server listener: Address already in use (os error 98)', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/warp-0.2.5/src/server.rs:213:27
```

This is because the tests do not use `:0` but instead hope racing to listen and close a port.

koivunej:
Now to passes after the "normal flaky." Lets try a third.

koivunej:
Issue has been created, seems that we'll need to find another PR to trigger the occasional `test-md` failure.

PR #695, opened 2021-03-24
Smaller cleanups or perf changes

koivunej:
This PR will:

- change all crates to use the `interledger-packet::hex::HexString` introduced in #690 instead of hex::encode which produces a String
  - hex will still be used in `ilp-node` for configuration and test support
- change all crates to use `hex_literal::hex!` or `&[u8]` for static test content instead of `Lazy<Vec<u8>>`
- use `&'static str` for the btp subprotocol names which are listed in the spec (assumed most often used)
- instead of cloning the whole connections map in clone the single needed sender

This work came out of #690 and follow-up work.


pradovic (approved):

koivunej:
Merging this as this is very much uncontroversial, and most likely non-breaking (interledger-btp::packet is not public).

PR #696, opened 2021-03-25
chore(*): clippy 1.51 lints

koivunej:
- 2021 edition lints re: panic
- 1.51 redundant_slicing
- needless_question_mark

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,6 @@
 
 All crates require Rust 2018 edition and are tested on the following channels:
 
-- `1.45.0` (minimum supported)
 - `stable`
 
 ## Connecting to the Testnet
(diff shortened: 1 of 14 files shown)


koivunej:
Celebrating the 1.51 release. These changes will hike the minimum supported rust version to 1.51 because of the removed now redundant slicing. It seems that the readme is already out of date as the minimum with our dependencies (the `sized-chunks`) uses `if` inside a const fn.

koivunej:
Added the removal of MSRV 1.45 in the latest commit. This isn't supposed to be a project wide decision, but I assume it had to be raised already for the `sized-chunks` security update. I'll create a separate issue for discussion but my commit already explains my argument.

I wouldn't want to block merging this PR on the README change as it is just updating the README to match the reality which has been for quite some time.

pradovic (approved):

koivunej:
Enough time has passed, merging.

PR #698, opened 2021-03-27
CI and test-md enhancements

koivunej:
Separated from #693 which had a tough evening yesterday with multiple consecutive failures before starting to work again. This fixes:
 - the `test-md` logs (artifacts) are now only uploaded on failure (was only on success)
 - limiting the workflow execution to only pull requests and master branch (will allow anyone contributing with spam)
 - `rm` failure is from the bogus `~/.interledger` deletion which I do not know the reason for
   - it should probably be removed altogether
   - I have no idea what was the original point nor do I consider these scripts safe to run locally because of things like these
 - refactor the "mdlp" out to a simple to use awk wrapper with help, for the next time one needs to decipher the README.mds

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,11 @@
 name: CI
 
-# limit ci building to pushes to master not to get twice the notification email from github
+# limit ci building to pushes to master not to get twice the notification email
+# from github.
+#
+# if you want to test your branch on your fork, just replace the 'branches'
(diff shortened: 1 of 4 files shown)


koivunej:
Gtihub actions seems to have a few issues at the moment.

pradovic (approved):

koivunej:
This should be good to go tomorrow.

PR #699, opened 2021-03-29
fix(docker): make .git available for the build image

### LINKED ISSUE #694: Docker build is broken
In #686 I added `vergen` which depends on `git` at build time but the `git` is not accessible during docker build, so the snapshot binary releases are not building at the moment. Either the commit information needs to be provided to the build process with other means than `git` (file, env variable) or `git` needs to be added as a dependency inside the docker. I don't really know docker that well so cannot comment if the latter would make sense. Build time dependencies would make most sense.

koivunej:
When doing #680 I didn't notice the docker build failure and have been figuring out how to fix it since. Looking more closely to the dockerfile it already uses a two container construction so I was able to identify the issue being "only parts of the working copy are copied into the build container". The fix is obvious in hindsight.

Fixes #694 but we will not know until I merge this into master, we should have a third workflow which depends on both `build` and `test-md` and attempts to build the docker images.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -65,6 +65,7 @@ jobs:
         if: startsWith(steps.tags.outputs.tag, 'ilp-cli-') != true
         uses: docker/build-push-action@v2
         with:
+          context: .
           file: ./docker/Dockerfile
           push: true
           tags: interledgerrs/testnet-bundle:${{steps.tags.outputs.ilp_node_image_tag}}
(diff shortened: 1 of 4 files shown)


koivunej:
Restarting workflow, not that it tests the changes.

pradovic (approved):

koivunej:
This is now tested, and should be good to go.

koivunej:
Should also note that the build process is currently very inefficient taking a long time and is quite difficult to test.

PR #700, opened 2021-03-31
Fix redis notifications pattern for db prefix

pradovic:
When`db_prefix` is used, messages used to be filtered in the closure, on the server, instead by the redis itself.


crates/interledger-store/src/redis/mod.rs
```diff
@@ -294,8 +294,9 @@ impl RedisStoreBuilder {
         std::thread::spawn(move || {
             #[allow(clippy::cognitive_complexity)]
             let sub_status =
-                sub_connection.psubscribe::<_, _, Vec<String>>(&["*"], move |msg| {
+                sub_connection.psubscribe::<_, _, Vec<String>>(&[format!("{}{}", &db_prefix, "*")], move |msg| {
                     let channel_name = msg.get_channel_name();
+                    println!("{}", channel_name);
```
koivunej, line 299:
```suggestion
```

crates/interledger-store/src/redis/mod.rs
```diff
@@ -294,8 +294,9 @@ impl RedisStoreBuilder {
         std::thread::spawn(move || {
             #[allow(clippy::cognitive_complexity)]
             let sub_status =
-                sub_connection.psubscribe::<_, _, Vec<String>>(&["*"], move |msg| {
+                sub_connection.psubscribe::<_, _, Vec<String>>(&[format!("{}{}", &db_prefix, "*")], move |msg| {
```
koivunej, line 297: wouldn't this subscribe to both, as in the asterisk should be removed?
```suggestion
                sub_connection.psubscribe::<_, _, Vec<String>>(&[format!("{}{}", &db_prefix)], move |msg| {
```

koivunej: I think the `*` means all publish targets https://redis.io/commands/psubscribe though it doesn't really say does it include targets which are created after subscription is started.

koivunej:
Also should there be a test case added for this...?

crates/interledger-store/src/redis/mod.rs
pradovic: What do you mean by subscribe to both? I think that pattern needs `*` if we don't match the exact string

pradovic: There is an array of patterns. We have only one. It used to be `*`, and now it is `db_prefix*`, meaning any channel with `db_prefix` prefix. Does it makes sense?

pradovic:
@koivunej Looking into a way to re-use all existing db related tests for db prefix. These are almost integration tests, so I would maybe do it in a separate PR as it would, in theory, need to test all the functionalities with and without prefix + test the combination with of several nodes over one store with different prefixes. For now it was tested manually.

crates/interledger-store/src/redis/mod.rs
koivunej: Ah, now I can see it, the `"*"` went to the format!. I thought it was going to the psubscribe, now I understand. My suggestion is bad, however now I'd suggest this:

```suggestion
                let prefix = format!("{}*", &db_prefix);
                sub_connection.psubscribe::<_, _, Vec<String>>(&[prefix], move |msg| {
```

pradovic:
@koivunej Ah, I had a feeling I did tweak some tests to include `db_prefix` as well before. Do you think these kind of tests are enough for now:crates/ilp-node/tests/redis/payments_incoming.rs?  They seem to test this logic. They did not fail before, because the logic was good, it was just filtered on the server by side-effect of handling error.

koivunej:
Added a simple test which at least reproduces the case at hand, sadly there's nothing to assert for the failure since they are only log messages.

koivunej (approved):

PR #701, opened 2021-04-01
Reduce the amount of asserts in coinbase test to make it less flaky

pradovic:
The coinbase test was flaky, and it started completely failing once coinbase suspended `XRP`. To avoid this in the future and to allow for coinbase to suspend some rates and not break our tests, I have added a simple logic to pass the test if 70% of related asserts were true.

### DIFF
--- a/scripts/run-md.sh
+++ b/scripts/run-md.sh
@@ -15,9 +15,16 @@ else
   MD_FILE=-
 fi
 
-cat "$MD_FILE" | "$(dirname $0)/parse-md.sh" > "$TMP_SCRIPT"
-bash -O expand_aliases "$TMP_SCRIPT"
+# run tcpdump in the same directory where other artifacts to be uploaded reside
+TCPDUMP_OUTPUT_FILENAME=$(echo "$MD_FILE" | sha256sum | cut -f1 -d\ ).pcap
+echo "saving packet capture for '$MD_FILE' as $TCP_DUMP_OUTPUT_FILENAME"
+sudo tcpdump -i lo -s 65535 -w "/tmp/run-md-test/${TCPDUMP_OUTPUT_FILENAME}" &
+TCPDUMP_PID=$!
 
+cat "$MD_FILE" | "$(dirname $0)/parse-md.sh" > "$TMP_SCRIPT"
+bash -x -O expand_aliases "$TMP_SCRIPT"
+
+sudo kill -2 $TCPDUMP_PID
 if [ $? -eq 0 ]; then
   rm "$TMP_SCRIPT"
   exit 0
--- a/crates/ilp-node/tests/redis/exchange_rates.rs
+++ b/crates/ilp-node/tests/redis/exchange_rates.rs
@@ -46,11 +46,17 @@ async fn coincap() {
         format!("{}", obj.get("USD").expect("Should have USD rate")).as_str(),
         "1.0"
     );
-    assert!(obj.get("EUR").is_some());
-    assert!(obj.get("JPY").is_some());
-    assert!(obj.get("BTC").is_some());
-    assert!(obj.get("ETH").is_some());
-    assert!(obj.get("XRP").is_some());
+
+    // since coinbase sometimes suspends some exchange rates we would consider the test correct if 70% of the following rates are available
+    let mut count = 0;
+    let expected_rates = ["EUR", "JPY", "BTC", "ETH", "XRP"];
+    for r in &expected_rates {
+        if obj.get(r).is_some() {
+            count += 1;
+        }
+    }
+
+    assert!(count as f32 >= 0.7 * expected_rates.len() as f32)
 }
 
 // TODO can we disable this with conditional compilation?


koivunej (approved):

koivunej:
Re-run for the workflow after a flaky test-md.

pradovic:
Re-run it several times and it failed every time. I'll try to inspect this one as well.

koivunej:
Re-ran the test case. I'd say we might want to put an `-x` option for bash here: https://github.com/interledger-rs/interledger-rs/blob/502365ed51fa6eb4240b87cc12865c354dc7d315/scripts/run-md.sh#L19

Also, the `tcpdump` idea we've chatted about for each test might work as well. I didn't implement it because the next day tests were passing. It might be best to launch from the `run-md.sh`. With capturing only `lo` traffic we should get dumps which never capture the `cargo build` time downloading of the dependencies, unless of course there's some system level caching config... In any case, if the dump was too large we'd just get failure to upload it so no worries. Also bz2 or lzma or zstd should compress the dumps nicely.

In #690 these failed for me the whole friday evening straight, but on the next day all passed. There's nothing obviously wrong but it could be that some cleanup action fails on the earlier cases and so the last two fail. Could be that the last two are more prone to failing just in general because of testnet stuff.

koivunej:
Failed again, it was 2/4 failures when I re-ran it, now 3/4. Re-running again, looking at logs.

koivunej:
Again the issue looks like before in https://github.com/interledger-rs/interledger-rs/pull/693#issuecomment-808343941, in `logs/xrp-settlement/1/node-bob-settlement-engine.log`:

```
2021-04-02T14:30:17.609Z settlement-xrp Generated new XRP testnet account: address=rnDQxEVRKFxqp3VrzgHy3bfKe4ZGDJKQGf secret=snycNRv7uQCABFKwNBZnN9mqpGbJS
2021-04-02T14:30:18.014Z settlement-core Started settlement engine server
2021-04-02T14:30:25.888Z settlement-xrp Received incoming XRP payment: xrp=0.0005 account=41248951-3a75-4368-a46a-db2dfddec394 txHash=DA33F5535B8BB26AEE1EA4A457B6EDEBC8DB1D5ED5968877DE44AFCCF8D49468
2021-04-02T14:30:25.892Z settlement-core Notifying connector to credit settlement: amountToCredit=0.0005 account=41248951-3a75-4368-a46a-db2dfddec394 settlementId=DA33F5535B8BB26AEE1EA4A457B6EDEBC8DB1D5ED5968877DE44AFCCF8D49468
2021-04-02T14:30:25.902Z settlement-core Error: Connector failed to process settlement: amountToCredit=0.0005 account=41248951-3a75-4368-a46a-db2dfddec394 settlementId=DA33F5535B8BB26AEE1EA4A457B6EDEBC8DB1D5ED5968877DE44AFCCF8D49468
2021-04-02T14:30:25.903Z settlement-core Connector credited incoming settlement: leftover=0.0005 credited=0 amountToCredit=0.0005 account=41248951-3a75-4368-a46a-db2dfddec394 settlementId=DA33F5535B8BB26AEE1EA4A457B6EDEBC8DB1D5ED5968877DE44AFCCF8D49468
[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6382
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16)
[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6382
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16)
[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6382
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16)
```

pradovic:
@koivunej I was trying to reproduce it locally, unsuccessful. As you pasted, it looks like a network problem// Hmmm, but I did comment out XRP in tests, maybe it is having problem with XRP. Did not notice it before, because on the logs i was looking it looked like there was an ETH problem. Taking a look...

koivunej:
@pradovic From the `settlement-xrp` logs it would seem that the *XRP payment* is received successfully, however it's unable to talk to bobs node. `tcpdump` would most certainly clarify this up. Also running the script in a namespace might...

koivunej:
Looking at the log above more closely, the log lines are:

`Notifying connector to credit settlement`: https://github.com/interledgerjs/settlement-core/blob/e5c394a982d6df67585a0abf6858dabbb935a258/src/index.ts#L211

`Connector failed to process settlement`: https://github.com/interledgerjs/settlement-core/blob/e5c394a982d6df67585a0abf6858dabbb935a258/src/index.ts#L219

`Connector credited incoming settlement`: https://github.com/interledgerjs/settlement-core/blob/e5c394a982d6df67585a0abf6858dabbb935a258/src/index.ts#L239

The last one is quite confusing but it happens because there's an indecipherable response. I wonder if the response is just malformed since it doesn't seem to go to any `catch` handler as if it was left out of the response? The POST request: https://github.com/interledgerjs/settlement-core/blob/e5c394a982d6df67585a0abf6858dabbb935a258/src/index.ts#L199-L208

scripts/run-md.sh
```diff
@@ -16,7 +16,8 @@ else
 fi
 
 cat "$MD_FILE" | "$(dirname $0)/parse-md.sh" > "$TMP_SCRIPT"
-bash -O expand_aliases "$TMP_SCRIPT"
+bash -x -O expand_aliases "$TMP_SCRIPT"
+sudo tcpdump -i lo0 &
```
koivunej, line 20: This here will be executed strictly after each test case, not sure what it will do but I suspect it will fail to find `lo0`. At least in my local ubuntu derivative loopback is just `lo`.

https://www.wireshark.org/docs/wsug_html_chunked/AppToolstcpdump.html has just `tcpdump -i lo -s 65535 -w <file>` and I think we should store the output under the same directory, which may have been something like `/tmp/run-md/`? Would have to look up the workflow for the artifact upload. This file's stdout should end up in the github actions log. Safe way to generate an output filename would be to `$(echo "$MD_FILE" | sha256sum | cut -f1 -d\ ).pcap` though I guess zip will surprise us by not supporting that long filenames :) Just joking, I hope it does support.

For process management the tcpdump needs to be launched into background (`&`) before the script. Afterwards we need it's PID to kill it, though this is already executed on bash so we should have some %1 shorthand for it. SIGINT should be enough to stop `tcpdump` gracefully.

koivunej: > some %1 shorthand for it

Probably better to use `TCPDUMP_PID=$!` on the line following the launch and use `"$TCPDUMP_PID` later to `kill` it. This does create a sort of race condition as we cannot know if the tcpdump started succesfully, but I guess as long as we get the commandline working it should most of the time work.

pradovic: I just lunched a dummy tcpdump in a background just to see what is going to happen.

pradovic: Ah bummer, now it worked :)

pradovic: @koivunej Sorry, I missed your first comments somehow which was the most valuable one. Done, I tried with a failed test as well and it seems to upload the tcpdump. 
NOTE: this was interesting but both `lo0` and `lo` seemd to work 😕

scripts/run-md.sh
```diff
@@ -15,9 +15,15 @@ else
...
+# run tcpdump in the same directory where other artifacts to be uploaded reside
+TCPDUMP_OUTPUT_FILENAME=$(echo "$MD_FILE" | sha256sum | cut -f1 -d\ ).pcap
+sudo tcpdump -i lo -s 65535 -w "/tmp/run-md-test/${TCPDUMP_OUTPUT_FILENAME}" &
+TCPDUMP_PID=$!
+
 cat "$MD_FILE" | "$(dirname $0)/parse-md.sh" > "$TMP_SCRIPT"
-bash -O expand_aliases "$TMP_SCRIPT"
+bash -x -O expand_aliases "$TMP_SCRIPT"
 
+sudo kill -9 $TCPDUMP_PID
```
koivunej, line 26: why is this SIGKILL instead of SIGINT? sigkill cannot be handled by the program, so we might end with some buffered output being cut from the file where as for sigint is specified in the tcpdump man page to stop it.

From https://www.tcpdump.org/manpages/tcpdump.1.html:
> Tcpdump will, if not run with the -c flag, continue capturing packets until it is interrupted by a SIGINT signal

scripts/run-md.sh
```diff
@@ -15,9 +15,15 @@ else
   MD_FILE=-
 fi
 
+# run tcpdump in the same directory where other artifacts to be uploaded reside
+TCPDUMP_OUTPUT_FILENAME=$(echo "$MD_FILE" | sha256sum | cut -f1 -d\ ).pcap
+sudo tcpdump -i lo -s 65535 -w "/tmp/run-md-test/${TCPDUMP_OUTPUT_FILENAME}" &
```
koivunej, line 20:
```suggestion
echo "saving packet capture for '$MD_FILE' as $TCP_DUMP_OUTPUT_FILENAME"
sudo tcpdump -i lo -s 65535 -w "/tmp/run-md-test/${TCPDUMP_OUTPUT_FILENAME}" &
```

scripts/run-md.sh
koivunej: > both lo0 and lo seemed to work

Are you sure about that? The script isn't executed as `bash -e` so failure to execute a single line will not stop the script. I'm also quite sure I remember seeing tcpdump reporting a "no such interface lo0" in some run, for example here: https://github.com/interledger-rs/interledger-rs/runs/2255185386?check_suite_focus=true#step:5:1729

koivunej:
Re-ran once more. I fear there's some end-of-week thing triggering this. Could be that it's just github load but should had lead with that tcpdump.

scripts/run-md.sh
pradovic: Was a bit afraid if SIGINT would not work sometimes, but I think you are right. Let's go with SIGINT

pradovic:
Can we merge this one now? I guess it is enough for coinbase flaky test + we have tcpdump for the future failures. Let's monitor and fix/change when needed in the following PRs

PR #703, opened 2021-04-06
Remove error decrypting unspecified

koivunej:
First commit has more of a "drive-by" refactoring, but the rest are related to the case. It turns out that the "Error decrypting Unspecified" is totally worthless `error!` level message, because:

 * when separating the F06 path and the F99 paths in 842268f it becomes apparent that the error message is only related to F06
 * so it can be removed in c709c94 
 * and the `send_request` impl can be simplified not to have any fallthrough in 5e98c4d

Specifically this removes two log messages:

 1. ["Error decrypting Unspecified"](https://github.com/interledger-rs/interledger-rs/blob/33055529c888f64a08fdbf12addcd23ea1bac7d3/crates/interledger-stream/src/crypto.rs#L172) (always formatted but same  string)
 2. ["Unable to parse data, rejecting Prepare packet"](https://github.com/interledger-rs/interledger-rs/blob/33055529c888f64a08fdbf12addcd23ea1bac7d3/crates/interledger-stream/src/server.rs#L272) (static string)
 
This makes handling of invalid interledger STREAM packets silent, whereas previously they had the second log message without any details. Details were probably not added in the first place, because they are a result of decrypting. However the logic of handling such invalid but decryptable STREAM packets are not changed; they will be forwarded to other layers as previously. This does sound like the wrong thing to do so perhaps this needs to be revisited.

### DIFF
--- a/crates/interledger-stream/src/crypto.rs
+++ b/crates/interledger-stream/src/crypto.rs
@@ -157,10 +157,7 @@ pub fn decrypt(shared_secret: &[u8], mut ciphertext: BytesMut) -> Result<BytesMu
             aead::Aad::from(additional_data),
             &mut ciphertext,
         )
-        .map_err(|err| {
-            // FIXME: many of the tests see this if you have logging on.
-            error!("Error decrypting {:?}", err);
-        })?
+        .map_err(|_| ())?
         .len();
     ciphertext.truncate(length);
     Ok(ciphertext)
--- a/crates/interledger-stream/src/server.rs
+++ b/crates/interledger-stream/src/server.rs
@@ -207,7 +207,7 @@ where
                 request.to.asset_scale(),
                 &request.prepare,
             );
-            return match response {
+            match response {
                 Ok(ReceiveOk { fulfill, sequence }) => {
                     self.store
                         .publish_payment_notification(PaymentNotification {
@@ -253,9 +253,10 @@ where
 
                     Err(reject)
                 }
-            };
+            }
+        } else {
+            self.next.send_request(request).await
         }
-        self.next.send_request(request).await
     }
 }
 
--- a/crates/interledger-stream/src/server.rs
+++ b/crates/interledger-stream/src/server.rs
@@ -276,18 +276,10 @@ fn receive_money(
     let condition = hash_sha256(&fulfillment);
     let is_fulfillable = condition == prepare.execution_condition();
 
-    // Parse STREAM packet
-    // TODO avoid copying data
     let prepare_amount = prepare.amount();
 
-    // Note that we are copying the Prepare packet data. This is a bad idea
-    // in cases where STREAM is used to send a significant amount of data.
-    // This implementation doesn't currently support handling the STREAM data
-    // so copying the bytes of the other STREAM frames shouldn't be a big
-    // performance hit in practice.
-    // The data is copied so that we can take the Prepare packet by
-    // reference in the case that the decryption fails and we want to pass
-    // the request on to the next service.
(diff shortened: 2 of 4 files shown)


crates/interledger-stream/src/crypto.rs
```diff
@@ -167,10 +167,7 @@ pub fn decrypt(shared_secret: &[u8], mut ciphertext: BytesMut) -> Result<BytesMu
             aead::Aad::from(additional_data),
             &mut ciphertext,
         )
-        .map_err(|err| {
-            // FIXME: many of the tests see this if you have logging on.
-            error!("Error decrypting {:?}", err);
-        })?
+        .map_err(|_| ())?
```
pradovic, line 170: maybe just reduce it to `debug` or `trace`?

pradovic (approved):
Looks good to me.

crates/interledger-stream/src/server.rs
```diff
@@ -111,12 +111,19 @@ struct ReceiveOk {
...
 /// The Err(ReceiveErr) variant of receive_money(...) return result
-struct ReceiveErr {
-    reject: Reject,
-    sequence: u64,
-    connection_closed: bool,
+#[derive(Debug)]
+enum ReceiveErr {
+    /// When decrypting fails, it means that the packet was definitely not for this service running
+    /// on this node. It should be handled by forwarding it to another handler.
+    FailedToDecrypt,
```
koivunej, line 119: This might actually incorrectly strict, there are other reasons as well. Could be that the packet was just malformed stream packet. However their handling was and will be the same.

crates/interledger-stream/src/server.rs
```diff
@@ -198,43 +205,44 @@ where
...
+                                sequence,
                                 connection_closed: true,
                             });
                     }
-                    if err.reject.code() == ErrorCode::F06_UNEXPECTED_PAYMENT {
-                        // Assume the packet isn't for us if the decryption step fails.
-                        // Note this means that if the packet data is modified in any way,
-                        // the sender will likely see an error like F02: Unavailable (this is
-                        // a bit confusing but the packet data should not be modified at all
-                        // under normal circumstances).
```
koivunej, line 230: Should have spared this comment

koivunej: Restored

crates/interledger-stream/src/crypto.rs
koivunej: It is meaningless. Time could be wasted similarly to the message regardless of the log level:
```
(TRACE|DEBUG|ERROR) interledger_stream::crypto - Error decrypting Unspecified
```

The confusion doesn't go away with lower log levels. The error meant nothing, because most likely the packet will be handled by the next level, either by forwarding it or in the case of rejecting because of no route, it will be logged by the first handler in `ilp-node`: https://github.com/interledger-rs/interledger-rs/blob/33055529c888f64a08fdbf12addcd23ea1bac7d3/crates/ilp-node/src/node.rs#L368-L388

pradovic: Makes sense, if it is logged on a different place with the same (or more) amount of info than  it is meaningless.

koivunej: Also, the `Unspecified` which was logged here, the error variant of called method https://docs.rs/ring/0.16.20/ring/aead/struct.LessSafeKey.html#method.open_in_place is intentionally unspecified error, as seen here: https://docs.rs/ring/0.16.20/ring/error/struct.Unspecified.html

koivunej:
Squashed a bit, rebased on top of the master, good to merge.

PR #706, opened 2021-04-06
Fix tcpdump output file name

pradovic:


scripts/run-md.sh
```diff
@@ -16,8 +16,9 @@ else
 fi
 
 # run tcpdump in the same directory where other artifacts to be uploaded reside
-TCPDUMP_OUTPUT_FILENAME=$(echo "$MD_FILE" | sha256sum | cut -f1 -d\ ).pcap
-echo "saving packet capture for '$MD_FILE' as $TCP_DUMP_OUTPUT_FILENAME"
+TCPDUMP_OUTPUT_FILENAME=${PWD##*/}.pcap
```
koivunej, line 19: I don't know what this does, but `readlink -e` might be better. At minimum this should be in quotes to protect against paths with spaces.

pradovic: I tried to minimize it to just the parent folder, as it appears there will be a `README.md` pre folder. Why do we need sha256 and `MD_FILE` at all in  the tcpdump output file name?

pradovic: The only thing I see missing might be printing both, I think  this is kind of overkill maybe, as if the folder is run `README.md` is the one that is run, so the filename is always the same.

koivunej: Output is now looking good. I usually use the readlink together with dirname to get the parent path. The sha256sum idea was for not caring about what the path would had been, all ways would had been mangled similarly.

scripts/run-md.sh
```diff
@@ -16,8 +16,10 @@ else
 fi
 
 # run tcpdump in the same directory where other artifacts to be uploaded reside
-TCPDUMP_OUTPUT_FILENAME=$(echo "$MD_FILE" | sha256sum | cut -f1 -d\ ).pcap
-echo "saving packet capture for '$MD_FILE' as $TCP_DUMP_OUTPUT_FILENAME"
+PARENT_DIR_NAME="${PWD##*/}"
+TCPDUMP_OUTPUT_FILENAME="$PARENT_DIR_NAME".pcap
+#TCPDUMP_OUTPUT_FILENAME=$(echo "$MD_FILE" | sha256sum | cut -f1 -d\ ).pcap
```
koivunej, line 21:
```suggestion
```

koivunej (approved):
Approved with the deletion of the commented line. No point keeping that.

Issue #705, opened 2021-04-06
Continued fuzzing and hardening


koivunej:
#693 introduced fuzzing targets to some crates and in the interest of avoiding forever lasting PRs, I didn't fix all of the found issues. Current status at the end of the PR is:

 *  packet: address is good, packet roundtrip added late, immediate roundtrip failure(s), #[ignore] tests
 *  btp: most fuzzed, looking quite good
 *  stream: roundtrip added late, immediate roundtrip failure(s), #[ignore] tests
 *  ccp: roundtripping, might be good

Other crates have not been fuzzed.

One obvious fix which isn't implemented yet is matching the `interledger-packet` datetime string against a regex before asking `chrono` to parse it in: https://github.com/interledger-rs/interledger-rs/blob/33055529c888f64a08fdbf12addcd23ea1bac7d3/crates/interledger-packet/src/packet.rs#L136-L138
This fix would be similar to what was done for the variable length btp timestamp in e6f21c9b486b8397e0d0a961a2148646ff1f987b.

Marking this as good-first-issue as with any hand-written parsing code, there are most likely issues still to be discovered even after fixing the `#[ignore]` test cases. Mentoring is available by pinging me or anyone who recently contributed to the repository.

koivunej:
> `interledger_packet::packet::fuzzed::fuzzed_0_trailing_bytes`

Just realized that there are existing test cases in `crates/interledger-packet/src/packet.rs` which would suggest that supporting additional trailer bytes is ok. The tests are inside multiple aspect testing `test_try_from` named `#[test]` fns. If this is per the RFCs then the fuzzing target should be fixed and the test removed. I can't really see why this should be supported however and there are no comments on the topic either.

koivunej:
Removed the `security` label because the above linked PRs haven't contained any security fixes, or even large mallocs, only strictness, or roundtrippability changes.

Strictness: #707, #710, #715, #716
Roundtrippability: #719, #720

The #720 highlights a possible correctness issue with the parsing of [leap seconds on every date](https://github.com/interledger-rs/interledger-rs/pull/720/files#diff-ec91192e65045112abc93b4d8dbe90acace4df9e7e4b2e8abd0ca457e721fb07R755) which should be further investigated.

koivunej:
The corpus files after minimization in `interledger-packet`:

```
$ du -h fuzz/corpus/
492K    fuzz/corpus/packet
436K    fuzz/corpus/address
[...]

PR #707, opened 2021-04-15
fix(packet): add precheck for timestamps in Prepares

whalelephant:
Cc: #705 
Check if bytes are numerical values before handing to chrono

### DIFF
--- a/crates/interledger-packet/src/packet.rs
+++ b/crates/interledger-packet/src/packet.rs
@@ -138,7 +138,11 @@ impl TryFrom<BytesMut> for Prepare {
         let mut expires_at = [0x00; 17];
         content.read_exact(&mut expires_at)?;
 
-        if !expires_at.iter().all(|e| (b'0'..=b'9').contains(e)) {
+        if !expires_at
+            .iter()
+            .map(|e| (b'0'..=b'9').contains(e))
+            .fold(true, |a, b| a & b)
+        {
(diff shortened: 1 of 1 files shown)


crates/interledger-packet/src/packet.rs
```diff
@@ -741,6 +757,33 @@ mod test_prepare {
...
+                amount: x.amount(),
+                expires_at: x.expires_at(),
+                execution_condition: x.execution_condition().try_into().unwrap(),
+                destination: x.destination(),
+                data: x.data(),
+            }
+            .build();
+
+            assert_eq!(x, built);
+            assert_eq!(BytesMut::from(x), BytesMut::from(built));
```
koivunej, line 783: I think this part could now be removed, replaced with:

```suggestion
            Prepare::try_from(prep).unwrap_err();
```

I guess for the future refactoring in #708 it might be useful to assert what does the error render as:

```suggestion
            let err = Prepare::try_from(prep).unwrap_err();
            assert_eq!("Some prefix: DateTime must be numeric", &err.to_string());
```

koivunej (approved):
Looking good!

Could do a `git rebase --signoff` onto the repos master branch?

koivunej:
This is a correctness update which will deny some patterns which were previously accepted as valid timestamps, but did not adhere to RFC and the asn1 specification. Merging slightly ahead of the 24h window.

PR #709, opened 2021-04-23
Change order of operations to reduce floating-point errors

KristianWahlroos:
Mitigates some floating-point errors in exchange_rates_service by changing order of operations. Did also some refactoring to enable better testing. Mitigates cases where exchange rate ratio and scale differences cancel each other out.


koivunej (approved):
We chatted and worked on this together off-github. The ignored test cases will be solved in follow-ups.

PR #710, opened 2021-04-26
disallow extra trailing bytes in Fulfillment packet

whalelephant:
cc: #705 

Trailing bytes with length included in the varlen octet string that do not contribute to the data field should not be allowed.

### DIFF
--- a/crates/interledger-packet/src/packet.rs
+++ b/crates/interledger-packet/src/packet.rs
@@ -313,6 +313,12 @@ impl TryFrom<BytesMut> for Fulfill {
         content.skip(FULFILLMENT_LEN)?;
         content.skip_var_octet_string()?;
 
+        if !content.is_empty() {
+            return Err(ParseError::InvalidPacket(
+                "Packet contains trailing bytes".into(),
+            ));
+        }
+
         Ok(Fulfill {
             buffer,
             content_offset,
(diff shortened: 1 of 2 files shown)


crates/interledger-packet/src/packet.rs
```diff
@@ -888,6 +856,33 @@ mod test_fulfill {
...
+            0, 93, 0, 14, 40, 40, 40,
+        ];
+
+        assert_eq!(
+            "Invalid Packet: Packet contains trailing bytes",
+            format!(
+                "{}",
+                Fulfill::try_from(BytesMut::from(with_trailing_bytes)).unwrap_err()
+            )
+        );
```
koivunej, line 885: I think this should continue to be separate test case. The existing format in this file is to do multiple things in a single test case, which I think is suboptimal.

crates/interledger-packet/src/packet.rs
```diff
@@ -562,7 +568,20 @@ fn deserialize_envelope(
         before - peek.len()
     };
 
-    let content = reader.peek_var_octet_string()?;
+    let content = reader.read_var_octet_string()?;
+
+    #[cfg(fuzzing)]
```
koivunej, line 573: I know I suggested this `cfg(fuzzing)` but we do additionally have the `strict` feature which might be useful in the future. However ok to leave this as is for now, perhaps if other crates find the need for this functionality, then it needs to be turned into feature.

```suggestion
    // TODO: consider moving behind #[cfg(feature = "strict")] for dependent crate benefit
    #[cfg(fuzzing)]
```

crates/interledger-packet/src/packet.rs
```diff
@@ -888,6 +856,33 @@ mod test_fulfill {
...
+            // fullfillment (32)
+            40, 136, 1, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 164, 255, 14, 13, 136,
+            136, 136, 1, 136, 8, 0, 238, 238, 238, 0, 1, 0, 0, 0,
+            // data var octet string length (0)
+            0,
+            // extra trailer which should not be there
+            0, 93, 0, 14, 40, 40, 40,
+        ];
+
+        assert_eq!(
```
koivunej, line 879: wouldn't this assert also be `#[cfg(fuzzing)]` and otherwise there should be an `.unwrap()`?

crates/interledger-packet/src/packet.rs
```diff
@@ -562,7 +568,21 @@ fn deserialize_envelope(
         before - peek.len()
     };
 
-    let content = reader.peek_var_octet_string()?;
+    let content = reader.read_var_octet_string()?;
```
koivunej, line 571: I wonder if a copy of the `reader` should be used instead. I guess a large number of tests should had failed if that was the case.

crates/interledger-packet/src/packet.rs
koivunej: I guess this is resolved given through the comment in the beginning of the test case.

koivunej (approved):
Looking good!

crates/interledger-packet/src/packet.rs
koivunej: Nope it was `mut reader: &[u8]` not `reader: &mut &[u8]`.

crates/interledger-packet/src/packet.rs
```diff
@@ -928,15 +955,29 @@ mod test_reject {
...
+            assert_eq!(
+                "Invalid Packet: Packet contains unaccounted for trailing bytes",
+                format!("{}", with_junk_data.unwrap_err())
+            );
+        }
+
+        #[cfg(not(feature = "strict"))]
+        {
+            let with_junk_data = with_junk_data.unwrap();
+            assert_eq!(with_junk_data.unwrap().code(), REJECT_BUILDER.code);
```
koivunej, line 973:
```suggestion
            assert_eq!(with_junk_data.code(), REJECT_BUILDER.code);
```

Extra unwrap left in. CI doesn't catch this because it doesn't try checking all crates with the default features. These tests probably only go through the `cargo test --workspace --all-targets`.

PR #713, opened 2021-04-29
fix and clean f64's normalize_scale function

KristianWahlroos:
normalize_scale had previously some conversion problems as there was a conversion from u8 to i8. This also makes one ignored test; named calculates_with_high_asset_scale; work.

### DIFF
--- a/crates/interledger-service-util/src/exchange_rates_service.rs
+++ b/crates/interledger-service-util/src/exchange_rates_service.rs
@@ -338,7 +338,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "Scale difference overflow"]
     fn calculates_with_high_asset_scale() {
         assert_eq!(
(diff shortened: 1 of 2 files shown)


koivunej (approved):
Looking good, just needs to have `cargo fmt`.

koivunej:
Hmm looks like the rebase didn't go through, it's currently missing a few merges. Update your local `master` before rebasing to this repos `master`, only then do the `git rebase --signoff master`.

Also I failed to notice first time around that `--signoff` is missing. We should really get an action to verify this automatically.

Re: conventional commit "tags":

> `test(types): refactor f64_test`
> `refactor(types): clean fn normalize_scale for f64`

The tag should also be `settlement` (or some abbrev like `sttlmnt` if it doesn't fit full) instead of `types`, which doesn't really reveal what's it about.

> `test(types): refactor f64_test`

This doesn't actually refactor the test but makes the assert more lenient so perhaps the commit should reflect this? Though, refactoring would be welcome as well to have more like a "single assert per test" way, but that's not really necessary to do now.

PR #712, opened 2021-04-29
Fix test flakyness

koivunej:
Fixes #711. Adds synchronization for the launched thread, and retry for the test case.

PR #714, opened 2021-04-30
Limit csv output in ci logs

koivunej:
This removes the csv dump functionality from `interledger-stream` which is enabled on all test runs thanks to `--all-features`. The 6e5b804 should be reverted if this dev feature is found being needed later on.

Removal of `async-trait` in `interledger-rates` is a drive-by cleanup motivated by `cargo udeps` finding it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1472,7 +1472,6 @@ dependencies = [
  "byteorder",
  "bytes 0.5.6",
  "chrono",
- "csv",
  "futures 0.3.13",
  "hex-literal",
  "interledger-errors",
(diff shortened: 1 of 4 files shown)


koivunej:
The build failure caused by https://github.com/interledger-rs/interledger-rs/pull/714/checks?check_run_id=2471770996#step:9:1683 which is #712.

pradovic (approved):
👏

koivunej:
Merging this without a waiting period because I hope/guess/assume the ci logs are only concerned, and this did make the log readings re: #712 more difficult.

Issue #704, opened 2021-04-06
Handling of invalid stream packets


koivunej:
#703 silenced two loggings which were output for decryption failures and for any parsing failure (including decryption failure). As before #703, the behaviour on invalid but successfully decrypting prepares was not changed, so they are still forwarded to next handlers.

Succeeding in decrypting the packet but failing to parse it should most lead to an immediate rejection. This should be confirmed from the [stream RFC](https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md) and fixed.

Marking this as `good-first-issue` as this code isn't too involved, and there are examples of rejection in the `interledger-stream` already. Also mentoring is available by pinging me or someone else who has been recently active.

PR #715, opened 2021-05-02
Stream fuzzing fixes

whalelephant:
cc: #705 

Focused on fuzzing `StreamPacket`

- when parsing of stream packet data into valid known frames errors, it is returned instead of ignored
- updated a fuzz test name for clarification of cause of failure
- added new test and code for removing trailer bytes for fuzzing

TODO: 
- add new test and handle Unknown frames for fuzzing

### DIFF
--- a/crates/interledger-stream/src/packet.rs
+++ b/crates/interledger-stream/src/packet.rs
@@ -888,24 +888,22 @@ mod fuzzing {
 
     #[test]
     #[cfg(feature = "strict")]
-    fn fuzzed_0_varint_length_prefix_with_leading_zeros_should_error() {
+    fn fuzzed_0_extra_trailer_bytes() {
+        // From the [RFC]'s it sounds like the at least trailer junk should be kept around,
+        // but only under the condition that they are zero-bytes and MUST be ignored in parsing.
+        // For Fuzzing, at least, we remove the extra bytes for roundtripping support.
+        //
+        // [RFC]: https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md#52-stream-packet
         #[rustfmt::skip]
         let input: &[u8] = &[
             // Version, packet type, sequence and prepare amount
-            1, 14, 3, 19, 5, 3, 1, 14,
-            // len + num frame (which is zero)
-            3, 0, 0, 0,
-            // junk since num frame is zero
-            1, 14, 3, 0, 14, 5, 17,
-        ];
+            1, 14, 1, 0, 1, 0,
+            // num frames
+            1, 0,
+            // junk data since zero frames
+            0, 31, 31, 31, 0, 96, 17];
 
-        let b = BytesMut::from(input);
-        let pkt = StreamPacket::from_decrypted(b);
-
-        assert_eq!(
-            "I/O Error: variable length prefix with leading zeros",
-            format!("{}", pkt.unwrap_err())
-        );
+        roundtrip(input);
     }
 
     #[test]
@@ -964,21 +962,22 @@ mod fuzzing {
     }
 
     #[test]
-    fn fuzzed_3_extra_trailer_bytes() {
-        // From the [RFC]'s it sounds like the at least trailer junk should be kept around,
-        // but only under the condition that they are zero-bytes and MUST be ignored in parsing.
-        // For Fuzzing, at least, we remove the extra bytes for roundtripping support.
-        //
-        // [RFC]: https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md#52-stream-packet
+    #[ignore]
+    fn fuzzed_3() {
         #[rustfmt::skip]
         let input: &[u8] = &[
             // Version, packet type, sequence and prepare amount
-            1, 14, 1, 0, 1, 0,
+            1, 14, 1, 14, 1, 14,
             // num frames
-            1, 0,
-            // junk data since zero frames
-            0, 31, 31, 31, 0, 96, 17];
+            1, 1,
+            // frame data (frame type + content)
+            1, 14, 1, 0, 1, 1, 4, 0, 255, 255, 255, 255, 255, 128, 255, 128
+            ];
 
+        // roundtrip result
+        // [1, 14, 1, 14, 1, 14, 1, 1, 1, 2, 1, 0]
+        //                                ^ ----^
+        //                                Suspect due to content prefix length
         roundtrip(input);
     }
 
--- a/crates/interledger-stream/src/packet.rs
+++ b/crates/interledger-stream/src/packet.rs
@@ -887,7 +887,6 @@ mod fuzzing {
     use bytes::{Buf, BytesMut};
 
     #[test]
-    #[cfg(feature = "strict")]
     fn fuzzed_0_extra_trailer_bytes() {
         // From the [RFC]'s it sounds like the at least trailer junk should be kept around,
         // but only under the condition that they are zero-bytes and MUST be ignored in parsing.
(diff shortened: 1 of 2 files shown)


crates/interledger-stream/src/packet.rs
```diff
@@ -173,24 +173,43 @@ impl StreamPacket {
...
+        #[cfg(feature = "strict")]
         {
+            let mut frame_iter = FrameIterator {
+                buffer: &buffer_unencrypted[frames_offset..],
+            };
+
             Ok(StreamPacket {
-                buffer_unencrypted,
+                buffer_unencrypted: frame_iter
+                    .check_num_frames_and_remove_junk(num_frames, buffer_unencrypted.clone())?,
```
koivunej, line 184: I was about to write that the called method can never modify the buffer_unencrypted here however I missed that it'll return the stripped bytes!

Ok, this is looking good to me except for the `clone`. Here it'll clone the whole packet and not generally what we'd want. It could be worked around by not having `FrameIterator` constructed at this point, but instead let a free function do it or just simply do it right here.

crates/interledger-stream/src/packet.rs
```diff
@@ -173,24 +173,43 @@ impl StreamPacket {
...
                 prepare_amount,
                 frames_offset,
             })
-        } else {
-            Err(ParseError::InvalidPacket(
-                "Incorrect number of frames or unable to parse all frames".to_string(),
-            ))
+        }
+
+        #[cfg(not(feature = "strict"))]
```
koivunej, line 192: I think you've kept the original here "just in case" but I don't see a reason for keeping it around once the `strict` version is good. Do you see a reason to keep this around?

crates/interledger-stream/src/packet.rs
```diff
@@ -299,20 +318,43 @@ impl<'a> FrameIterator<'a> {
...
+        if !self.buffer.is_empty() {
+            let _ = buffer_unencrypted.split_off(self.buffer.len() + 1);
+        }
+
+        Ok(buffer_unencrypted)
+    }
 }
 
 impl<'a> Iterator for FrameIterator<'a> {
     type Item = Frame<'a>;
```
koivunej, line 346: I think this could even be:

`type Item = Result<Frame<'a>, WhateverTheErrorIs>;`

And instead of returning a `FrameIterator` from `StreamPacket::frames()` we could return a `impl Iterator<Item = Frame<'a>>` which would be constructed from `FrameIterator { ... }.map(|res| res.expect("checked at parsing time"))`.

This would however require some state in the `FrameIterator` to return `None` after the first `Err` return value.

However this isn't something that should be handled right away, perhaps polish at the end if anything.

crates/interledger-stream/src/packet.rs
```diff
@@ -906,9 +953,63 @@ mod fuzzing {
...
+			3, 5, 14, 9,
+			// len + num frame
+			1, 3,
+			// frame type - ConnectionClose
+			1,
+			// len + frame content
+			// This content does not parse successfully as ConnectionCloseFrame
+			3, 0, 4, 20,
+			// rest of frames and junk data
+			3, 7, 14, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
koivunej, line 981: The indentation looks a bit wonky here :) but don't just remove the `#[rustfmt::skip]` but just re-indent manually.

crates/interledger-stream/src/packet.rs
```diff
@@ -906,9 +953,63 @@ mod fuzzing {
...
+        //
+        // [RFC]: https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md#52-stream-packet
+        #[rustfmt::skip]
+		let input: &[u8] = &[
+            // Version, packet type, sequence and prepare amount
+			1, 14, 1, 0, 1, 0,
+			// num frames
+			1, 0,
+			// junk data since zero frames
+			0, 31, 31, 31, 0, 96, 17];
```
koivunej, line 1007: Same indentation issue.

koivunej (commented):
Few comments and indentation nits.

crates/interledger-stream/src/packet.rs
koivunej: I see you streamlined this in the latest push.

crates/interledger-stream/src/packet.rs
koivunej: This is now outdated as well.

crates/interledger-stream/src/packet.rs
```diff
@@ -173,13 +173,26 @@ impl StreamPacket {
...
+        for _ in 0..num_frames {
+            // First byte is the frame type
+            reader.skip(1)?;
+            reader.skip_var_octet_string()?;
+        }
+
+        let junk_data_len = reader.len();
+
+        if junk_data_len > 0 {
+            let _ = buffer_unencrypted.split_off(buffer_unencrypted.len() - junk_data_len);
```
koivunej, line 186: could add a comment here with a link to the stream rfc about the junk bytes. Something like:

```suggestion
            // trailing bytes are supported for future compatibility, see
            // https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md#52-stream-packet
            let _ = buffer_unencrypted.split_off(buffer_unencrypted.len() - junk_data_len);
```

Also this could just be `buffer_unencrypted.split_off(...)`; there's no `#[must_use]` on the `split_off`.

koivunej: Though now I wonder if we should keep the trailer either way, just explicitly support it via the builder?

The rfc looks fuzzy on this as in it doesn't say that:
- successfully decrypted stream packet
- with ignored junk bytes (hopefully padded with leading zeroes)
- must be serialized again with the not understood trailing junk

However perhaps the "successfully decrypted stream packet" is a give-away that this will never be serialized again, with #704 fixed.

Perhaps supporting it through the builder is something that can be done later on if needed.

crates/interledger-stream/src/packet.rs
```diff
@@ -868,23 +884,28 @@ fn saturating_read_var_uint<'a>(reader: &mut impl BufOerExt<'a>) -> Result<u64,
...
-            1, 14, 3, 19, 5, 3, 1, 14, 3, 0, 0, 0, 0, 14, 3, 0, 14, 5, 17,
-        ]);
+    #[cfg(feature = "strict")]
+    fn fuzzed_0_varint_length_prefix_with_leading_zeros_should_error() {
+        #[rustfmt::skip]
+        let input: &[u8] = &[
+            // Version, packet type, sequence and prepare amount
+            1, 14, 3, 19, 5, 3, 1, 14,
+            // len + num frame (which is zero)
+            3, 0, 0, 0,
```
koivunej, line 897:
```suggestion
            1, 0,
```

With this it'll `roundtrip(input);` as discussed and the `#[cfg(feature = "strict")]` can be removed and the name changed.

koivunej: Aah now I see you added the case 3 for this one, my bad. I'd be inclined to only save one of these, since this has been fixed in the `interledger-packet` even though it's likely this was discovered through `interledger-stream`?

whalelephant: Yes, you are correct that this was fixed in `interledger-packet`.  I think this was kept just so that the actual error was clarified and not mixed in with trailer bytes. Will remove.

crates/interledger-stream/src/packet.rs
```diff
@@ -868,23 +884,26 @@ fn saturating_read_var_uint<'a>(reader: &mut impl BufOerExt<'a>) -> Result<u64,
 #[cfg(test)]
 mod fuzzing {
     use super::{StreamPacket, StreamPacketBuilder};
+    use bytes::{Buf, BytesMut};
 
     #[test]
-    #[ignore]
+    #[cfg(feature = "strict")]
```
koivunej, line 890:
```suggestion
```

This feature does no longer seem to be required.

koivunej (approved):
Woops, I had forgotten to approve this.

PR #716, opened 2021-05-04
Fix interledger-packet::ErrorCode too lenient parsing

koivunej:
Found this crash stashed away, which wasn't because of ErrorCode or parsing, but because of too strict fuzz_target. This PR fixes:

- allow only IA5String (ascii) error codes as is in the ASN.1 definition
- handle debug and display output for the error codes with control characters gracefully
- remove string allocation in ErrorCode's Debug impl

There's some churn related to the roundtrip test fn being added and removed, but I'll follow up with a better `fuzz/fuzz_targets/packet.rs` after this PR.

Cc: #705

### DIFF
--- a/crates/interledger-packet/src/packet.rs
+++ b/crates/interledger-packet/src/packet.rs
@@ -671,49 +671,6 @@ mod fuzzed {
             "Invalid Packet: Reject.ErrorCode was not IA5String"
         );
     }
-
-    #[allow(unused)]
-    fn roundtrip(bytes: &[u8]) {
-        use super::{FulfillBuilder, Packet::*, PrepareBuilder, RejectBuilder};
-        use std::convert::TryInto;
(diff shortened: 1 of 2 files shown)


koivunej:
Oki, a few force pushes later, I think this is ready to go :)

crates/interledger-packet/src/packet.rs
```diff
@@ -643,6 +646,33 @@ impl From<Prepare> for BytesMut {
...
 }
 
+#[cfg(test)]
+mod fuzzed {
+    use super::Packet;
+    use bytes::BytesMut;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn fuzzed_0_non_utf8_error_code() {
```
koivunej, line 656:
```suggestion
    fn fuzzed_0_non_ascii_error_code() {
```

I'll handle this in the follow-up.

pradovic (approved):

PR #717, opened 2021-05-05
ci: add testing of without and with strict feats

koivunej:
I had forgotten to do this a while ago, when introducing the `interledger-packet/strict` feature.


pradovic (approved):

koivunej:
Seeing how this is a forgotten CI related change, merging this without the 24h grace period.

PR #718, opened 2021-05-05
chore: remove CODEOWNERS for now

koivunej:
it'd seem that the original maintainers are no longer responsive for the review requests, so delete this to spam their accounts less.


matdehaast:
@koivunej agreed. you can merge this

PR #719, opened 2021-05-05
stream: Allow fuzzing for Unknown Data (ErrorCode and FrameType)

whalelephant:
cc: #705 

This PR addresses:
- ensures no inner trailing bytes in frame data
- adds roundtrip-only feature for fuzzing in the case of saturated value replacement 
- stores and return Unknown ErrorCode and FrameTypes

### DIFF
--- a/crates/interledger-stream/src/packet.rs
+++ b/crates/interledger-stream/src/packet.rs
@@ -466,6 +466,18 @@ impl From<u8> for ErrorCode {
     }
 }
 
+fn check_frame_data_prefix_length(_reader: &[u8]) -> Result<(), ParseError> {
+    // Content slice passed to the `read_contents` function should not
+    // contain extra bytes.
+    #[cfg(feature = "strict")]
+    if !_reader.is_empty() {
+        return Err(ParseError::InvalidPacket(
+            "Frame content length mismatch content".to_string(),
+        ));
+    }
+    Ok(())
+}
+
 /// Helper trait for having a common interface to read/write on Frames
 pub trait SerializableFrame<'a>: Sized {
     fn put_contents(&self, buf: &mut impl MutBufOerExt);
@@ -487,6 +499,7 @@ impl<'a> SerializableFrame<'a> for ConnectionCloseFrame<'a> {
     fn read_contents(mut reader: &'a [u8]) -> Result<Self, ParseError> {
         let code = ErrorCode::from(reader.read_u8()?);
         let message_bytes = reader.read_var_octet_string()?;
+        check_frame_data_prefix_length(reader)?;
         let message = str::from_utf8(message_bytes)?;
 
         Ok(ConnectionCloseFrame { code, message })
@@ -508,6 +521,7 @@ pub struct ConnectionNewAddressFrame {
 impl<'a> SerializableFrame<'a> for ConnectionNewAddressFrame {
     fn read_contents(mut reader: &'a [u8]) -> Result<Self, ParseError> {
         let source_account = reader.read_var_octet_string()?;
+        check_frame_data_prefix_length(reader)?;
         let source_account = Address::try_from(source_account)?;
 
         Ok(ConnectionNewAddressFrame { source_account })
@@ -543,6 +557,7 @@ impl<'a> SerializableFrame<'a> for ConnectionAssetDetailsFrame<'a> {
     fn read_contents(mut reader: &'a [u8]) -> Result<Self, ParseError> {
         let source_asset_code = str::from_utf8(reader.read_var_octet_string()?)?;
         let source_asset_scale = reader.read_u8()?;
+        check_frame_data_prefix_length(reader)?;
 
         Ok(ConnectionAssetDetailsFrame {
             source_asset_scale,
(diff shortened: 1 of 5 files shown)


koivunej:
Could you add a similar test run for the `roundtrip-only` below this: https://github.com/interledger-rs/interledger-rs/blob/f054cad6beb12c5d4bf83c5456fbfb2e28c61f65/.github/workflows/ci.yml#L70

crates/interledger-stream/src/packet.rs
```diff
@@ -438,17 +442,18 @@ impl From<u8> for FrameType {
 #[derive(Debug, PartialEq, Clone, Copy)]
 #[repr(u8)]
```
koivunej, line 443: We discussed this earlier and since you haven't deleted it I guess this was this no longer an issue? Aah so it must really mean that it's the discriminant storage. All good then!

crates/interledger-stream/src/packet.rs
```diff
@@ -875,7 +942,17 @@ impl<'a> SerializableFrame<'a> for StreamDataBlockedFrame {
 fn saturating_read_var_uint<'a>(reader: &mut impl BufOerExt<'a>) -> Result<u64, ParseError> {
     if reader.peek_var_octet_string()?.len() > 8 {
         reader.skip_var_octet_string()?;
-        Ok(u64::MAX)
+
+        if cfg!(feature = "roundtrip-only") {
+            // This is needed because the returned value u64::MAX
+            // will make rounttrip fail, i.e. ByteMut::from(packet)
```
koivunej, line 948:
```suggestion
            // will make rounttrip fail, i.e. BytesMut::from(packet)
```

crates/interledger-stream/src/packet.rs
```diff
@@ -961,23 +1031,87 @@ mod fuzzing {
...
         // [1, 14, 1, 14, 1, 14, 1, 1, 1, 2, 1, 0]
         //                                ^ ----^
         //                                Suspect due to content prefix length
-        roundtrip(input);
+        let b = BytesMut::from(input);
+        let pkt = StreamPacket::from_decrypted(b);
+
+        assert_eq!(
+            "Invalid Packet: Incorrect number of frames or unable to parse all frames",
+            format!("{}", pkt.unwrap_err())
```
koivunej, line 1064: I think this would be nicer as no formatting is really done:
```suggestion
            &pkt.unwrap_err().to_string()
```

crates/interledger-stream/src/packet.rs
```diff
@@ -961,23 +1031,87 @@ mod fuzzing {
...
+            // other frame data
+            1, 0
+        ];
+
+        let b = BytesMut::from(input);
+        let pkt = StreamPacket::from_decrypted(b);
+
+        assert_eq!(
+            "Invalid Packet: Incorrect number of frames or unable to parse all frames",
+            format!("{}", pkt.unwrap_err())
```
koivunej, line 1113: The same as https://github.com/interledger-rs/interledger-rs/pull/719/files#r626632082
```suggestion
            &pkt.unwrap_err().to_string()
```

crates/interledger-stream/src/packet.rs
```diff
@@ -1080,6 +1214,46 @@ mod serialization {
...
+                frame_type: 89,
+                content: &[1, 2, 3],
+            })],
+        }
+        .build()
+    });
+
+    static SERIALIZED_UNKNOWN_FRAME_PACKET: Lazy<BytesMut> = Lazy::new(|| {
+        BytesMut::from(
+            &vec![
```
koivunej, line 1232:
```suggestion
            &[
```
Vec is probably not needed here...

crates/interledger-stream/src/packet.rs
koivunej, line 948:
```suggestion
            // will make roundtrip fail, i.e. BytesMut::from(packet)
```
(deleted previous version of this comment just in case it shows up strange in the history)

koivunej (approved):
I think this is very much looking good to go with the few suggestions I added, and the workflow additions.

crates/interledger-stream/src/packet.rs
```diff
@@ -961,23 +1031,87 @@ mod fuzzing {
...
+            0,
+            // Here are all the extra bytes that
+            // frame data length included but not in content
+            1, 1, 4, 0, 255, 255, 255, 255, 255, 128, 255, 128
             ];
 
         // roundtrip result
         // [1, 14, 1, 14, 1, 14, 1, 1, 1, 2, 1, 0]
         //                                ^ ----^
         //                                Suspect due to content prefix length
```
koivunej, line 1058:
```suggestion
```

This comment is probably no longer valid?

crates/interledger-stream/src/packet.rs
whalelephant: otherwise I think the [default] is `isize`

[default]: https://doc.rust-lang.org/nightly/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations

koivunej:
Going ahead of merge slightly ahead of 24h as this only changes the `feature = "strict"` and adds the `feature = "roundtrip-only"` functionality.

Issue #708, opened 2021-04-15
Overuse of interledger_packet::errors::ParseError


koivunej:
The most popular case is `ParseError::InvalidPacket` which requires allocation of a String, which is most often a static string but might be result of a `format!`.

Some variants like `ParseError::WrongType` are never constructed, instead `ParseError::InvalidPacket` is used:

* https://github.com/interledger-rs/interledger-rs/blob/8820e8de7f5f14644032346fce9e58ab4c20d1fd/crates/interledger-packet/src/packet.rs#L40
* https://github.com/interledger-rs/interledger-rs/blob/8820e8de7f5f14644032346fce9e58ab4c20d1fd/crates/interledger-packet/src/packet.rs#L54-L57
* recheck at https://github.com/interledger-rs/interledger-rs/blob/8820e8de7f5f14644032346fce9e58ab4c20d1fd/crates/interledger-packet/src/packet.rs#L539-L544

Going over through all calls to `ParseError::InvalidPacket` and replacing them with actual variants might be the way forward. Since ParseError is also used from other places (ccp, stream), and in relation to #704 it might be better to separate OER errors from the `interledger-packet` errors. This coupling currently forces the #704 case to be quite invisible since when decryption fails:

1. It is reported as a String https://github.com/interledger-rs/interledger-rs/blob/8820e8de7f5f14644032346fce9e58ab4c20d1fd/crates/interledger-stream/src/packet.rs#L143-L144
2. The string gets immediatedly dropped on the next callsite https://github.com/interledger-rs/interledger-rs/blob/8820e8de7f5f14644032346fce9e58ab4c20d1fd/crates/interledger-stream/src/server.rs#L285-L286

If instead `interledger-stream` would use smaller custom enum, like below, it would allow more clear separation for "bad packet" and "failed to decrypt".

 ```rust
enum InvalidPacket {
    // this assumes that interledger_packet::oer::* functionality would
    // report EncodingError instead of ParseError
    Encoding(interledger_packet::EncodingError),
    FailedToDecrypt,
}
```

Tagging this as breaking, as at least the interledger-packet apis are public. Inside `{ccp,stream}` the errors should not leak, and changes should only affect the crate-internal API.

PR #720, opened 2021-05-06
Continued interledger-packet fuzzability

koivunej:
Cc: #705

This is a follow-up to #716 with:

- `interledger-packet/roundtrip-only` feature for denying ILP timestamps which don't roundtrip [^1]
- rejects trailing bytes on `interledger-packet/strict` for Prepare and Reject (continues #710 which did this for Fulfill) [^2]
- stylistic refactoring around unifying the style (assert_eq order, `.to_string()` calls, `if cfg!(...) { } else { }`)
- moves the fuzz target verification to `src/lib.rs` where it can be used with tests as well
- adds `PartialEq` derivation for `interledger-packet::hex::HexString` for use in `assert_eq!(HexString(...), HexString(...));`

[^1]: Unsure if this should be the default; it'd seem that the `60` as the `%S` field is accepted for every date, which might not match the idea in the RFCs. "Every date" as in best to my knowledge, 2000-05-31 was not any special day w.r.t. leap seconds.

[^2]: Changes include string changes for consistent "Unexpected {inner,outer} trailing bytes", where outer means outside the interledger-packet envelope and inner means inside the envelope.

### DIFF
--- a/crates/interledger-packet/src/packet.rs
+++ b/crates/interledger-packet/src/packet.rs
@@ -671,6 +671,88 @@ mod fuzzed {
             "Invalid Packet: Reject.ErrorCode was not IA5String"
         );
     }
+
+    #[test]
+    fn fuzzed_0_with_fixed_errorcode() {
+        #[rustfmt::skip]
+        let orig = [
+            // reject
+            14,
+            // varlen for the content
+            13,
+            // fixed error code
+            116, 119, 127,
+            // varlen address
+            6, 116, 101, 115, 116, 46, 116,
+            // varlen message
+            0,
+            // varlen data
+            0,
+            // extra byte in the end
+            42,
+        ];
+
+        roundtrip(&orig);
+    }
+
+    fn roundtrip(bytes: &[u8]) {
+        use super::{FulfillBuilder, Packet::*, PrepareBuilder, RejectBuilder};
+        use std::convert::TryInto;
+        match Packet::try_from(BytesMut::from(bytes)).unwrap() {
+            Prepare(p) => {
+                let other = PrepareBuilder {
+                    amount: p.amount(),
+                    expires_at: p.expires_at(),
+                    destination: p.destination(),
+                    execution_condition: p
+                        .execution_condition()
+                        .try_into()
+                        .expect("wrong length slice"),
+                    data: p.data(),
+                }
+                .build();
+
+                if p != other {
+                    assert_eq!(p.amount(), other.amount());
+                    assert_eq!(p.expires_at(), other.expires_at());
+                    assert_eq!(p.destination(), other.destination());
+                    assert_eq!(p.execution_condition(), other.execution_condition());
+                    assert_eq!(p.data(), other.data());
+                }
+            }
+            Fulfill(f) => {
+                let other = FulfillBuilder {
+                    fulfillment: f.fulfillment().try_into().expect("wrong length slice"),
+                    data: f.data(),
+                }
+                .build();
+
+                if f != other {
+                    assert_eq!(f.fulfillment(), other.fulfillment());
+                    assert_eq!(f.data(), other.data());
+                }
+            }
+            Reject(r) => {
+                let other = RejectBuilder {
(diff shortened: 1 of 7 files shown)


pradovic (approved):

crates/interledger-packet/src/packet.rs
```diff
@@ -135,22 +135,41 @@ impl TryFrom<BytesMut> for Prepare {
...
+            let mut roundtripped = [0u8; 17];
+            write!(
+                &mut roundtripped[..],
+                "{}",
+                DateTime::<Utc>::from(expires_at).format(INTERLEDGER_TIMESTAMP_FORMAT),
+            )
+            .unwrap();
+
+            if roundtripped != read_expires_at {
+                return Err(ParseError::InvalidPacket(
```
whalelephant, line 167: Question:
Not sure if we need to align error types / have roundtrip error. I used a [different type] in the stream crate. 

[different type]: https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-stream/src/packet.rs#L950

crates/interledger-packet/src/packet.rs
```diff
@@ -667,10 +701,71 @@ mod fuzzed {
...
         let e = Packet::try_from(BytesMut::from(&orig[..])).unwrap_err();
         assert_eq!(
+            "Invalid Packet: Reject.ErrorCode was not IA5String",
             &e.to_string(),
-            "Invalid Packet: Reject.ErrorCode was not IA5String"
         );
     }
+
+    #[test]
+    fn fuzzed_0_with_fixed_errorcode() {
```
whalelephant, line 710:
```suggestion
    fn fuzzed_1_with_fixed_errorcode() {
```

following the fuzzed tests naming convention

crates/interledger-packet/src/packet.rs
```diff
@@ -667,10 +701,71 @@ mod fuzzed {
...
+                &res.unwrap_err().to_string(),
+            );
+        } else {
+            res.unwrap();
+        }
+    }
+
+    #[test]
+    #[cfg(feature = "roundtrip-only")]
+    fn fuzzed_1_chrono_60s_rollover() {
```
whalelephant, line 743:
```suggestion
    fn fuzzed_2_chrono_60s_rollover() {
```

whalelephant (approved):
Only 1 question on error type and minor tests naming suggestions

crates/interledger-packet/src/packet.rs
koivunej: my logic was to reuse the `fuzzed_0` because this was not a new finding per-say, it's was the same the original, just the single offending errorcode byte modified as acceptable (255 => 127).

Perhaps the `1` would had been more in line!

crates/interledger-packet/src/packet.rs
koivunej: At the moment all of the code I've seen either depends on:

- just *any* error value (or "non-ok") coming from methods which return `Result<T, ParseError>`
- the string formatting (the tests, maybe some logs)

So I'd be inclined not to change this *now* but leave it for #708, since it's pretty much the same -- both will need that static string as allocated one which I'd very much prefer not to have!

It's a good point however. Perhaps we should just go ahead and add a `#[cfg(feature = "roundtrip-only")]` variant there, and maybe make the whole thing `#[non_exhaustive]`. But I'm hoping the #708 will end up with more than one single massive error type.

koivunej:
Oh yeah, new stable rust! I'll do a separate PR for the clippy errors if possible.

koivunej:
I accidentially dropped the last commit once as I failed to understand what did this mean:

```
$ git push --force-with-lease
To github.com:koivunej/interledger-rs.git
 ! [rejected]          packet_errorcode_debug_cont -> packet_errorcode_debug_cont (stale info)
error: failed to push some refs to 'git@github.com:koivunej/interledger-rs.git'
```

Maybe that error message should say that there are new commits or something :)

Last commit is now restored.

PR #722, opened 2021-05-12
fix: Clean up and decouple interledge-packet::ParseError in other crates 

whalelephant:
cc #708 

This attempts to decouple the error type `ParseError` in interledger-packet with other crates. 
The idea is to have more explicit errors that are more efficient.

### DIFF
--- a/crates/interledger-ccp/src/packet.rs
+++ b/crates/interledger-ccp/src/packet.rs
@@ -7,7 +7,7 @@ use interledger_packet::{
 };
 use once_cell::sync::Lazy;
 use std::{
-    convert::TryFrom,
+    convert::{TryFrom, TryInto},
     fmt::{self, Debug},
     io::Read,
     str::{self, FromStr},
@@ -39,50 +39,55 @@ pub static CCP_UPDATE_DESTINATION: Lazy<Address> =
     Lazy::new(|| Address::from_str("peer.route.update").unwrap());
 
 #[derive(Debug)]
-pub enum CCPPacketErr {
+pub enum CcpPacketError {
     UnexpectedMode(u8),
     PacketExpired,
     UnexpectedDestination(Address),
-    UnexpectedCondition,
+    UnexpectedCondition([u8; 32]),
     IOError,
     Utf8Conversion,
     AddresssInvalid(AddressError),
 }
 
-// TODO: better error messages
-impl fmt::Display for CCPPacketErr {
+impl fmt::Display for CcpPacketError {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            CCPPacketErr::UnexpectedMode(val) => write!(fmt, "Unexpected Mode found: {}", val),
-            CCPPacketErr::PacketExpired => write!(fmt, "Packet expired"),
-            CCPPacketErr::UnexpectedDestination(address) => {
-                write!(fmt, "Packet is not a CCP message. Destination: {}", address)
+            CcpPacketError::UnexpectedMode(val) => {
+                write!(fmt, "Invalid Packet: Unexpected Mode found: {}", val)
             }
-            CCPPacketErr::UnexpectedCondition => {
-                write!(fmt, "Packet has an unexpected execution condition",)
+            CcpPacketError::PacketExpired => write!(fmt, "Invalid Packet: Packet expired"),
+            CcpPacketError::UnexpectedDestination(address) => {
+                write!(
+                    fmt,
+                    "Invalid Packet: Packet is not a CCP message. Destination: {}",
+                    address
+                )
             }
-            CCPPacketErr::IOError => write!(fmt, "IOError"),
-            CCPPacketErr::Utf8Conversion => write!(fmt, "Unable to convert data to utf-8"),
-            CCPPacketErr::AddresssInvalid(err) => write!(fmt, "Address Invalid {:?}", err),
+            CcpPacketError::UnexpectedCondition(c) => {
+                write!(fmt, "Invalid Packet: Wrong condition: {:?}", HexString(c))
+            }
+            CcpPacketError::IOError => write!(fmt, "I/O Error"),
+            CcpPacketError::Utf8Conversion => write!(fmt, "Utf-8 Conversion Error"),
+            CcpPacketError::AddresssInvalid(err) => write!(fmt, "Address Invalid {:?}", err),
         }
     }
 }
 
-impl From<std::io::Error> for CCPPacketErr {
+impl From<std::io::Error> for CcpPacketError {
     fn from(_err: std::io::Error) -> Self {
-        CCPPacketErr::IOError
+        CcpPacketError::IOError
     }
 }
 
-impl From<std::str::Utf8Error> for CCPPacketErr {
+impl From<std::str::Utf8Error> for CcpPacketError {
     fn from(_err: std::str::Utf8Error) -> Self {
-        CCPPacketErr::Utf8Conversion
+        CcpPacketError::Utf8Conversion
     }
 }
 
-impl From<AddressError> for CCPPacketErr {
+impl From<AddressError> for CcpPacketError {
     fn from(err: AddressError) -> Self {
-        CCPPacketErr::AddresssInvalid(err)
+        CcpPacketError::AddresssInvalid(err)
     }
 }
 
@@ -97,13 +102,13 @@ pub enum Mode {
 }
 
 impl TryFrom<u8> for Mode {
-    type Error = CCPPacketErr;
+    type Error = CcpPacketError;
 
     fn try_from(val: u8) -> Result<Self, Self::Error> {
         match val {
             0 => Ok(Mode::Idle),
             1 => Ok(Mode::Sync),
-            _ => Err(CCPPacketErr::UnexpectedMode(val)),
+            _ => Err(CcpPacketError::UnexpectedMode(val)),
         }
     }
 }
@@ -134,26 +139,31 @@ impl Debug for RouteControlRequest {
 }
 
 impl TryFrom<&Prepare> for RouteControlRequest {
-    type Error = CCPPacketErr;
+    type Error = CcpPacketError;
 
     fn try_from(prepare: &Prepare) -> Result<Self, Self::Error> {
         if prepare.expires_at() < SystemTime::now() {
-            return Err(CCPPacketErr::PacketExpired);
+            return Err(CcpPacketError::PacketExpired);
         }
         RouteControlRequest::try_from_without_expiry(prepare)
     }
 }
 
 impl RouteControlRequest {
-    pub(crate) fn try_from_without_expiry(prepare: &Prepare) -> Result<Self, CCPPacketErr> {
+    pub(crate) fn try_from_without_expiry(prepare: &Prepare) -> Result<Self, CcpPacketError> {
         let destination = prepare.destination();
         if destination != *CCP_CONTROL_DESTINATION {
-            return Err(CCPPacketErr::UnexpectedDestination(destination));
+            return Err(CcpPacketError::UnexpectedDestination(destination));
         }
 
         if prepare.execution_condition() != PEER_PROTOCOL_CONDITION {
             error!("Unexpected condition: {:x?}", prepare.execution_condition());
-            return Err(CCPPacketErr::UnexpectedCondition);
+            return Err(CcpPacketError::UnexpectedCondition(
+                prepare
+                    .execution_condition()
+                    .try_into()
+                    .expect("Always return a length of 32"),
+            ));
         }
 
         Self::try_from_data(prepare.data())
@@ -168,7 +178,7 @@ impl RouteControlRequest {
         }
     }
 
-    fn try_from_data(mut data: &[u8]) -> Result<Self, CCPPacketErr> {
+    fn try_from_data(mut data: &[u8]) -> Result<Self, CcpPacketError> {
         let mode = Mode::try_from(data.read_u8()?)?;
         let mut last_known_routing_table_id: [u8; 16] = [0; 16];
         data.read_exact(&mut last_known_routing_table_id)?;
@@ -228,7 +238,7 @@ pub(crate) struct RouteProp {
 }
 
 impl TryFrom<&mut &[u8]> for RouteProp {
-    type Error = CCPPacketErr;
+    type Error = CcpPacketError;
 
     // Note this takes a mutable ref to the slice so that it advances the cursor in the original slice
     fn try_from(data: &mut &[u8]) -> Result<Self, Self::Error> {
@@ -299,7 +309,7 @@ impl Debug for Route {
 }
 
 impl TryFrom<&mut &[u8]> for Route {
-    type Error = CCPPacketErr;
+    type Error = CcpPacketError;
 
     // Note this takes a mutable ref to the slice so that it advances the cursor in the original slice
     fn try_from(data: &mut &[u8]) -> Result<Self, Self::Error> {
@@ -389,26 +399,31 @@ impl Debug for RouteUpdateRequest {
 }
 
 impl TryFrom<&Prepare> for RouteUpdateRequest {
-    type Error = CCPPacketErr;
+    type Error = CcpPacketError;
 
     fn try_from(prepare: &Prepare) -> Result<Self, Self::Error> {
         if prepare.expires_at() < SystemTime::now() {
-            return Err(CCPPacketErr::PacketExpired);
+            return Err(CcpPacketError::PacketExpired);
         }
         RouteUpdateRequest::try_from_without_expiry(prepare)
     }
 }
 
 impl RouteUpdateRequest {
-    pub(crate) fn try_from_without_expiry(prepare: &Prepare) -> Result<Self, CCPPacketErr> {
+    pub(crate) fn try_from_without_expiry(prepare: &Prepare) -> Result<Self, CcpPacketError> {
         let destination = prepare.destination();
         if destination != *CCP_UPDATE_DESTINATION {
-            return Err(CCPPacketErr::UnexpectedDestination(destination));
+            return Err(CcpPacketError::UnexpectedDestination(destination));
         }
 
         if prepare.execution_condition() != PEER_PROTOCOL_CONDITION {
             error!("Unexpected condition: {:x?}", prepare.execution_condition());
-            return Err(CCPPacketErr::UnexpectedCondition);
+            return Err(CcpPacketError::UnexpectedCondition(
+                prepare
+                    .execution_condition()
+                    .try_into()
+                    .expect("Always return a length of 32"),
+            ));
         }
 
         Self::try_from_data(prepare.data())
@@ -423,7 +438,7 @@ impl RouteUpdateRequest {
         }
     }
 
-    fn try_from_data(mut data: &[u8]) -> Result<Self, CCPPacketErr> {
+    fn try_from_data(mut data: &[u8]) -> Result<Self, CcpPacketError> {
         let mut routing_table_id: [u8; 16] = [0; 16];
         data.read_exact(&mut routing_table_id)?;
         let current_epoch_index = data.read_u32::<BigEndian>()?;
@@ -520,7 +535,7 @@ mod route_control_request {
         let result = RouteControlRequest::try_from_without_expiry(&prepare);
         assert_eq!(
             result.unwrap_err().to_string(),
-            "Packet is not a CCP message. Destination: peer.route.controk"
+            "Invalid Packet: Packet is not a CCP message. Destination: peer.route.controk"
         );
     }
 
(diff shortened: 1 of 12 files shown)


crates/interledger-packet/src/errors.rs
```diff
@@ -1,24 +1,41 @@
...
+    #[error("Input data for timestamp was not numerical")]
+    TimestampConversion,
     #[error("Invalid Address: {0}")]
     InvalidAddress(#[from] AddressError),
     #[error("Invalid Packet: {0}")]
     InvalidPacket(String),
-    #[error(transparent)]
-    OtherErr(Box<dyn std::error::Error + Send>),
+    #[error("Unexpected error from Roundtrip(Fuzzing)")]
+    Roundtrip,
```
koivunej, line 22: this could be `#[cfg(feature = "roundtrip-only")]` (if supported by thiserror) also note that the error isn't *unexpected*, it's by definition expected since that's why the feature exists in the first place.

crates/interledger-packet/src/packet.rs
```diff
@@ -164,9 +160,7 @@ impl TryFrom<BytesMut> for Prepare {
             .unwrap();
 
             if roundtripped != read_expires_at {
-                return Err(ParseError::InvalidPacket(
-                    "Non-roundtripping datetime".into(),
-                ));
+                return Err(ParseError::Roundtrip);
```
koivunej, line 163: Looking at the site of use re my earlier note on this variant; `NonRoundtrippableTimestamp` would probably work even better; that way you could you peek at the `Error` type of the crate, learn about the `roundtrip-only` feature and what kind of input are not accepted when `--features roundtrip-only` (`NonRoundtrippableTimestamp` => "some timestamps are not roundtrippable").

crates/interledger-packet/src/packet.rs
```diff
@@ -658,25 +647,21 @@ impl From<Prepare> for BytesMut {
...
 /// the [`deserialize_envelope`].
-fn ensure_no_inner_trailing_bytes(content: &[u8]) -> Result<(), ParseError> {
+fn ensure_no_inner_trailing_bytes(content: &[u8]) -> Result<(), TrailingBytesError> {
     if content.is_empty() || !cfg!(feature = "strict") {
         Ok(())
     } else {
-        Err(ParseError::InvalidPacket(
-            "Unexpected inner trailing bytes".into(),
-        ))
+        Err(TrailingBytesError::Inner)
```
koivunej, line 654: I think this is looking really good! The correspondance of the deleted string message to the new value is clear!

crates/interledger-packet/src/packet.rs
```diff
@@ -700,10 +685,7 @@ mod fuzzed {
         ];
 
         let e = Packet::try_from(BytesMut::from(&orig[..])).unwrap_err();
-        assert_eq!(
-            "Invalid Packet: Reject.ErrorCode was not IA5String",
-            &e.to_string(),
-        );
+        assert_eq!("Input data is not IA5String characters", &e.to_string());
```
koivunej, line 688: This does make me think of the expected compatibility... I think as this crate is largely part of the API and these might even be formatted into rejects (or other protocol level things, or at the http; didn't check), did you consider keeping the prefix?

I fumbled the line selection so cannot suggest; but I'd remove the word "characters" (as in general, that term is very difficult with encodings, even though it's clear with ascii and similar).

Also, is there point in making this a general error instead of keeping it tied to the specific case where this is returned (namely `Reject.ErrorCode`).

crates/interledger-packet/src/packet.rs
```diff
@@ -729,10 +711,7 @@ mod fuzzed {
         let res = Packet::try_from(BytesMut::from(&orig[..]));
 
         if cfg!(feature = "strict") {
-            assert_eq!(
-                "Invalid Packet: Unexpected inner trailing bytes",
-                &res.unwrap_err().to_string(),
-            );
+            assert_eq!("Trailing Bytes Error: Inner", &res.unwrap_err().to_string(),);
```
koivunej, line 714: I think the message could be better, but I guess with `thiserror` that flexibility is lost? `match`'ing through the both variants with in a plain rust `std::fmt::Display` would be easy.

crates/interledger-packet/src/packet.rs
```diff
@@ -783,7 +762,7 @@ mod test_packet_type {
     #[test]
     fn try_from_empty() {
         assert_eq!(
-            "Invalid Packet: Unknown packet type: None",
+            "PacketType data not found",
```
koivunej, line 765:
```suggestion
            "PacketType not found",
```
String-nit: I don't think the word `data` is meaningful here; when parsing we cannot hope to find anything but data.

crates/interledger-stream/src/error.rs
```diff
@@ -1,12 +1,35 @@
...
-    PollError(String),
-    #[error("Error polling: {0}")]
-    SendMoneyError(String),
-    #[error("Error maximum time exceeded: {0}")]
-    TimeoutError(String),
+    #[error("Terminating payment since too many packets are rejected ({0} packets fulfilled, {1} packets rejected)")]
+    PaymentFailFast(u64, u64),
+    #[error("Packet was rejected with ErrorCode: {0}")]
+    SendMoneyPacket(ErrorCode),
+    #[error("Error maximum time exceeded since last fulfill")]
```
koivunej, line 11:
```suggestion
    #[error("Maximum time exceeded since last fulfill")]
```
Or are these variants used in non-error contexts? Note the changed prefixes altogether, in case this was used in the stream api/rejects/at some level that someone might get access to comparing the server sent messages and hardcode these.

crates/interledger-stream/src/error.rs
```diff
@@ -1,12 +1,35 @@
...
+    #[error("Unable to decrypt message")]
+    FailedToDecrypt,
+    #[error("Stream version not supported")]
+    UnsupportedVersion,
+    #[error("Frames Error: Not enough successfully parsed frames")]
+    NotEnoughValidFrames,
+    #[error("Trailing bytes error: Inner")]
+    TrailingInnerBytes,
+    #[error("Roundtrip only: Error expected for roundtrip fuzzing")]
+    Roundtrip,
```
koivunej, line 26:
```suggestion
    #[cfg(feature = "roundtrip-only")]
    #[cfg_attr(feature = "roundtrip-only", error("Roundtrip only: Error expected for roundtrip fuzzing"))]
    Roundtrip,
```

This might work?

crates/interledger-stream/src/packet.rs
```diff
@@ -159,15 +165,12 @@ impl StreamPacket {
...
+    fn from_bytes_unencrypted(mut buffer_unencrypted: BytesMut) -> Result<Self, StreamPacketError> {
         // TODO don't copy the whole packet again
         let mut reader = &buffer_unencrypted[..];
         let version = reader.read_u8()?;
         if version != STREAM_VERSION {
-            return Err(ParseError::InvalidPacket(format!(
-                "Unsupported STREAM version: {}",
-                version
-            )));
+            return Err(StreamPacketError::UnsupportedVersion);
```
koivunej, line 173: Variant-note: Is the version lost on purpose?

crates/interledger-stream/src/packet.rs
```diff
@@ -939,17 +938,15 @@ impl<'a> SerializableFrame<'a> for StreamDataBlockedFrame {
...
         reader.skip_var_octet_string()?;
 
         if cfg!(feature = "roundtrip-only") {
             // This is needed because the returned value u64::MAX
             // will make roundtrip fail, i.e. BytesMut::from(packet)
             // will not equal to the original data.
-            Err(ParseError::WrongType(
-                "Fuzzing roundtrip for var_uint larger than u64::MAX unavailable".to_string(),
-            ))
+            Err(StreamPacketError::Roundtrip)
```
koivunej, line 949: Variant-note: Same thing as with `interledger-packet::Error::Roundtrip`: the meaning is now diluted away. I think there might be value in keeping it. Perhaps a `NonRoundtrippableSaturatingAmount`?

crates/interledger-ccp/src/packet.rs
```diff
@@ -38,6 +38,53 @@ pub static CCP_CONTROL_DESTINATION: Lazy<Address> =
 pub static CCP_UPDATE_DESTINATION: Lazy<Address> =
     Lazy::new(|| Address::from_str("peer.route.update").unwrap());
 
+#[derive(Debug)]
+pub enum CcpPacketError {
+    UnexpectedMode(u8),
+    PacketExpired,
+    UnexpectedDestination(Address),
+    UnexpectedCondition,
+    IOError,
```
koivunej, line 47: The other variants still have `std::io::Error` inside of them. I think having an empty variant for it is a good compromise as information will be lost.

Could it be eradicated completely, since no I/O is actually done while parsing?

This would mean at least:

1. adding length checks
2. calling `bytes::Buf` methods instead of `byteorder` methods in `interledger-packet`

The first one can be tested as `bytes::Buf` methods except for `get_uint` should all panic; if I remember correctly the `get_uint` doesn't require any bytes so it doesn't panic (just returns 0).

koivunej (changes requested):
Overall I think this looks very good; you updated it during my scan through, so some things might already be good to resolve.

There are some string nits; perhaps the biggest issue is with the changed messages. Are you sure those can change? I think I reverted some changes as well in the bytes 0.5 upgrade, so I'd be inclined to *keep* the external behaviour (the messages) as the same by default when refactoring.

Either way, I think at least the `interledger-packet::Error` commits need a `!` to signify a breaking change (since those at minimum are exported).

EDIT: thought about it a bit, expanded the I'd be inclined to keep the external behaviour (the messages) as the same by default when refactoring.

crates/interledger-stream/src/error.rs
```diff
@@ -1,12 +1,35 @@
...
+    #[error("I/O Error: {0}")]
+    IoErr(#[from] std::io::Error),
+    #[error("Ilp PacketType Error: {0}")]
+    IplPacketType(#[from] IlpPacketTypeError),
```
koivunej, line 30:
```suggestion
    IlpPacketType(#[from] IlpPacketTypeError),
```
Was about to suggest `InnerPacketType` but then I was wondering if it was the inner or outer again. I'm leaning on it being the outer. Might be a good one to comment what this is..

crates/interledger-stream/src/error.rs
```diff
@@ -1,12 +1,35 @@
...
+    #[error("Address Error: {0}")]
+    Address(#[from] AddressError),
+    #[error("UTF-8 Error: {0}")]
+    Utf8Err(#[from] Utf8Error),
```
koivunej, line 34: When does this happen?

crates/interledger-stream/src/error.rs
whalelephant: Note sure what is inner / outer here. It is the same `PacketType` parsed in interledger-packet/packet.rs. I think these are explicitly ILP PacketType.

crates/interledger-stream/src/error.rs
whalelephant: There are `str` conversion for some frame types, `StreamCloseFrame`, `ConnectionCloseFrame` etc.

crates/interledger-stream/src/packet.rs
whalelephant: This is related to a previous discussion on `CcpPacketError::UnexpectedCondition` where because `STREAM_VERSION` is a `const` we do not store the compared to value.

koivunej: Oki. I think there might still be value in keeping the "user provided" version `u8` even though we always match it to a constant.. At least to keep the messages the same.

crates/interledger-stream/src/error.rs
koivunej: Interestingly I am not finding this from the [rfc](https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md) either, but there were of course many references to Prepares, Fulfills and Rejects.

crates/interledger-stream/src/error.rs
koivunej: From the rfc it would seem like two cases of human readable errorcode and single place for asset code. Well, probably good to leave it as is for now. I was just thinking the [`std::str::Utf8Error`](https://doc.rust-lang.org/stable/std/str/struct.Utf8Error.html) might not bring much value but at least it shouldn't be more than an `usize` + `Option<u8>`.

crates/interledger-packet/src/packet.rs
whalelephant: Kept the messages the same and specific to the `Reject.ErrorCode` as this is a refactor.

crates/interledger-stream/src/error.rs
whalelephant: This are used in error context. 
Have kept the message the same as this is a refactor

crates/interledger-ccp/src/packet.rs
koivunej: Discussed on the slack, leaving this to a follow-up PR.

crates/interledger-ccp/src/packet.rs
```diff
@@ -349,31 +399,31 @@ impl Debug for RouteUpdateRequest {
...
+            return Err(CcpPacketError::UnexpectedDestination(destination));
         }
 
         if prepare.execution_condition() != PEER_PROTOCOL_CONDITION {
             error!("Unexpected condition: {:x?}", prepare.execution_condition());
-            return Err(ParseError::InvalidPacket(format!(
-                "Wrong condition: {:?}",
-                HexString(&prepare.execution_condition()),
-            )));
+            return Err(CcpPacketError::UnexpectedCondition(
```
koivunej, line 421: the `execution_condition` should be made into `&[u8; 32]` instead. wonder if I left there a FIXME already?

crates/interledger-packet/src/errors.rs
```diff
@@ -1,24 +1,40 @@
...
 #[derive(Debug, thiserror::Error)]
 pub enum ParseError {
     #[error("I/O Error: {0}")]
     IoErr(#[from] std::io::Error),
-    #[error("UTF-8 Error: {0}")]
-    Utf8Err(#[from] Utf8Error),
-    #[error("UTF-8 Conversion Error: {0}")]
-    FromUtf8Err(#[from] FromUtf8Error),
     #[error("Chrono Error: {0}")]
     ChronoErr(#[from] chrono::ParseError),
```
koivunej, line 8: Just a note for the follow-ups: I dont think there's any point in carrying this around either. If the timestamp isn't good it doesn't really matter how bad is it. I think chrono::ParseError is quite uninformative as well.

crates/interledger-ccp/src/packet.rs
whalelephant: You left a [comment] in the code.

[comment]: https://github.com/interledger-rs/interledger-rs/pull/722/files#diff-ec91192e65045112abc93b4d8dbe90acace4df9e7e4b2e8abd0ca457e721fb07R220

crates/interledger-packet/src/packet.rs
```diff
@@ -576,10 +569,7 @@ fn deserialize_envelope(
     let got_type = reader.read_u8()?;
 
     if got_type != packet_type as u8 {
-        return Err(ParseError::InvalidPacket(format!(
-            "Unexpected packet type: {:?}",
-            got_type,
-        )));
+        return Err(PacketTypeError::Unexpected(got_type, packet_type as u8).into());
```
koivunej, line 572: Probably would not have harmed to leave the expected as PacketType, or did you check if there was some difference?

crates/interledger-packet/src/packet.rs
```diff
@@ -783,7 +769,7 @@ mod test_packet_type {
     #[test]
     fn try_from_empty() {
         assert_eq!(
-            "Invalid Packet: Unknown packet type: None",
+            "Invalid Packet: Unknown packet type",
             &PacketType::try_from(&[][..]).unwrap_err().to_string()
```
koivunej, line 773: I think this is an okay change.

crates/interledger-stream/src/client.rs
```diff
@@ -613,11 +610,12 @@ where
...
                     // Any other error will stop the rest of the payment
-                    _ => Err(Error::SendMoneyError(format!(
-                        "Packet was rejected with error: {} {}",
+                    _ => Err(Error::SendMoneyPacket(
                         reject.code(),
-                        str::from_utf8(reject.message()).unwrap_or_default(),
-                    ))),
+                        str::from_utf8(reject.message())
+                            .unwrap_or_default()
+                            .to_owned(),
```
koivunej, line 617: using `String::from_utf8_lossy(reject.message()).to_owned()` might be better because the `unwrap_or_default()` will default to empty `&'static str`.
```suggestion
                        String::from_utf8_lossy(reject.message())
                            .into_owned(),
```

crates/interledger-ccp/src/packet.rs
koivunej: Yes I did.

crates/interledger-ccp/src/packet.rs
```diff
@@ -89,31 +139,31 @@ impl Debug for RouteControlRequest {
...
-                "Packet is not a CCP message. Destination: {}",
-                prepare.destination(),
-            )));
+    pub(crate) fn try_from_without_expiry(prepare: &Prepare) -> Result<Self, CcpPacketError> {
+        let destination = prepare.destination();
+        if destination != *CCP_CONTROL_DESTINATION {
+            return Err(CcpPacketError::UnexpectedDestination(destination));
         }
 
         if prepare.execution_condition() != PEER_PROTOCOL_CONDITION {
```
koivunej, line 159: Interesting aside note: This check is duplicated along with it's log stmt... Probably not great to log and return an error but ok to clean this up later.

crates/interledger-packet/src/packet.rs
koivunej: Well, no need to anything now since this (u8, u8) was repeated in btp at least.

crates/interledger-stream/src/error.rs
```diff
@@ -1,12 +1,41 @@
...
-    #[error("Error polling: {0}")]
-    PollError(String),
-    #[error("Error polling: {0}")]
-    SendMoneyError(String),
-    #[error("Error maximum time exceeded: {0}")]
-    TimeoutError(String),
+    #[error("Terminating payment since too many packets are rejected ({0} packets fulfilled, {1} packets rejected)")]
+    PaymentFailFast(u64, u64),
+    #[error("Packet was rejected with ErrorCode: {0} {1}")]
+    SendMoneyPacket(ErrorCode, String),
```
koivunej, line 10: Was thinking maybe this should be something like `UnexpectedReject` or similar...? As in: Send money is the operation, and if and it can be stopped with unexpected reject?

crates/interledger-stream/src/error.rs
```diff
@@ -1,12 +1,41 @@
+use interledger_packet::{AddressError, ErrorCode, PacketTypeError as IlpPacketTypeError};
 /// Stream Errors
+use std::str::Utf8Error;
+
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
```
koivunej, line 6: The prefixes went away here but so I guess that means that there are no tests in the interledger-stream for error cases?

koivunej (approved):
I think this is looking great! Much less `format!(...)`.

As discussed the `std::io::Error` will be gotten rid from `interledger-packet/src/oer.rs` with a more specific simpler enum in a follow-up.

crates/interledger-stream/src/error.rs
koivunej: Well, lets just leave it like this.

crates/interledger-packet/src/packet.rs
koivunej: ... since I added the None case when I started fuzzing in #690 I think.

crates/interledger-stream/src/client.rs
koivunej: However with the lossy it might be better to format with the `{:?}` since it will escape any control chars... Did this one have a test?

crates/interledger-stream/src/error.rs
whalelephant: I agree and since we match to other `ErrorCode`, this is an unexpected rejection

crates/interledger-stream/src/client.rs
whalelephant: Yes, it does test but only `assert!(result.is_err())` for unexpected reject 
https://github.com/interledger-rs/interledger-rs/pull/722/files#diff-cdf957aff976c708dad406b806a72a2820bfe77c49895c064783bbb117d202fcR773

crates/interledger-stream/src/error.rs
whalelephant: Yes, but also the error message for `SendMoneyError` previously was incorrect.
For example, for external interface, the PaymentFailFast message was 
`Error polling: Terminating payment since too many packets are rejected (x packets fulfilled, y packets rejected)`

Therefore I have removed the prefix and used the same `String` messages.

koivunej: Oki, well that makes sense for me! Resolving this one.

crates/interledger-stream/src/client.rs
koivunej: All right, looks good! Resolving.

koivunej:
The latest `test-md` fail is bening installation failure, no worries there.

koivunej:
Rerunning as the other one failed to start as well.

koivunej:
Github actions is having issues, we'll need to restart later.

PR #723, opened 2021-05-19
refactor: replace std::io::Error with custom OerError

whalelephant:
Remove requirement to keep mapping `std::io::Error` into different crates` error type.


crates/interledger-packet/src/errors.rs
```diff
@@ -38,3 +38,49 @@ pub enum TrailingBytesError {
...
+    #[error("indefinite lengths are not allowed")]
+    IndefiniteLength,
+    #[error("length prefix too large")]
+    TooLarge,
+    #[error("length prefix Overflow")]
+    Overflow,
+    #[error("length prefix with leading zero")]
+    LeadingZeros,
+    #[error("variable length prefix with unnecessary multibyte length")]
+    UnexpectedBytes,
```
koivunej, line 67: How are these two different?

koivunej: the only place I can find is the `read_var_octet_length` for the UnexpectedBytes, but I think the issue will be the same?

As in with input `0x80 | 1, x` where `0 <= x < 128` the initial byte should had been used for the length as it would had fit. I think this part is quite explicit in the RFC: https://github.com/interledger/rfcs/blob/master/0030-notes-on-oer-encoding/0030-notes-on-oer-encoding.md#long-form

Aah yes but now I understand, the messages were different, so two variants were created. I think that two distinct variants makes sense. The `LeadingZeros` could even be `#[cfg(feature = "strict")]` which should not be an issue.

Wonder if the `UnexpectedBytes` be more fitting with `UnexpectedLongForm` (per rfc?) or could just have two LeadingZeros:
- LeadingZeros -- could explicitly link to the rfc's "Long form" header
- StrictLeadingZeros -- only in strict, I guess this would be the ref: https://github.com/interledger/rfcs/blob/master/0030-notes-on-oer-encoding/0030-notes-on-oer-encoding.md#variable-length-unsigned-integer even though it says it's not used...

crates/ilp-node/Cargo.toml
```diff
@@ -75,13 +75,11 @@ metrics-core = { version = "0.5.1", default-features = false, optional = true }
 metrics-runtime = { version = "0.13.0", default-features = false, features = ["metrics-observer-prometheus"], optional = true }
 
 [dev-dependencies]
-approx = { version = "0.3.2", default-features = false }
 base64 = { version = "0.11.0", default-features = false }
 socket2 = "0.3.15"
 rand = { version = "0.7.2", default-features = false }
 reqwest = { version = "0.10.0", default-features = false, features = ["default-tls", "json"] }
 serde_json = { version = "1.0.41", default-features = false }
-tokio-retry = { version = "0.2.0", default-features = false }
```
koivunej, line 84: Very nice finds! Big win here already for tokio upgrading..

crates/interledger-btp/src/errors.rs
```diff
@@ -1,15 +1,20 @@
+use interledger_packet::OerError;
 use std::str::Utf8Error;
 
 #[derive(Debug, thiserror::Error)]
 pub enum BtpPacketError {
-    #[error("I/O Error: {0}")]
-    IoErr(#[from] std::io::Error),
+    #[error("too short packet")]
+    UnexpectedEof,
```
koivunej, line 7: This would duplicate the `Oer(OerError::UnexpectedEof)` and similarly in others but I guess this is for the different string...?

koivunej (commented):
I think this is looking great! The major question is to go with the added `read_{u{8,16,32},exact}` or just with `bytes::Buf` methods and have an oblig length check before deserialization.

I think the single big length check would be preferable, because afterwards the LLVM can probably see that the `bytes::Buf` methods cannot panic.

Regarding the failed `exchange_rates.rs` test... We might need to disable it, unless there was just some temporary issue.

crates/interledger-packet/src/packet.rs
```diff
@@ -126,7 +125,9 @@ impl TryFrom<BytesMut> for Prepare {
     fn try_from(buffer: BytesMut) -> Result<Self, Self::Error> {
         let (content_offset, mut content) = deserialize_envelope(PacketType::Prepare, &buffer)?;
         let content_len = content.len();
-        let amount = content.read_u64::<BigEndian>()?;
+
+        // This will always return since content was returned by read_var_octet_string
```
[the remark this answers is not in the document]
koivunej: Might be good to recheck the fuzzing on this one either way. If we don't find a fast panic then something is wrong with the fuzz target or on my reading comprehension.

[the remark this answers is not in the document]
koivunej: `[12, 0]` will also hit a panic in this line.

crates/interledger-btp/src/errors.rs
whalelephant: Yes for the string but also the check for packet length in this case is done in this crate, not code in `Oer`.

crates/interledger-packet/src/packet.rs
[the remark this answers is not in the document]
whalelephant: you are correct. This does not always return and need to be checked.

crates/interledger-packet/src/errors.rs
```diff
@@ -38,3 +38,49 @@ pub enum TrailingBytesError {
     #[error("Unexpected inner trailing bytes")]
     Inner,
 }
+
+#[derive(PartialEq, Debug, thiserror::Error)]
+pub enum OerError {
+    #[error("OerError")]
+    Error,
```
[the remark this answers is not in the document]
whalelephant: removed

crates/interledger-ccp/src/packet.rs
```diff
@@ -439,12 +448,20 @@ impl RouteUpdateRequest {
...
+        }
         let mut routing_table_id: [u8; 16] = [0; 16];
-        data.read_exact(&mut routing_table_id)?;
-        let current_epoch_index = data.read_u32::<BigEndian>()?;
-        let from_epoch_index = data.read_u32::<BigEndian>()?;
-        let to_epoch_index = data.read_u32::<BigEndian>()?;
-        let hold_down_time = data.read_u32::<BigEndian>()?;
+        data.copy_to_slice(&mut routing_table_id);
+        let current_epoch_index = data.get_u32();
+        let from_epoch_index = data.get_u32();
```
koivunej, line 462: Crash here with the fuzz_target_2: `[10, 255, 255, 255, 255, 159, 159, 159, 159, 49, 10, 255, 10, 159, 159, 159, 46, 255, 255, 255]`.

The min length should be at least: 16 where it's currently 2 for the routing table id, then there should also be one for the Address or read_var_octet_string (0 is the minimum payload for that).

So up until Address there should be 16 + 4 + 4 + 4 + 4 + 1, after that I think a "+ 2" can be added (minimum varuint) and after that it's more complicated. If the routes were simple I would had done the preallocation checks here as well.

crates/interledger-ccp/src/packet.rs
```diff
@@ -439,12 +448,20 @@ impl RouteUpdateRequest {
     }
 
     fn try_from_data(mut data: &[u8]) -> Result<Self, CcpPacketError> {
+        // RouteUpdateRequest: routing_table (2)
+        // + current_epoch_index (4)
+        // + from_epoch_index (4)
+        // + to_epoch_index (4)
+        // + hold_down_time (4)
+        if data.remaining() < 2 + 4 + 4 + 4 + 4 {
```
koivunej, line 456:
```suggestion
        // RouteUpdateRequest: routing_table_id (16)
        // + current_epoch_index (4)
        // + from_epoch_index (4)
        // + to_epoch_index (4)
        // + hold_down_time (4)
        // + address (1, too low bound)
        // + var uint (2, correct low bound)
        if data.remaining() < 16 + 4 + 4 + 4 + 4 + 1 + 2 {
```

crates/interledger-ccp/src/packet.rs
```diff
@@ -179,10 +177,14 @@ impl RouteControlRequest {
     }
 
     fn try_from_data(mut data: &[u8]) -> Result<Self, CcpPacketError> {
-        let mode = Mode::try_from(data.read_u8()?)?;
+        // RouteControlRequest: mode (4) + last_known_routing_table_id (16) + last_known_epoch (4)
```
koivunej, line 180:
```suggestion
        // RouteControlRequest: mode (1) + last_known_routing_table_id (16) + last_known_epoch (4)
```

koivunej (approved):
With the suggestions I think this is looking good!

crates/interledger-ccp/src/packet.rs
koivunej: Could you add the crash as a test case?

crates/interledger-ccp/src/packet.rs
koivunej: Resolving, as I pr'd the changes: https://github.com/whalelephant/interledger-rs/pull/1

crates/interledger-ccp/src/packet.rs
koivunej: Resolving as I pr'd: https://github.com/whalelephant/interledger-rs/pull/1

crates/interledger-ccp/src/packet.rs
koivunej: Added in https://github.com/whalelephant/interledger-rs/pull/1.

crates/interledger-service-util/src/echo_service.rs
```diff
@@ -73,19 +72,17 @@ where
...
-                eprintln!("Could not read packet type: {:?}", error);
-                return Err(RejectBuilder {
-                    code: ErrorCode::F01_INVALID_PACKET,
-                    message: b"Could not read echo packet type.",
-                    triggered_by: Some(&ilp_address),
-                    data: &[],
-                }
-                .build());
+        if reader.remaining() < 1 {
+            eprintln!("Could not read packet type: Unexpected Eof");
```
koivunej, line 76: This is a rather interesting one. I would assume that this was supposed to be an `error!`, there are also some similar cases in other crates. Can PR this in once we have merged this PR.

koivunej:
Oki I understand that the PR is now ready to be merged.

crates/interledger-btp/src/errors.rs
```diff
@@ -1,15 +1,18 @@
...
-    IoErr(#[from] std::io::Error),
+    #[error("extra trailing bytes")]
+    TrailingBytesErr,
     #[error("UTF-8 Error: {0}")]
     Utf8Err(#[from] Utf8Error),
     #[error("Chrono Error: {0}")]
     ChronoErr(#[from] chrono::ParseError),
     #[error("PacketType Error: {0}")]
     PacketType(#[from] PacketTypeError),
+    #[error("Oer Error: {0:?}")]
```
pradovic, line 14: I would maybe omit the word `Error` from errors here and where formatting as well. 
The reason for this is subjective so feel free to ignore if you don't agree. The reason being, in my mind, is that it is of an Error enum, so we know that it is an Error + if it ends up being printed it is usually in `error!()` so it will have an error scope anyways.

crates/interledger-btp/src/packet.rs
```diff
@@ -164,11 +162,15 @@ pub struct BtpMessage {
 impl Serializable<BtpMessage> for BtpMessage {
     fn from_bytes(bytes: &[u8]) -> Result<BtpMessage, BtpPacketError> {
         let mut reader = bytes;
-        let packet_type = reader.read_u8()?;
+        // BtpMessage: packet type (1) + request_id (4) + protocol data (length checked in Oer)
+        if reader.remaining() < 1 + 4 {
```
pradovic, line 166: I would maybe extract 1 and 4 as constants and use it, together with a comment. This also makes it more sensical from the doc standpoint, imo.

crates/interledger-btp/src/packet.rs
```diff
@@ -201,13 +203,17 @@ pub struct BtpResponse {
 impl Serializable<BtpResponse> for BtpResponse {
     fn from_bytes(bytes: &[u8]) -> Result<BtpResponse, BtpPacketError> {
         let mut reader = bytes;
-        let packet_type = reader.read_u8()?;
+        // BtpReponse: packet type (1) + request_id (4) + protocol_data (length checked in Oer)
+        if reader.remaining() < 1 + 4 {
```
pradovic, line 207: Same as above :)

crates/interledger-btp/src/packet.rs
```diff
@@ -242,17 +248,25 @@ pub struct BtpError {
 impl Serializable<BtpError> for BtpError {
     fn from_bytes(bytes: &[u8]) -> Result<BtpError, BtpPacketError> {
         let mut reader = bytes;
-        let packet_type = reader.read_u8()?;
+        // BtpError: packet type (1) + request_id (4) + content (checked in Oer) + code (3) + rest
+        // of data (length checked in Oer)
+        if reader.remaining() < 1 + 4 {
```
pradovic, line 253: Same as above :)

crates/interledger-ccp/src/packet.rs
```diff
@@ -66,16 +64,16 @@ impl fmt::Display for CcpPacketError {
             CcpPacketError::UnexpectedCondition(c) => {
                 write!(fmt, "Invalid Packet: Wrong condition: {:?}", HexString(c))
             }
-            CcpPacketError::IOError => write!(fmt, "I/O Error"),
+            CcpPacketError::Oer(err) => write!(fmt, "Oer Error {:?}", err),
```
pradovic, line 67: Same comment for formatting as the one above, maybe remove `Error` word.

crates/interledger-ccp/src/packet.rs
```diff
@@ -179,10 +177,14 @@ impl RouteControlRequest {
     }
 
     fn try_from_data(mut data: &[u8]) -> Result<Self, CcpPacketError> {
-        let mode = Mode::try_from(data.read_u8()?)?;
+        // RouteControlRequest: mode (1) + last_known_routing_table_id (16) + last_known_epoch (4)
+        if data.remaining() < 1 + 16 + 4 {
```
pradovic, line 181: Similar to the above comment, these could be constants.

crates/interledger-ccp/src/packet.rs
```diff
@@ -242,14 +244,18 @@ impl TryFrom<&mut &[u8]> for RouteProp {
 
     // Note this takes a mutable ref to the slice so that it advances the cursor in the original slice
     fn try_from(data: &mut &[u8]) -> Result<Self, Self::Error> {
-        let meta = data.read_u8()?;
+        // RouteProp: flag meta data (1) + id (2)
+        if data.remaining() < 1 + 2 {
```
pradovic, line 248: Same as above.

crates/interledger-packet/src/errors.rs
```diff
@@ -2,8 +2,8 @@ use super::AddressError;
 
 #[derive(Debug, thiserror::Error)]
 pub enum ParseError {
-    #[error("I/O Error: {0}")]
-    IoErr(#[from] std::io::Error),
+    #[error("Oer Error: {0}")]
```
pradovic, line 5: Same as above about the  word :)

crates/interledger-packet/src/errors.rs
```diff
@@ -2,8 +2,8 @@ use super::AddressError;
...
+    Oer(#[from] OerError),
     #[error("Chrono Error: {0}")]
     ChronoErr(#[from] chrono::ParseError),
     #[error("PacketType Error: {0}")]
```
pradovic, line 9: I know it';s not part of this PR but we could ommit the `Error` word here as well.

pradovic (changes requested):
Great job!
I added a couple of subjective comments, so if you don't agree with them you can skip, it is still good to go from me!

crates/interledger-btp/src/errors.rs
koivunej: I had missed this one and I think we need to align the how the OerError is formatted; as in it's `{}` in the latter two and `{:?}` in the first two.

pradovic: @koivunej correct, I have missed this as well!

crates/interledger-packet/src/errors.rs
koivunej: This was added in the prev PR #722 so should be good to alter; however there's an argument to be made for keeping it as others have as well... In generally I think we should try not to change error messages when refactoring as those could be dependended on log scanners and possibly used in interfaces like http.

crates/interledger-btp/src/errors.rs
koivunej: I remember wondering on the weekend that "Oer Error" with or without "Error" might not be so clear on the logs. It could be restored to "Invalid Packet: $reason", which it used to be.

koivunej: I'll resolve these as I've now PR'd whalelephant these fixes.

crates/interledger-btp/src/packet.rs
koivunej: extracted soon in a commit I'll push once tested.

crates/interledger-ccp/src/packet.rs
koivunej: Replaced with "older" (dunno if really) `Invalid Packet: {}`

pradovic (approved):
LGTM! 👏  nice job!

koivunej:
Cannot get any reliable perf changes, but at least there's no more IO error, so this is a big win for readability.

PR #729, opened 2021-08-02
feat: tokio upgrade

Mirko-von-Leipzig:
Quite a large and difficult refactor. There are some issues remaining which require feedback.

1. There is now a stack-overflow when running in debug mode. This is a [bug in warp](https://github.com/seanmonstar/warp/issues/811#issuecomment-873558661)
2. Some FIXME `todo!` I left for format to strings where I don't know what info is valuable. [here](https://github.com/interledger-rs/interledger-rs/compare/master...Mirko-von-Leipzig:feat/tokio_upgrade?expand=1#diff-90a5a393611aaa47be31de7b4c5af6c74f954b83e7dd074021218a5642489ce6R21) and [here](https://github.com/interledger-rs/interledger-rs/compare/master...Mirko-von-Leipzig:feat/tokio_upgrade?expand=1#diff-6231b5a098f08f2308be37069ea123fe5ec93e19f70bf4fa1dfb85eeeb792d45R21)

Test status remained the same for me before and after. Tests can all pass; but the `ilp-node` tests seem plagued by racey start-up conditions (?) and sporadically fail -- this behaviour was the same before the refactor.


crates/ilp-node/Cargo.toml
```diff
@@ -77,7 +77,7 @@ metrics-runtime = { version = "0.13.0", default-features = false, features = ["m
 
 [dev-dependencies]
 base64 = { version = "0.11.0", default-features = false }
-socket2 = "0.3.15"
+socket2 = "0.4.0"
```
koivunej, line 80: Just a note while walking the commits, I think I remember a related RUSTSEC being ignored on cargo-audit running of CI.

crates/ilp-node/tests/redis/exchange_rates.rs
```diff
@@ -64,12 +63,13 @@ async fn coincap() {
...
-    let api_key = env::var("ILP_TEST_CRYPTOCOMPARE_API_KEY");
-    if api_key.is_err() {
-        error!("Skipping cryptocompare test. Must configure an API key by setting ILP_TEST_CRYPTOCOMPARE_API_KEY to run this test");
-        return;
-    }
-    let api_key = SecretString::new(api_key.unwrap());
+    let api_key = match env::var("ILP_TEST_CRYPTOCOMPARE_API_KEY") {
+        Ok(value) => value,
+        Err(_) => {
+            error!("Skipping cryptocompare test. Must configure an API key by setting ILP_TEST_CRYPTOCOMPARE_API_KEY to run this test");
```
koivunej, line 69: perhaps this should be just a println, because there isn't a logger configured, which is probably why I've never learned of existence of this :) Cannot suggest while in commit-per-commit view.

crates/ilp-node/Cargo.toml
koivunej: I misremembered, it was about net2, however it should no longer be in the dependencies right? https://github.com/interledger-rs/interledger-rs/blob/629f8b52809e58141401b37548cad39152626d4c/.github/workflows/ci.yml#L52

koivunej:
> There is now a stack-overflow when running in debug mode.

How come this didn't manifest itself in test-md build...? Surely it does a lot of filters. Did you add unconditional boxing to the filters and I missed it while doing the commit-per-commit?

Mirko-von-Leipzig:
> 
> 
> > There is now a stack-overflow when running in debug mode.
> 
> How come this didn't manifest itself in test-md build...? Surely it does a lot of filters. Did you add unconditional boxing to the filters and I missed it while doing the commit-per-commit?

I never managed to "fix" the issue by boxing or anything -- I presumed its because I didn't know where to look in the codebase. So its possible its a different issue; but its unlikely as:

1. Release build works, Debug does not (as described in the issue)
2. Using `println` debugging indicates that its when the warp serve gets called.

Mirko-von-Leipzig:
Force pushed with `fmt` fixes

crates/ilp-node/Cargo.toml
Mirko-von-Leipzig: `cargo tree` shows no `net2` so its a dependency.

koivunej: > `net2` so its not a dependency

`RUSTSEC-2020-0016` could be removed as well then?

crates/ilp-node/tests/redis/redis_helpers.rs
```diff
@@ -13,17 +13,23 @@ use std::time::Duration;
...
         ConnectionAddr::Unix(path) => {
-            format!("redis+unix:{}?db={}", path.to_str().unwrap(), info.db)
+            format!("redis+unix:{}?db={}", path.to_str().unwrap(), info.redis.db)
         }
+        // FIXME: no idea what this is used for
+        ConnectionAddr::TcpTls {
+            host,
+            port,
+            insecure,
+        } => todo!(),
```
koivunej, line 26:
```suggestion
        // FIXME: no idea what this is used for
        x => unimplemented!("unsupported {:?}", x),
```

The above could work. However sadly the redis_helpers has been copypasted around, which reminds me I never remembered to PR that refactoring in which would had made these into on. Perhaps leave these as is, you probably marked all of these as FIXME and we'll create an issue on unifying these?

Mirko-von-Leipzig: I marked the two I added/touched during the upgrade as FIXME -- so if there is other similar code around then I didn't mark that.

Mirko-von-Leipzig:
Rebase'd master changes (CI speed improvements)

Mirko-von-Leipzig:
Commit [e72bd5c](https://github.com/interledger-rs/interledger-rs/pull/729/commits/e72bd5cfb04f47c70e2a59da71b82b802bcaaa01) introduces a work-around fix for the warp stack-overflow bug (mentioned in PR description).

Further investigation indicates that no other stack-overflow issues remain (from this bug) -- used test coverage to check all instances of filters are covered or manually inspected if missed.

crates/ilp-node/tests/redis/redis_helpers.rs
koivunej: Created the #731 for this so resolving.

koivunej:
Allrighty, this is ready to merge. This a big step forwards, thanks a lot!

PR #730, opened 2021-08-03
Clippy warnings in 1.54

omertoast:


koivunej (approved):
Looking good! Thanks for this!

koivunej:
The error variant renames are luckily only on ilp-cli, so I am not marking this as breaking.

crates/ilp-cli/src/interpreter.rs
```diff
@@ -12,22 +12,22 @@ use url::Url;
...
     // Foreign errors
     #[error("Error sending HTTP request: {0}")]
-    SendErr(#[from] reqwest::Error),
+    Send(#[from] reqwest::Error),
     #[error("Error receving HTTP response from testnet: {0}")]
-    TestnetErr(reqwest::Error),
+    Testnet(reqwest::Error),
     #[error("Error altering URL scheme")]
-    SchemeErr(()), // TODO: should be part of UrlError, see https://github.com/servo/rust-url/issues/299
+    Scheme(()), // TODO: should be part of Urlor, see https://github.com/servo/rust-url/issues/299
```
koivunej, line 24:
```suggestion
    Scheme(()), // TODO: should be part of UrlError, see https://github.com/servo/rust-url/issues/299
```

Probably accidential? The linked issue is still open, so looking good.

koivunej:
Merging this ahead of grace period once the build completes per this being a regular chore and this blocking #729.

PR #732, opened 2021-08-06
CI speedup

Mirko-von-Leipzig:
This PR adds dependency caching to the CI workflow, dramatically speeding up builds (after the first one). Note: does not include the docker workflow -- may be of interest too (?) but I'm not sure how it could work.

Also adds `workflow_dispatch` trigger, which will let us manually trigger the workflow. Mainly useful for testing on forks without having to fiddle with the workflow file.

Uses the [rust-cache action](https://github.com/Swatinem/rust-cache). We could try manually figure out which parts of `/target` to cache, but I found it difficult and prone to error. This action seems to do the right thing and ends up caching about 2 GB total with the way I have it setup (multiple jobs, with independent caches).

It may be possible to further optimize the cache by sharing amongst the jobs, but I'm not quite certain how.

What I found surprising is that the `clippy` check is so slow -- it seems to be checking *all* depedencies. Might be worth investigating. I did a small search which turned up the `--no-deps` option but it didn't change anything for me. This is only for a non-cached run -- since the cache will include the dependency clippy checks for the next run.

Comparison between runs:
| Run  | Duration | Cumulative time |
| --- | --- | --- |
| [before](https://github.com/interledger-rs/interledger-rs/actions/runs/1093773784) | 17m 39s | 33m 37s |
| [no cache](https://github.com/Mirko-von-Leipzig/interledger-rs/actions/runs/1104556424) | 12m 51s | 30m 1s |
| [cache](https://github.com/Mirko-von-Leipzig/interledger-rs/actions/runs/1104557123) | 7m 54s | 15m 47s |

Cache size per job:
| Job | Cache |
| --- | --- |
| Interledger | ~440 MB |
| Interledger-packet | ~166 MB |
| Interledger-stream | ~251 MB |
| Interledger-btp | ~294 MB |
| Settlement-engines | ~364 MB |
| test-md | ~335 MB |
| **Total** | **~1.85 GB** |

Some restrictions on [github action caching](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy):
- maximum size of 5GB *total* per repo - replaces oldest cache automatically on overflow
- gets cleared automatically after 7 days of inactivity

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,11 +2,8 @@ name: CI
 
 # limit ci building to pushes to master not to get twice the notification email
 # from github.
-#
-# if you want to test your branch on your fork, just replace the 'branches'
-# filter below for 'push' with the one for 'pull_request' temporarily and drop
-# the commit when creating the PR.
 on:
+  workflow_dispatch:
   push:
     branches:
       - 'master'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,9 @@ jobs:
           rust-version: stable
           components: clippy, rustfmt
 
+      - name: Cache
+        uses: Swatinem/rust-cache@v1
+
       - name: Build
         run: cargo build --all-features --all-targets
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,16 +130,69 @@ jobs:
       - name: Test --features strict
         run: timeout 15m cargo test -p interledger-btp --features strict
 
-  test-md:
+
+  settlement-engines:
+    # This job builds the ethereum-settlement-engine required by `test-md`.
+    # A separate job to enable dependency caching.
     runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+        with:
+          repository: interledger-rs/settlement-engines
+          ref: master
+
+      - name: Install rust toolchain
+        uses: hecrj/setup-rust-action@v1.3.4
+        with:
+          rust-version: stable
+
+      - name: Cache
+        uses: Swatinem/rust-cache@v1
+
+      - name: Build
+        run: cargo build --bin ilp-settlement-ethereum
+
+      - name: Upload settlement-engine artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: settlement-engine
+          path: target/debug/ilp-settlement-ethereum
+
+  test-md:
+    env:
+      RUST_LOG: "interledger=trace"
+      RUST_BACKTRACE: "full"
+
+    needs: settlement-engines
+
+    runs-on: ubuntu-latest
+
     steps:
       - name: Checkout sources
         uses: actions/checkout@v2
 
-      - name: Install node
-        uses: actions/setup-node@v2
+      - name: Install rust toolchain
+        uses: hecrj/setup-rust-action@v1.3.4
         with:
-          node-version: 'v12.18.4'
+          rust-version: stable
+
+      - name: Cache
+        uses: Swatinem/rust-cache@v1
+
+      - name: Build binaries
+        run: |
+          mkdir -p ~/.interledger/bin
+          cargo build --bin ilp-node --bin ilp-cli
+          mv target/debug/ilp-node ~/.interledger/bin
+          mv target/debug/ilp-cli ~/.interledger/bin
+
+      - name: Download settlement engine artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: settlement-engine
+          path: ~/.interledger/bin
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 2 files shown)


koivunej (approved):
This seems very nice, thanks! I'm thinking this will not need any grace period as the run just passes.

.github/workflows/ci.yml
```diff
@@ -54,33 +54,154 @@ jobs:
...
+      - name: Install rust toolchain
+        uses: hecrj/setup-rust-action@v1.3.4
+        with:
+          rust-version: stable
+
+      - name: Cache
+        uses: Swatinem/rust-cache@v1
+
+      - name: Build
+        run: cargo build -p interledger-stream
```
koivunej, line 106: Dunno if I had added this previously but I don't think this is needed. Did you have something special in mind if you added this in addition to the tests? Same for `-packet` and `-btp`.

I think the `test` compilations will not benefit from the earlier build, nor should there be anything which is not covered by these.

koivunej:
> What I found surprising is that the clippy check is so slow -- it seems to be checking all depedencies.

I think it wont benefit much from the previous `cargo build` above it because it'll do the `--all-targets` dance but it might help the `cargo test` which comes after it.

koivunej:
What's still missing from the CI workflow which I suddenly remembered is checking the fuzz targets. I don't think it can be done without `cargo +nightly fuzz build`, which I think `serde/json` was also using. This would also add nightly compiler installation together with `cargo install cargo-fuzz` which costs around 3min of user time for me which would be I guess 90s with 2 cpus.

Could just create a new issue out of this.

.github/workflows/ci.yml
Mirko-von-Leipzig: Ah no. I added it in to see more clearly what was failing (while checking things still).

The benefit is having better glance value when seeing which "step" failed.

So if the build itself fails then one will see `Build` as red. Otherwise one has to dig one level deeper since the `Test` will be red, and then inside one will see that its because the build failed *shrugs* -- and not the tests that failed.

I'm happy to remove them again.

koivunej: > The benefit is having better glance value when seeing which "step" failed.

Aah yes, I was only thinking about optimizing for speed/cache size. Could be this might be of a valid help in some cases, never thought about it as I haven't never spent long time wondering which crate failed in the workspace. I think the `cargo` error text is quite easy to read, but can't say I have spent any non-trivial amount of time wondering about this.

Thinking this a bit more: If the `cargo build` failure caused by some crate was difficult to read, I would suspect that to be very high priority bug to fix, so I'm leaning towards at least testing removing these `cargo build`s.

Mirko-von-Leipzig:
> 
> 
> What's still missing from the CI workflow which I suddenly remembered is checking the fuzz targets. I don't think it can be done without `cargo +nightly fuzz build`, which I think `serde/json` was also using. This would also add nightly compiler installation together with `cargo install cargo-fuzz` which costs around 3min of user time for me which would be I guess 90s with 2 cpus.
> 
> Could just create a new issue out of this.

A new issue would be best I think. Different improvent :)

.github/workflows/ci.yml
Mirko-von-Leipzig: Sure 👍 Will remove them.

.github/workflows/ci.yml
```diff
@@ -91,8 +203,7 @@ jobs:
           npm install -g ganache-cli ilp-settlement-xrp conventional-changelog-cli
```
koivunej, line 203:
```suggestion
          npm install -g ganache-cli ilp-settlement-xrp
```

This hit my eye, `rg conventional-changelog` doesn't bring up any matches.

Could create a ci cleanups issue though I'm sure this PR already covers the most.

koivunej:
Latest run in 7min 7s, very nice.

koivunej:
@Mirko-von-Leipzig could you restore a workflow called "build" to be a dummy `Always succeed` because this repository is configured with both of the named workflows being required for merge? :)

.github/workflows/ci.yml
```diff
@@ -2,19 +2,16 @@ name: CI
...
+  workflow_dispatch:
   push:
     branches:
       - 'master'
   pull_request:
     branches: '*'
 
 jobs:
-  build:
+  interledger:
```
koivunej, line 14:
```suggestion
  build:
```

Per repo configuration requirement.

.github/workflows/ci.yml
Mirko-von-Leipzig: I'll remove this then 👍

.github/workflows/ci.yml
Mirko-von-Leipzig: I'll make another `build` job, which just depends on the `interledger*` jobs -- this will keep the semantics the same as previously?

koivunej: Aah yes, makes sense. The easiest thing that comes to mind is a single step of `true`, but perhaps more true answer would be to have it wait for the interledger one, but can't remember how did those events work.

koivunej: `needs` sounds about right, resolving.

PR #734, opened 2021-08-11
ci: add commit checks for PRs

Mirko-von-Leipzig:
This PR introduces a new github workflow which ensures all commits in a PR conform to our requirements. It checks for the following:

1. conventional commit messages
2. used `git commit --signoff`
3. no extra merge commits

### Conventional commits and `--signoff`

The first two points are taken care of by the NPM [`commitlint`](https://github.com/conventional-changelog/commitlint) package. Unfortunately, they provide guides for CircleCI and TravisCI but not for Github Actions ([open issue](https://github.com/conventional-changelog/commitlint/issues/586)). I managed to get it working after some trial and error. 

There is a 3rd party Github Action which looks great, but unfortunately I ran into [this bug](https://github.com/wagoid/commitlint-github-action/issues/187). It also seemed like a lot of code to audit, considering we don't need much.

Another alternative could be to use [`git interpret-trailers`](https://git-scm.com/docs/git-interpret-trailers) to check for `--signoff`, but since this came bundled with `commitlint` I didn't explore further.

### no extra merge commits

This gets checked with some simple bash code walking the PR commits using `git log --merges --ancestry-path`. Luckily the Github PR event contains the `HEAD` and number of commits in the PR.

### Why a new workflow file?

It has different run `on` requirements and therefore requires its own file (I think?).

It made sense to me that it should only run on actual PRs. It would otherwise trigger (and probably fail) when wanting to test things quickly.

I also enabled it for PRs into any branch (and not just into master), since its likely the code would eventually make its way into a PR for `master`, but by that time it may be too late to get the commit requirements fixed.

### cargo audit

I've also added in the ignore for the two hyper RUSTSECs.

### DIFF
--- a/.github/workflows/commits.yml
+++ b/.github/workflows/commits.yml
@@ -1,0 +1,69 @@
+# Checks all commits in a PR follow the repo rules:
+#
+#   1. conventional commit messages
+#   2. used `git commit --signoff`
+#   3. no extra merge commits
+name: Commits
+
+on:
+  workflow_dispatch:
+  # Perform these checks on PRs into *any* branch.
+  #
+  # Motivation:
+  #   Commits which are not --signoff but merged into other branches
+  #   will likely make their way into PRs for master. At which
+  #   point it will be difficult to get the original author to --signoff.
+  pull_request:
+
+jobs:
+  commit-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set commit range variables
+        # Finding the commit range is not as trivial as it may seem.
+        #
+        # At this stage git's HEAD does not refer to the latest commit in the PR,
+        # but rather to the merge commit inserted by the PR. So instead we have
+        # to get 'HEAD' from the PR event.
+        #
+        # One cannot use the number of commits (github.event.pull_request.commits)
+        # to find the start commit i.e. HEAD~N does not work, this breaks if there
+        # are merge commits.
+        run: |
+          echo "PR_HEAD=${{ github.event.pull_request.head.sha }}" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/commits.yml
```diff
@@ -0,0 +1,64 @@
...
+          # This will list any merge commits in the PR commit path
+          MERGE=$(git log --merges --ancestry-path $PR_HEAD~${{ github.event.pull_request.commits }}..$PR_HEAD)
+
+          # The merge list should be empty
+          [[ ! -z "$MERGE" ]] && {
+            echo "PR contains merge commits:";
+            echo $MERGE;
+            exit 1;
+          }
+          exit 0;
```
koivunej, line 64:
```suggestion
          exit 0;

```

Not sure if my suggestion works; github and git complain whenever text files don't end with newline which makes them binary files according to some.

Mirko-von-Leipzig: That's.. interesting.

koivunej:
Could you add the missing newline in a non-signed off commit, possibly being result of something you merged into this PR branch, later rebase it?

Mirko-von-Leipzig:
> 
> 
> Could you add the missing newline in a non-signed off commit, possibly being result of something you merged into this PR branch, later rebase it?

I'll add a dummy commit 👍

.github/workflows/commits.yml
koivunej: Recommend you to find this option in your `$EDITOR` for future. In my global editorconfig it's just:

```
[*]
insert_final_newline = true
```

But I think this is a noop with vim/neovim.

koivunej:
Oki could you now add a merge commit as well?

Mirko-von-Leipzig:
> 
> 
> Oki could you now add a merge commit as well?

mmm. What I did broke something here

koivunej:
Thank you @Mirko-von-Leipzig very much!

PR #735, opened 2021-11-30
Add tracking info

KristianWahlroos:
There is now `get_spsp_with_tracking`


darentuzi:
This change allows for payment initiators to append some metadata to a payment.
Currently an SPSP request 
`GET http://url/accounts/main_account/spsp/`
would be transformed into a reply containing ilp address
`test.example.main_account`

with this additional method, the new code supports the old format but also an optional format like
`GET http://url/accounts/main_account/spsp/key_value~orAnyOtherBlobOfText`
would be transformed into a reply containing ilp address
`test.example.main_account.key_value~orAnyOtherBlobOfText`

The extra information is appended as ILP address interaction as defined by https://interledger.org/rfcs/0015-ilp-addresses/

frnsskk (approved):

PR #744, opened 2022-07-30
Require unicase >= 2.6 to ensure correct username comparison

amousset:
When using unicase 2.5.1 the username comparison is not correct and substrings are considered equal due to https://github.com/seanmonstar/unicase/issues/38.

Requiring 2.6.0 (which is already used in the `Cargo.lock`) prevents the problem.


koivunej:
Thanks for this, I'll figure out what's wrong with string escaping first in another PR.

koivunej (approved):

koivunej:
@amousset apologies, could you first refresh your `master` branch, then do a `git rebase --signoff master` and force-push? There's a repo requirement for signed off commits.

